PR #17, opened 2021-04-01
JSON parser

mrLSD:
* added json parser (for eth-connector)
* modified panic invocations as unreachable


artob (approved):
LGTM, thanks :+1:

PR #19, opened 2021-04-02
Add eth-prover & extend types

mrLSD:
* Added eth-prover
* Extended types for eth-prover


src/types.rs
```diff
@@ -26,6 +34,136 @@ pub struct InternalMetaCallArgs {
     pub input: Vec<u8>,
 }
 
+/// eth-connector initial args
+#[cfg(feature = "contract")]
+pub struct InitCallArgs {
```
artob, line 39: Borsh-encoded parameter structs should go in `src/parameters.rs`

mrLSD: @artob moved

PR #20, opened 2021-04-03
fix meta parsing test

joshuajbouw:
This was a bit of a pain to fix. I assumed the wrong parts of it were broken initially. After much frusteration and pulling my hair out and reading and re-reading ETH documentation, I knew full well about the extra byte and had assumed that certainly had to do with it. Finally, had an epiphany which helped progress. Unfortunately it was still failing. That was because of a new addition that I added which I originally thought would be part of the solution. There were two minor fixes that had solved this issue. Mostly, positioning of that extra byte. It just was not entirely obvious initially. The good news is that I feel quite well versed with ecrecovery now.

### DIFF
--- a/src/meta_parsing.rs
+++ b/src/meta_parsing.rs
@@ -228,7 +228,7 @@ impl MethodAndTypes {
         let mut parsed_types = HashMap::new();
         let mut type_sequences = vec![];
         let (method, mut types) = Method::parse(method_def)?;
-        while types.is_empty() {
+        while !types.is_empty() {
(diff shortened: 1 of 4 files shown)


artob (approved):
LGTM :+1: Indeed, our original `ecrecover` implementation wanted the `v` byte in the wrong place, which explains why all this was needed after we fixed `ecrecover`in fc5a73f4e60f638cdedd1fd4175212dfde582be1.

mrLSD (approved):

PR #21, opened 2021-04-05
feat: add precompiles

joshuajbouw:
Adds precompiles to the aurora engine.

- [x] identity
- [x] sha256
- [x] ripemd
- [x] modexp
- [x] alt_bn128_add
- [x] alt_bn128_mul
- [x] alt_bn128_pair
- [x] blake2f


joshuajbouw:
I had changed the inputs but kept the same output for the blake2f function @artob. Would you like me to revert it back? According to the implementation on go ETH client, this is the same inputs, but the output is different. As in, normally it just returns the bytes. Anyways, let me know what I should do there.

joshuajbouw:
Shouldn't there be checks for gas spent in something like blake2f or is that checked elsewhere?

joshuajbouw:
All precompiles are in now. I just need to find actual inputs with expected results. All the failures as per spec are caught.

artob:
> All precompiles are in now. I just need to find actual inputs with expected results. All the failures as per spec are caught.

@joshuajbouw We have test cases for every Istanbul HF precompile here:

https://github.com/near/nearcore/blob/master/runtime/near-evm-runner/tests/contracts/StandardPrecompiles.sol

joshuajbouw:
Ok, good for review. Ther is an issue with one of the dependencies I think with the `bn` lib prevent it from compiling it on our test target. The code otherwise is good for review.

I put all the costs at the top of each file so that it is easier to grab and put into a config later down the line.

artob:
> [There] is an issue with one of the dependencies I think with the `bn` lib prevent it from compiling it on our test target.

When do you expect to fix that? I'd like to merge soonest, but wouldn't want to break the test target and CI.

joshuajbouw:
First thing that I'm doing tomorrow.

mrLSD (approved):

joshuajbouw:
This is good to go now @artob, though I'm going to double check something pertaining to the `target_gas`, if we need it or not. From what i understand, there should be another component that checks the gas.

joshuajbouw:
Going back to draft as I had realised it needs to accommodate for Berlin and others.

joshuajbouw:
@artob quite a large framework update, but only increased the size of the contract by 164 bytes. It uses one of my favourite trait marker + empty struct tricks in Rust which allows for basically 0 cost conditional compilations given the trait, so was quite excited to use it here.

@sept-en @mfornet take note as I know you have been implementing 2 new precompiles.

artob (approved):
LGTM :+1: Good job!

PR #23, opened 2021-04-06
Eth connector

mrLSD:


src/connector.rs
```diff
@@ -0,0 +1,330 @@
...
+            hex::encode(&event.eth_custodian_address),
+            hex::encode(&self.contract.eth_custodian_address),
+        ));
+
+        assert_eq!(
+            event.eth_custodian_address, self.contract.eth_custodian_address,
+            "ERR_WRONG_EVENT_ADDRESS",
+        );
+        assert!(
+            event.amount.sub(event.fee).as_u128() > 0,
```
birchmd, line 93: Wouldn't `event.amount > event.fee` be equivalent and more direct?

mrLSD: Good point. Changed.

src/fungible_token.rs
```diff
@@ -0,0 +1,368 @@
...
+                    sdk::panic_utf8(b"ERR_WRONG_AMOUNT");
+                }
+                _ => storage_balance,
+            }
+        } else {
+            sdk::panic_utf8(b"ERR_ACCOUNT_NOT_REGISTERED");
+        }
+    }
+
+    fn ft_key(&self, account_id: AccountId) -> Vec<u8> {
```
birchmd, line 349: Performance nit: I think most of the calls to `clone` could be removed if this function (and essentially all the other functions in this module) took `&str` instead of `AccountId`.

mrLSD: I think you mean reference instead of move value. It's a really small overhead. But I have got a lot of issues when I tried `&str`. And I tried it. But return String. So if it'll a real point it will possible to spend time on that. But IMHO for that concrete point it'll really small overhead.

birchmd: I agree, it is likely a small overhead. I don't think it should block the PR.

This is a change I would be happy to contribute though. I'm sure you have more important things to be doing with regards to this work, but I could help in this small way if you think that's ok.

mrLSD: Awesome! Really thanks! Yes, please.

mrLSD: Bit just keep in mind, that for now better place for changes is #24

birchmd: I have created https://github.com/aurora-is-near/aurora-engine/pull/29 which targets the branch #24 is working from.

PR #25, opened 2021-04-07
Fix: Makefile

birchmd:
* `make check` includes running tests, running `clippy`, checking the code formatting
* `make test` depends on release since it needs the `release.wasm` artifact

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ check-format:
 	$(CARGO) fmt -- --check
 
 clippy:
-	$(CARGO) clippy -- -D warnings
+	$(CARGO) +nightly clippy -- -D warnings
 
 clean:
 	@rm -Rf *.wasm target *~
--- a/Makefile
+++ b/Makefile
@@ -29,11 +29,17 @@ target/wasm32-unknown-unknown/debug/aurora_engine.wasm: Cargo.toml Cargo.lock $(
 deploy: release.wasm
 	$(NEAR) deploy --account-id=$(or $(NEAR_EVM_ACCOUNT),aurora.test.near) --wasm-file=$<
 
-check:
+check: test check-format
+
+# test depends on release since `tests/test_upgrade.rs` includes `release.wasm`
+test: release
 	$(CARGO) test
 
 format:
 	$(CARGO) fmt
+
+check-format:
+	$(CARGO) fmt -- --check
 
 clean:
 	@rm -Rf *.wasm target *~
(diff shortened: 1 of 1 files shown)


joshuajbouw (changes requested):
clippy & check on nightly catches quite a bit more than what is on stable right now. I would highly suggest that we do use it `+nightly`

birchmd:
@joshuajbouw  I'll add `+nightly` because it doesn't hurt either way. But just to confirm for myself, shouldn't the nightly `fmt` and `clippy` be used as is because that is what is specified in the `rust-toolchain` file?

birchmd:
Actually, I'll hold off on adding `+nightly` because I think it overrides the toolchain specified in `rust-toolchain`, which I don't think is what we want. @joshuajbouw let me know if you think differently.

joshuajbouw:
Ah right, I missed the rust-toolchain file there. Well noted. @birchmd

joshuajbouw (approved):
Looks good

joshuajbouw (changes requested):
Sorry, flipping back to request changes, just waking up and a bit bleary eyed. I do really believe that clippy & check should be latest nightly and not part of the locked toolchain due to the extra lint checks you get which can improve code quality further.

birchmd:
Thanks @joshuajbouw always happy to assist in the pursuit of better code. `+nightly` added in https://github.com/aurora-is-near/aurora-engine/pull/25/commits/fe10b5ec12d8d8fd0ac04d32db111cd273be1780

One thing I'd like you to clarify for my own understanding is how the `nightly` toolchain works. When I run `cargo +nightly clippy` does it automatically find and use the latest `clippy` build, or will I need to manually update my local `nightly` toolchain occasionally?

joshuajbouw:
Right, we already discussed this privately. You would need to manually update your local nightly toolchain.

joshuajbouw (approved):
Happy with this.

Makefile
```diff
@@ -29,12 +29,21 @@ target/wasm32-unknown-unknown/debug/aurora_engine.wasm: Cargo.toml Cargo.lock $(
...
 
 format:
 	$(CARGO) fmt
 
+check-format:
+	$(CARGO) fmt -- --check
+
+clippy:
+	$(CARGO) +nightly clippy -- -D warnings
+
```
artob, line 46: New targets like these need to be added to the `.PHONY` line as well.

artob (approved):

Makefile
artob: Fixed in 34d247a.

artob:
> I do really believe that clippy & check should be latest nightly and not part of the locked toolchain due to the extra lint checks you get which can improve code quality further.

I've merged the dependency on current nightly for now, but depending on how much that slows down our CI cycle we may need to revert that bit. (It was dog slow as it was.)

mrLSD (approved):

PR #26, opened 2021-04-08
Add twice weekly run of nightly clippy

birchmd:
@joshuajbouw pointed out that there is value in running the nightly clippy, not just the one for our fixed toolchain, because there are new lints being added. This PR adds a regularly scheduled GitHub action to run nightly clippy. The reason to make it run on a schedule instead of modifying the lints which run on pushes/PRs is because we don't want failures to randomly happen if there is a new lint which flags code that was not changed by the commit/PR. The reason to make it run at a strange time (instead of on an hour or half hour) is to avoid congestion in the GitHub actions infrastructure.

### DIFF
--- a/.github/workflows/scheduled_lints.yml
+++ b/.github/workflows/scheduled_lints.yml
@@ -1,0 +1,24 @@
+---
+on:
+  schedule:
+    - cron: '27 8 * * 1,4'
+name: Scheduled_Lints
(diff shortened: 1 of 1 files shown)


joshuajbouw (approved):
LGTM :D

artob (approved):
LGTM :+1:

PR #27, opened 2021-04-13
Fix: Use clippy on contract code too

birchmd:
Previously `clippy` was not being used to check code gated behind the `contract` feature flag. I assume this is because it was failing due to the presence of `on_panic` and `on_alloc_error`. However, we can only include those functions when we compile to wasm, then `clippy` will work with the `contract` feature flag.

In this PR we change the Makefile and CI to have `clippy` check the code with the `contract` feature enabled (and with no-std since that is how we compile the contract).

`clippy` found one error in `engine.rs` and two in `lib.rs`. These are also corrected in this PR.

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -38,4 +38,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- -D warnings
+          args: --no-default-features --features=contract -- -D warnings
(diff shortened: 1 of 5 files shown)


joshuajbouw (commented):
Wouldn't `--all-features --all-targets` be the better choice here then we don't have to update this per feature?

artob (approved):
LGTM :+1:

artob:
> Wouldn't `--all-features --all-targets` be the better choice here then we don't have to update this per feature?

The `contract` feature is not on par with other feature flags. It's basically a boolean toggle for compilation mode.

Issue #2, opened 2021-03-09
Add `begin_block` function to contract


frankbraun:
For the [evm-bully](https://github.com/aurora-is-near/evm-bully) we need a `begin_block` function.
The `begin_block` function should allow to set the necessary block context the `evm-bully` needs to replay transactions successfully (see https://github.com/aurora-is-near/evm-bully/issues/3).
It should only be available in a testing version of the contract.

artob:
@frankbraun Commits 3f15f0646c38ae841fa29ad0a6a9e5e73efe489f and 78c7c1f3d34b4da028cdf5f59a27bcf184494faf added the parameter definition for `begin_block`:

https://github.com/aurora-is-near/aurora-engine/blob/78c7c1f3d34b4da028cdf5f59a27bcf184494faf/src/types.rs#L50-L65

Note that the input to the `begin_block` function is Borsh-encoded, not JSON-encoded.

frankbraun:
This needs to be put behind a feature flag and, as @ilblackdragon noted, should only be callable by the owner.

artob:
> This needs to be put behind a feature flag and, as @ilblackdragon noted, should only be callable by the owner.

As of 529a751351ef8acf0360ef72c91a51eafb9ff01c, these methods are now only callable by the owner.

joshuajbouw:
Isn't this closed now?

artob:
@frankbraun Can this be closed? Please close issues when they are done.

frankbraun:
The `evm-bully` calls `begin_block` with the correct arguments and they are decoded in the function in the contract:

https://github.com/aurora-is-near/aurora-engine/blob/51442811006a8803d2bcaaf6239e36eb2fb4eaaf/src/lib.rs#L364

However, the engine state is not updated correctly yet. @joshuajbouw, could you take a stab at this please? You are deeper into the engine. Might be related to #122

joshuajbouw:
Yes, still looking into it.

mrLSD:
@frankbraun  Is that Issue still should be open?

joshuajbouw:
Considering that this functionality has been added and there are other issues that are in a succession of this, I'm going to close it.

Issue #1, opened 2021-03-09
Add `begin_chain` function to contract


frankbraun:
For the [evm-bully](https://github.com/aurora-is-near/evm-bully) we need a `begin_chain` function.
The function should allow to set the account balances in the genesis state and should only be available in a testing version of the contract.

artob:
@frankbraun Commit 526005ddbfcf198f469ad3ccdf81e64b45f5b098 added an initial parameter definition for `begin_chain`:

https://github.com/aurora-is-near/aurora-engine/blob/78c7c1f3d34b4da028cdf5f59a27bcf184494faf/src/types.rs#L44-L48

Note that the input to the `begin_chain` function is Borsh-encoded, not JSON-encoded.

I'm still working on figuring out a workable Borsh encoding for setting genesis block balances.

If more scalar parameters beyond `chain_id` are needed, add them in a PR.

frankbraun:
This needs to be put behind a feature flag and, as @ilblackdragon noted, should only be callable by the owner.

@artob: You mentioned that you already have some code for this?

artob:
> @artob: You mentioned that you already have some code for this?

I've restored the basic `begin_chain` functionality in 1458997. It had broken when #6 was merged.

joshuajbouw:
Isn't this closed now?

artob:
@frankbraun Can this be closed? Please close issues when they are done.

frankbraun:
Sorry, this got buried in GitHub notifications.

PR #30, opened 2021-04-14
Implement necessary contract method `begin_chain` for `evm-bully`.

frankbraun:
Issues #1 and #2.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -10,8 +10,8 @@ use crate::sdk;
 
 pub type AccountId = String;
 pub type RawAddress = [u8; 20];
-pub type RawU256 = [u8; 32];
-pub type RawH256 = [u8; 32];
+pub type RawU256 = [u8; 32]; // Little-endian large integer type.
+pub type RawH256 = [u8; 32]; // Unformatted binary data of fixed length.
 
 pub const STORAGE_PRICE_PER_BYTE: u128 = 100_000_000_000_000_000_000; // 1e20yN, 0.0001N
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -287,7 +287,12 @@ mod contract {
         let args = BeginChainArgs::try_from_slice(&input).expect("ERR_ARG_PARSE");
         state.chain_id = args.chain_id;
         Engine::set_state(state);
+
         // TODO: https://github.com/aurora-is-near/aurora-engine/issues/1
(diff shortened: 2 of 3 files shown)


frankbraun:
Most work related to this was done on on the [evm-bully](https://github.com/aurora-is-near/evm-bully) side.

Regarding issue #1 everything is implemented, issue #2 needs more work. With this PR we can now replay Görli up to block 12842 where we fail with:

```
Transaction Id B4cfxkbuPG2xjSUzbxzVY7WDVT2udaSXdyajDZ2DwBjn
{
  "Failure": {
    "ActionError": {
      "index": 0,
      "kind": {
        "FunctionCallError": {
          "ExecutionError": "Exceeded the prepaid gas."
        }
      }
    }
  }
}
```
I would merge this now and do further work with additional PRs.

joshuajbouw (approved):
Straight forward.

artob (approved):
LGTM :+1:

src/types.rs
```diff
@@ -10,8 +10,8 @@ use crate::sdk;
 
 pub type AccountId = String;
 pub type RawAddress = [u8; 20];
-pub type RawU256 = [u8; 32];
-pub type RawH256 = [u8; 32];
+pub type RawU256 = [u8; 32]; // Little-endian large integer type.
```
birchmd, line 13: Is it little endian? I thought EVM was always big endian

frankbraun: Yes, it's pulled from here: https://docs.rs/ethereum-types/0.1.4/ethereum_types/

frankbraun: Hmm, I'm finding conflicting info on this.

birchmd: My question was based on what I read in the Mastering Ethereum book

> All numbers in Ethereum are encoded as big-endian integers

https://github.com/ethereumbook/ethereumbook/blob/develop/06transactions.asciidoc

joshuajbouw: In blake2f, they use little endian on parts too, not all of it. *shrug*

frankbraun: @birchmd You were right with this. The documentation https://docs.rs/ethereum-types/0.1.4/ethereum_types/ confused me. Web assembly uses little-endian, but the EVM uses big-endian integers. So larger number were encoded wrongly in the `evm-bully` which also caused #122.

PR #37, opened 2021-04-21
feat: add `transfer` method to engine

### LINKED ISSUE #35: Implement balance transfers
In order to successfully replay transactions from the Görli testnet we need [balance transfers](https://github.com/aurora-is-near/aurora-engine/blob/c3430193229298e2aeeaa8f6f104d3ea5b1e167f/src/engine.rs#L165).

joshuajbouw:
Quite simply, adds a method, `transfer`, to transfer a balance from one to another to the `Engine`.

Comes with additional added methods which will allow others to easily increase or decrease a balance, `increase_balance` and `decrease_balance`.

Closes #35.


artob (approved):
LGTM :+1:

PR #38, opened 2021-04-21
chore: change deprecated `u64::max_value` to `u64::MAX`

joshuajbouw:
Really straight forward, `max_value` was flagged as deprecated, this changes it to the suggested new `const MAX`.


artob (approved):
LGTM :+1:

Issue #11, opened 2021-03-28
Implement a faucet method


artob:
As discussed on the [2021-03-26 weekly update](https://github.com/near/nearcore/discussions/4172).

PR #39, opened 2021-04-21
feat: faucet

joshuajbouw:
This adds a faucet with the method `make_it_rain` including improvements tackled that were suggested from the previous PR #37.

Please look and verify, I still am getting used to how we do these but I think this is correct.

Closes: #11

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -5,7 +5,7 @@ use evm::{Config, CreateScheme, ExitError, ExitReason, ExitSucceed};
 
 use crate::parameters::{FunctionCallArgs, NewCallArgs, ViewCallArgs};
 use crate::precompiles;
-use crate::prelude::{Address, Vec, H256, U256, Borrowed};
+use crate::prelude::{Address, Borrowed, Vec, H256, U256};
 use crate::sdk;
 use crate::storage::{address_to_key, storage_to_key, KeyPrefix};
 use crate::types::{bytes_to_hex, log_to_bytes, u256_to_arr, AccountId};
@@ -129,10 +129,12 @@ impl Engine {
     /// * If the balance is > `U256::MAX`
     fn check_increase_balance(address: &Address, amount: &U256) -> Result<U256, ExitError> {
         let balance = Self::get_balance(address);
-        if let Some (new_balance) = balance.checked_add(*amount) {
+        if let Some(new_balance) = balance.checked_add(*amount) {
             Ok(new_balance)
         } else {
-            Err(ExitError::Other(Borrowed("balance is too high, can not increase")))
+            Err(ExitError::Other(Borrowed(
+                "balance is too high, can not increase",
+            )))
         }
     }
 
@@ -148,7 +150,9 @@ impl Engine {
         if let Some(new_balance) = balance.checked_sub(*amount) {
             Ok(new_balance)
         } else {
-            Err(ExitError::Other(Borrowed("balance is too low, can not decrease")))
+            Err(ExitError::Other(Borrowed(
+                "balance is too low, can not decrease",
+            )))
         }
     }
 
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -121,7 +121,6 @@ impl Engine {
     }
 
     /// Increases the balance for a given address.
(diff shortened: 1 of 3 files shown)


src/engine.rs
```diff
@@ -121,17 +121,25 @@ impl Engine {
     }
 
     /// Increases the balance for a given address.
-    pub fn increase_balance(address: &Address, amount: &U256) {
-        let mut balance = Self::get_balance(address);
-        balance += *amount;
-        Self::set_balance(address, &balance);
+    fn increase_balance(address: &Address, amount: &U256) -> Result<(), ExitError> {
+        let balance = Self::get_balance(address);
+        let (new_balance, overflow) = balance.overflowing_add(*amount);
```
birchmd, line 126: Thanks for adding the overflow checks! I was going to comment on the original transfer PR, but then saw you had already corrected it here.

I personally [prefer `checked_add`](https://docs.rs/primitive-types/0.9.0/primitive_types/struct.U256.html#method.checked_add) because I think the `Option` type captures our use case a little better (when overflow happens there is no meaningful value), but I also recognize it's mostly a style choice.

src/engine.rs
```diff
@@ -121,17 +121,25 @@ impl Engine {
...
+        if overflow {
+            return Err(ExitError::StackOverflow);
```
birchmd, line 128: This is not a stack overflow, it is just arithmetic overflow. It doesn't look like there is an explicit enum variant which supports this error, so I guess the right thing to do would be to use `ExitError::Other`.

src/engine.rs
```diff
@@ -121,17 +121,25 @@ impl Engine {
...
     /// Decreases the balance for a given address.
-    pub fn decrease_balance(address: &Address, amount: &U256) {
-        let mut balance = Self::get_balance(address);
-        balance -= *amount;
-        Self::set_balance(address, &balance);
+    fn decrease_balance(address: &Address, amount: &U256) -> Result<(), ExitError> {
+        let balance = Self::get_balance(address);
+        let (new_balance, underflow) = balance.overflowing_sub(*amount);
+        if underflow {
+            return Err(ExitError::StackUnderflow);
```
birchmd, line 139: This is not a stack underflow, I think we should use `ExitError::Other` (see also above comment).

src/engine.rs
joshuajbouw: That works.

src/engine.rs
joshuajbouw: Ah, sorry was late, STACK overflow, of course.

src/lib.rs
```diff
@@ -232,6 +232,16 @@ mod contract {
         process_exit_reason(status, &result);
     }
 
+    #[cfg(feature = "testnet")]
+    #[no_mangle]
+    pub extern "C" fn make_it_rain() {
```
birchmd, line 237: This method makes the faucet available if you have a Near account, but what if you only have an EVM address? Should we also make the faucet available via an ethereum transaction? (I'm not sure how to do this to be honest, but it seems like a good thing for UX if we think not all users will also have Near accounts)

joshuajbouw: Good point, for that, I'm not entirely sure. Wouldn't that need to be built into the ETH contract itself?

birchmd: One way we could do this would be to have an faucet eth contract (i.e. contract inside the EVM) and we would just have to regularly fund it via `make_it_rain`. This could probably even be done with some automation (something watches the balance of the contract and when it gets too low submits a `make_it_rain` transaction to Near).

Is this how faucets work on Ethereum testnets? Do they have a normal contract with a large balance, and when it gets low they refill it using some admin magic?

joshuajbouw: I've written and worked with a few UTXO coin faucets, but not one with an ETH smart contract. My understanding is that they create the balance, and then you can call a method which transfers coins from the contract to the user.

artob: We'll have a TestNet faucet inside the EVM as well, but haven't got down the task list quite that far yet.

artob (approved):
LGTM :+1:

PR #40, opened 2021-04-22
feat: return formatted custom messages

joshuajbouw:
Really small quality of life improvement which will just simply return the error messages contained in `Custom` option.


artob:
@joshuajbouw Looking good, but please check the Clippy errors

artob (approved):
LGTM :+1:

PR #41, opened 2021-04-22
Aurora Engine Benchmarks

### LINKED ISSUE #31: Collect benchmarking data for runtime improvements
## References
- https://near-protocol.slack.com/archives/C0173NU5QJC/p1618229606100100 (nonpublic)

birchmd:
The Near runtime has the following requirements:

1. Regardless of what work is done in a transaction (computation, io, etc), Near gas used should always be proportional to the wall clock execution time
2. This proportionality should be such that 1000 Tgas can be consumed in < 1s (otherwise Near's blocktime can be thrown off)

The purpose of this PR is to create benchmarks involving the EVM wasm contract to test these requirements hold.

Wall-clock time is an inherently volatile metric because many factors contribute to it, so results from this  should be taken with a grain of salt; though great deviations from expectations should be investigated more closely.

The EVM contract is executed within the [`near-vm-runner`](https://github.com/near/nearcore/tree/master/runtime/near-evm-runner). This also records the Near gas spent on the computation. The wall clock measurements are done using [`criterion`](https://crates.io/crates/criterion).

Two work loads are considered:

* Transferring eth between two accounts
* Deploying an eth smart contract (of various sizes)

Note: these work loads are done as eth transactions sent to the EVM contract, so both also involve checking the validity of the transaction which includes a Keccak256 hash computation and Secp256k1 signature check. (In fact, this is likely the bulk of the work involved).

Using the timing measurements from `criterion` and the gas measurements from the `near-vm-runner`, we can measure the relationship between wall-clock time and gas used.

Known limitation: all state is kept in memory, so IO time may not be accurately accounted for.

Closes #31


birchmd:
I think this is in good enough shape to review (and merge if there are no changes requested). There are a couple remaining TODO items that I have captured as issues, but they are not blocking to this PR in my opinion.

This PR is important because it contains some of the only tests we have for some features. Creating the PR even revealed some bugs which could then be fixed (https://github.com/aurora-is-near/aurora-engine/pull/42 https://github.com/aurora-is-near/aurora-engine/pull/44). We will focus more on testing in the future as well of course.

The original impetus for this PR was to benchmark the Near runtime using the EVM contract, as well as get a sense for the cost of common EVM operations. the results of this are summarized in the plot below (note: I've called out a few points of interest in orange, but all the points are present from the benchmarks in this PR):

aurora_benchmark

Some takeaways from this:

- Due to signature checking currently happening in wasm itself (instead of being a host function in the runtime), everything is very expensive in terms of Near gas. This makes getting the new host functions in place very important for usability. https://github.com/near/nearcore/pull/3954
- The points generally lie on a straight line, which is good for our assumption that time is proportional to gas usage. However, the slope of the line looks too steep to me. It means it would take ~1.3 seconds to consume 1000 Tgas, which I think is too long based on Near's target block time (please correct me if I am wrong cc @bowenwang1996 @ailisp @olonho)
- There is some deviation from the straight line at the top end gas usage, which I am not sure how to reconcile.

joshuajbouw:
Interesting. Shoot. I didn't expect them to be so high. That really adds a lot more importance on https://github.com/near/nearcore/pull/3954.

djsatok:
@birchmd can you please create a plot NEAR Gas vs. Ethereum Gas. This will help us to determine the recalculation formula + estimate what size of a transaction we can accommodate within 300 Tgas.

benches/eth_deploy_code.rs
```diff
@@ -0,0 +1,70 @@
...
+    let mut rng = rand::thread_rng();
+    let source_account = SecretKey::random(&mut rng);
+    runner.create_address(
+        address_from_secret_key(&source_account),
+        INITIAL_BALANCE.into(),
+        INITIAL_NONCE.into(),
+    );
+    let inputs: Vec<_> = [1, 4, 8, 12, 16]
+        .iter()
+        .copied()
```
joshuajbouw, line 23: Minor note, since these are digits, they don't need to be copied in this way as they can be de-referenced:

 ```rust
     let inputs: Vec<_> = [1, 4, 8, 12, 16]
        .iter()
        .map(|n| {
            let code_size = 2usize.pow(*n);
            let code: Vec<u8> = std::iter::repeat(0).take(code_size).collect();
            let transaction = create_eth_transaction(
                None,
                TRANSFER_AMOUNT.into(),
                code,
                Some(runner.chain_id),
                &source_account,
            );
            rlp::encode(&transaction).to_vec()
        })
        .collect();
```

**Review by joshuajbouw — APPROVED** [review]
Looks really good. Nothing to really note or point out from me. Good work.

File: benches/eth_deploy_code.rs (hunk 1 above)
Comment (birchmd) [reply benches/eth_deploy_code.rs:23]: Dereferencing is equivalent to copying though right? I.e.

```rust
[...].iter().copied()
```

produces the same instructions on compilation as

```rust
[...].iter().map(|n| *n)
```

?

birchmd:
@djsatok The graph you asked for is below:

eth_gas_benchmark

The correlation here is not very strong. The differences seem to be that we charge much more for pre-compiles, and much less for deploying code. Both of these make sense to me. Our "pre-compiles" are actually executing in wasm right now, so the gas cost is far off what it should be; again pointing to the importance of https://github.com/near/nearcore/pull/3954 . And I think Near storage is supposed to be much cheaper in terms of gas than on Eth because of the "storage staking" concept.

The Eth transfer and ERC20 transfer both happen to lie close-ish to the linear regression line, so maybe it could be used to give a rough estimate of the relationship between Near and Eth gas (obviously it totally does not work in the extreme cases of transactions that are all storage or all pre-compiles). Using that measure, you should be able to fit approximately 230,000 Eth gas into 300 Near Tgas.

benches/eth_deploy_code.rs
matklad: There are three equivalent ways to write this (in order of my personal preference): 

* `iter().copied().map(|n| ...)` -- clearly factors copying out, so that it's not distracing
* `iter().map(|&n| ...)` -- dereference once in in a pattern
* `iter().map(|n| *n)` -- dereference every time n is used

Perf-wise, they are all equivalent.

birchmd: Thanks @matklad !

bowenwang1996:
> and much less for deploying code

This will be changed once we land https://github.com/near/nearcore/pull/4231

benches/eth_deploy_code.rs
```diff
@@ -0,0 +1,78 @@
...
+        .map(|n| {
+            let code_size = 2usize.pow(n);
+            let code: Vec<u8> = std::iter::repeat(0).take(code_size).collect();
```
matklad, line 26:
```suggestion
            let code: Vec<u8> = vec![0; code_size];
```

benches/eth_erc20.rs
```diff
@@ -0,0 +1,269 @@
...
+use aurora_engine::transaction::EthTransaction;
+use criterion::{criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion};
+use near_vm_logic::VMOutcome;
+use secp256k1::SecretKey;
+use std::path::{Path, PathBuf};
+
+// We don't use everything in `common`, but that's ok, other benchmarks do
+#[allow(dead_code)]
+mod common;
+mod solidity;
```
matklad, line 11: I suggest using the following layout for benchmarks:

```
benches/
  main.rs
  eth_erc20.rs
  ...
```

where `main.rs` looks like this:

```
mod eth_erc20;
...

Stuff which is currently in common.rs
```

and with the following Cargo.toml:

```
[package]
...
autobenches = false

[[bench]]
name = "benches"
path = "benches/main.rs"
```

That way, you'll have only single integration binary compiled, which should save you a lot of time on linking, considering the fact that benchmarks are compiled in release mode. As a side-effect, you won't need `allow(deadcode)`. 

See https://github.com/near/nearcore/pull/4027 for details.

src/engine.rs
```diff
@@ -318,6 +326,16 @@ impl Engine {
...
 }
 
+#[cfg(feature = "profile_eth_gas")]
+fn log_eth_gas_used(executor: &StackExecutor<MemoryStackState<Engine>>) {
+    use alloc::format;
+    sdk::log(format!(
+        "{}: {}",
+        crate::prelude::ETH_GAS_USED,
+        executor.used_gas()
+    ));
```
matklad, line 336: Feels a bit strange that it's `pub fn log(data: String) {` rather than `pub fn log(data: &str) {`

benches/eth_erc20.rs
birchmd: How does this interact with the `criterion_group` and `criterion_main` macros I'm using at the bottom of each file? I assume I somehow want to say in `main.rs` that the benchmarks I want to run are contained in the various sub-modules.

matklad: Good question! I haven't used criterion myself, but I think the following should work:

```rust
// benches/main.rs
mod eth_erc20;
mod eth_deploy_code;

criterion_group!(benches, eth_erc20::eth_erc20_benchmark, eth_deploy_code::deploy_code_benchmark);
criterion_main!(benches);

// benches/eth_erc20.rs

pub fn eth_erc20_benchmark(c: &mut Criterion) { ... }
```

Both group and main accept paths: https://docs.rs/criterion/0.3.4/criterion/macro.criterion_group.html

Alternatively, you can define a separate `criterion_group` in every submodule, and register all groups with criterion_main in main.rs, but it seems that you have one bench for submodule, so this isn't really neceesary.

artob:
@birchmd Assigning this back to you to wrap up review discussions. Please assign to me once ready to merge.

benches/eth_deploy_code.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/41/commits/4fb6f98cad83c948c7f07d741b46426f4a39f48c

src/engine.rs
birchmd: Fixed in https://github.com/aurora-is-near/aurora-engine/pull/41/commits/58e0cc479a92e2aabd6beadf27087a5c5fef74eb

joshuajbouw (approved):
New changes LGTM.

artob:
> Note: there will need to be a follow-up PR after #48 lands because it changes the output format of the contract, and I rely on the output to learn the address of deployed contracts (e.g. ERC-20). So once that PR is merged we'll need to fix the benchmarks to consume the new output format. I can take care of that though, should be quite small.

@birchmd I've merged #48 just now. Please update this as needed and then re-assign back to me for final review and merge.

joshuajbouw (approved):
New changes look fine.

birchmd:
@artob should be good to go now

artob (approved):
LGTM :+1: Really great work, @birchmd!

bowenwang1996:
@birchmd any idea what the benchmark would be if the precompiles are executed as host functions? I am not sure what host functions are needed, but I believe some are already available on nightly protocol.

birchmd:
@bowenwang1996 
@djsatok also asked me to get an estimate with precompiles as host functions. I thought the biggest contribution to current gas costs was probably `ecrecover`, so I ran a benchmark with that one as a host function using the work from https://github.com/near/nearcore/pull/3954 . With that change, an eth transfer drops to around 1Tgas and and erc20 transfer drops to around 15 Tgas. In short, it was a very big improvement.

I didn't run all the benchmarks or re-generate the Tgas vs time graph, but I can do that if you are interested.

The code with `ecrecover` integrated as a host function is in a branch on my fork of the repo. https://github.com/birchmd/aurora-engine/tree/evm-benchmark-with-math-api

bowenwang1996:
@birchmd I see. Do we need any other host functions that have not been implemented in nearcore yet?

birchmd:
@bowenwang1996 As far as I know everything in that nearcore PR is needed. I'm not sure how often each precompile is used in real Eth contracts, so I don't know if some could be considered higher priority than others. But I do know that the EVM contract itself uses `ecrecover` to validate transactions it receives which is why it is very important it can be executed efficiently.

I don't know if there are other host functions which are needed and not implemented. Maybe @artob or @joshuajbouw would know if there is.

PR #42, opened 2021-04-23
Feat: Check and increment nonce

birchmd:
To prevent replaying transactions, we need to check the nonce matches the nonce of the account, and increment it if it matches.

In the case of directly calling `call` or `deploy_code` replaying the transaction is not a concern because it originates from Near, and so is protected by the base system, but I still increment the nonce in those cases to be consistent.

One issue with this is the case where the transaction runs out of Near gas. In that case the state changes are rolled back, including the incremented nonce, which is different behaviour from if the transaction ran on Ethereum directly. I think the best way to handle this would be to make running out of Near gas impossible by setting appropriate limits on the Eth gas. But this probably is not feasible in the short term. We might need to leave this as a known bug for now. I'm interested in what others think about this issue @artob @djsatok @joshuajbouw @mrLSD

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -138,9 +138,7 @@ mod contract {
     #[no_mangle]
     pub extern "C" fn deploy_code() {
         let input = sdk::read_input();
-        let origin = predecessor_address();
-        Engine::increment_nonce(&origin).sdk_unwrap();
-        let mut engine = Engine::new(origin);
+        let mut engine = Engine::new(predecessor_address());
         let (status, address) = Engine::deploy_code_with_input(&mut engine, &input);
         // TODO: charge for storage
         process_exit_reason(status, &address.0)
@@ -151,9 +149,7 @@ mod contract {
     pub extern "C" fn call() {
         let input = sdk::read_input();
         let args = FunctionCallArgs::try_from_slice(&input).expect("ERR_ARG_PARSE");
-        let origin = predecessor_address();
-        Engine::increment_nonce(&origin).sdk_unwrap();
-        let mut engine = Engine::new(origin);
+        let mut engine = Engine::new(predecessor_address());
         let (status, result) = Engine::call_with_args(&mut engine, args);
         // TODO: charge for storage
         process_exit_reason(status, &result)
@@ -186,7 +182,8 @@ mod contract {
             None => sdk::panic_utf8(b"ERR_INVALID_ECDSA_SIGNATURE"),
         };
 
-        Engine::check_nonce(&sender, signed_transaction.transaction.nonce).sdk_unwrap();
+        let next_nonce =
+            Engine::check_nonce(&sender, &signed_transaction.transaction.nonce).sdk_unwrap();
 
         // Figure out what kind of a transaction this is, and execute it:
         let mut engine = Engine::new_with_state(state, sender);
@@ -194,7 +191,10 @@ mod contract {
         let data = signed_transaction.transaction.data;
         if let Some(receiver) = signed_transaction.transaction.to {
             let (status, result) = if data.is_empty() {
-                // Execute a balance transfer:
+                // Execute a balance transfer. We need to save the incremented nonce in this case
+                // because it is not handled internally by the SputnikVM like it is in the case of
+                // `call` and `deploy_code`.
+                Engine::set_nonce(&sender, &next_nonce);
                 (
                     Engine::transfer(&mut engine, &sender, &receiver, &value),
                     vec![],
@@ -229,7 +229,7 @@ mod contract {
             }
         };
 
-        Engine::check_nonce(&meta_call_args.sender, meta_call_args.nonce).sdk_unwrap();
+        Engine::check_nonce(&meta_call_args.sender, &meta_call_args.nonce).sdk_unwrap();
 
         let mut engine = Engine::new_with_state(state, meta_call_args.sender);
         let (status, result) = engine.call(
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -366,4 +375,26 @@ mod contract {
...
+            match self {
+                Self::NonceOverflow => "ERR_NONCE_OVERFLOW",
+                Self::IncorrectNonce => "ERR_INCORRECT_NONCE",
+            }
+        }
+    }
+
+    trait OrSdkPanic<T, E> {
+        fn unwrap_or_sdk_panic(self) -> T;
+    }
```
joshuajbouw, line 390: I think a better name for this, given Rust conventions, would simply be `sdk_unwrap` then at least it will be clear that it behaves like a standard unwrap, but for the SDK.

src/types.rs
```diff
@@ -35,6 +35,14 @@ pub enum ErrorKind {
...
 }
 
+/// Errors involving the nonce
+pub enum NonceError {
+    /// Attempted to increment the nonce, but overflow occurred
+    NonceOverflow,
+    /// Account nonce did not match the transaction nonce
+    IncorrectNonce,
+}
+
```
joshuajbouw, line 45: I'm personally a big fan of keeping errors limited to the struct they are in. So in this case, it might be better to change all errors to an `enum` called `EngineError` to generalise it instead of having quite a bit of boutique errors.

src/lib.rs
```diff
@@ -366,4 +375,26 @@ mod contract {
             }
         }
     }
+
+    impl ToStr for crate::types::NonceError {
+        fn to_str(&self) -> &str {
+            match self {
+                Self::NonceOverflow => "ERR_NONCE_OVERFLOW",
+                Self::IncorrectNonce => "ERR_INCORRECT_NONCE",
```
joshuajbouw, line 383: Speaking of this, what is our typical Error string? In Rust, you typically would have it as all lower case as when it panics on screen, the error reads out like a grammatical sentence where as if it is a panic, it would be a written sentence. Of course with the SDK it probably is something different and spits out to a log somewhere, but I never understood why we would detract from that as it becomes a new thing for Rust developers to pick up on, which we do differently than every other Rust lib.

joshuajbouw (changes requested):

src/lib.rs
artob: @joshuajbouw These are not error strings, they are error codes. It isn't humans interacting with the contract, it is machines. Libraries (such as Aurora.js) and tools (such as the Aurora CLI) will need to translate the codes into error strings; the contract's error handling should merely focus on conciseness and unambiguity in parsing flagged error conditions.

src/types.rs
birchmd: I like when a function's return types are constrained. I think it makes reasoning about the possible errors easier because I don't first need to decide which enum variants are even possible to be returned.

Upstream where error types need to be combined you can always nest the enums

```rust
enum SuperError {
    SubError1(crate::sub_error1::Error),
    // ...
}
```

I understand that nested enums can be a little painful to match on, so another option would be to have a mapping into something like the [`ErrorKind` enum](https://github.com/aurora-is-near/aurora-engine/blob/master/src/types.rs#L31) (a flat, simple structure) we already have

```rust
impl From<SpecificError> for ErrorKind {
    // ...
}
```

All that said, I don't have too strong of a feeling about this, so if you feel very strongly about not having specific error enums, then I will just add variants to [`ErrorKind`](https://github.com/aurora-is-near/aurora-engine/blob/master/src/types.rs#L31) and use that.

joshuajbouw: Sure yes, that was the other part of that, having enums wrapping more granular errors. They are a bit of a pain, it is a bit of boilerplate, but its a good structure.

joshuajbouw:
If in a usual NEAR contract execution, the nonce being rolled back on failure, I think that should be fine to continue with, as long as it doesn't hinder any data between ETH. But if otherwise, a sort of mandatory cost is a good idea, if thats the only way to solve that.

src/types.rs
joshuajbouw: As per discussion, since it is separate, internal, and what not, sure its fine to leave as is.

src/lib.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/42/commits/d93d0a88a70c2ca33c6d8c4ad02427848af646f3

birchmd:
> If in a usual NEAR contract execution, the nonce being rolled back on failure

It's not that on the Near base runtime nonces are always rolled back, that is not the case. The problem here is that we track the nonce explicitly as part of the EVM contract state, and contract state changes are rolled back on OOG (this is true on Eth as well of course). So when we run out of Near gas, the EVM contract state changes are rolled back (it is just a normal contract on Near after all), including the nonce change.

A more "near native" solution than what I suggested above (limiting the Eth gas so that we can never run out of Near gas), would be to take advantage of the Near asyc runtime. We could save the nonce change as a transaction first, then execute the EVM call as an async call (receipt) afterwards. So if the receipt runs out of gas, then only those state changes will be rolled back, not the nonce change since it was already committed. One drawback of this solution is that EVM transactions would then always be split over two Near blocks, which could be awkward for tooling.

joshuajbouw:
The first solution in the description seems the most favourable for me, and likely will be the solution we will require when we go to main net. The two transaction with the first being the nonce does seem quite awkward for the reasons you posted.

I say, lets leave it open ended for now, be aware of it, and ensure we track it before main net.

joshuajbouw (approved):

artob (approved):
LGTM :+1: Nice job!

PR #44, opened 2021-04-27
Fix Precompiles: RIPEMD160, modexp

birchmd:
Fixes two small issues with the precompiles:

1. In RIPEMD160 the result needs to be padded out to 32 bytes
2. In modexp the result needs to be padded out to the same length as the modulus


artob (approved):
LGTM :+1: Good catch!

PR #47, opened 2021-04-28
fix: gas_limit

joshuajbouw:
This is pretty short, I reviewed the code for the backend implementation and checked and diagnosed where we may have issues. The only real issue that I had found is that we likely have an issue if the `gas_limit` returns 0. This was changed to max value and docs were added in. The other change is the `block_coinbase` returning the first 2^160 of the keccak256 hash of `aurora`.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -335,8 +335,16 @@ impl evm::backend::Backend for Engine {
         U256::from(sdk::block_index())
     }
 
+    /// Returns a mocked coinbase which is the first 160 bits of
+    /// `keccak256(b"aurora")`.
+    ///
+    /// It is not possible to return the address of the current block's miner in
+    /// NEAR as it isn't relevant.
     fn block_coinbase(&self) -> Address {
-        Address::zero()
+        Address([
+            0x2b, 0x0b, 0xf3, 0xb8, 0xff, 0xaa, 0x4f, 0x3d, 0x1f, 0x97, 0x76, 0x0d, 0x44, 0x44,
+            0x58, 0x84, 0x43, 0xc3, 0xa9, 0x12,
+        ])
     }
 
     fn block_timestamp(&self) -> U256 {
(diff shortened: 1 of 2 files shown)


src/engine.rs
```diff
@@ -319,61 +319,91 @@ impl Engine {
...
     }
 
+    /// Returns the current block difficulty.
     fn block_difficulty(&self) -> U256 {
         U256::zero()
     }
 
+    /// Returns the current block's gas limit.
+    ///
+    /// Currently, this returns 0 as there is no concept of a gas limit.
```
artob, line 371: The inline docs don't match the code here.

src/engine.rs
```diff
@@ -319,61 +319,91 @@ impl Engine {
...
+    /// `keccak256(b"aurora")`.
+    ///
+    /// It is not possible to return the address of the current block's miner in
+    /// NEAR as it isn't relevant.
     fn block_coinbase(&self) -> Address {
-        Address::zero()
+        Address([
+            0x2b, 0x0b, 0xf3, 0xb8, 0xff, 0xaa, 0x4f, 0x3d, 0x1f, 0x97, 0x76, 0x0d, 0x44, 0x44,
+            0x58, 0x84, 0x43, 0xc3, 0xa9, 0x12,
+        ])
```
artob, line 356: This is incorrect. The EVM address for the `aurora` account is _0x4444588443C3a91288c5002483449Aba1054192b_.

The EVM address derivation does not use the initial bytes. Relevant references:
- https://doc.aurora.dev/develop/compat/evm#coinbase
- https://github.com/aurora-is-near/aurora.js/blob/7f7bc70/src/account.ts#L24

artob (changes requested):

joshuajbouw:
Both fixed @artob, good to go.

artob (approved):
LGTM now :+1:

PR #48, opened 2021-04-30
fix: logs

joshuajbouw:
There is an issue with returning logs from the EVM currently. It is currently sending the logs out on the SDK but after a discussion with @artob we should instead return them back upon execution. This does exactly that, and omits the logs from being applied to the NEAR SDK.

- [x] Implement new logs
- [x] Find and include execution result

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -404,10 +404,7 @@ impl Engine {
         // allows a return of UTF-8 strings.
         self.apply(values, Vec::<Log>::new(), true);
 
-        let mut res_logs = Vec::new();
-        for log in logs {
-            res_logs.push(log.into());
-        }
+        let res_logs = logs.into_iter().map(Into::into).collect();
 
         let res = match status {
             ExitReason::Succeed(_) | ExitReason::Revert(_) => Some( SubmitResult {
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -4,7 +4,7 @@ use evm::executor::{MemoryStackState, StackExecutor, StackSubstateMetadata};
 use evm::ExitFatal;
 use evm::{Config, CreateScheme, ExitError, ExitReason};
 
-use crate::parameters::{FunctionCallArgs, NewCallArgs, ViewCallArgs, RawCallResult, ResultLog};
+use crate::parameters::{FunctionCallArgs, NewCallArgs, ViewCallArgs, SubmitResult};
 use crate::precompiles;
 use crate::prelude::{Address, Vec, H256, U256};
 use crate::sdk;
@@ -366,18 +366,36 @@ impl Engine {
         );
         maybe_error(status, &result.0)?;
         let (values, logs) = executor.into_state().deconstruct();
-        self.apply(values, logs, true);
-        Ok(result)
+        self.apply(values, Vec::<Log>::new(), true);
+
+        let mut res_logs = Vec::new();
+        for log in logs {
+            res_logs.push(log.into());
+        }
+
+        let (res_status, used_gas) = match status {
+            ExitReason::Succeed(_) => (true, used_gas),
+            ExitReason::Revert(_) => (false, used_gas),
+            ExitReason::Error(_) | ExitReason::Fatal(_) => (false, 0),
+        };
+
+        let res = SubmitResult {
+            status: res_status,
+            gas_used: used_gas,
+            logs: res_logs,
+            result: result.0.to_vec(),
+        };
+        (status, res)
     }
 
     pub fn call_with_args(&mut self, args: FunctionCallArgs) -> EngineResult<Vec<u8>> {
         let origin = self.origin();
         let contract = Address(args.contract);
         let value = U256::zero();
-        self.call(origin, contract, value, args.input)
+        self.submit(origin, contract, value, args.input)
     }
 
-    pub fn call(
+    pub fn submit(
         &mut self,
         origin: Address,
         contract: Address,
@@ -386,7 +404,7 @@ impl Engine {
     ) -> EngineResult<Vec<u8>> {
         let mut executor = self.make_executor();
         let (status, result) = executor.transact_call(origin, contract, value, input, u64::MAX);
-        maybe_error(status, &result)?;
+        let used_gas = executor.used_gas();
         let (values, logs) = executor.into_state().deconstruct();
         // There is no way to return the logs to the NEAR log method as it only
         // allows a return of UTF-8 strings.
@@ -397,9 +415,15 @@ impl Engine {
             res_logs.push(log.into());
         }
 
-        let res = RawCallResult {
-            status: false,
-            gas_used: executor.used_gas(),
+        let (res_status, used_gas) = match status {
+            ExitReason::Succeed(_) => (true, used_gas),
+            ExitReason::Revert(_) => (false, used_gas),
+            ExitReason::Error(_) | ExitReason::Fatal(_) => (false, 0),
+        };
+
+        let res = SubmitResult {
+            status: res_status,
+            gas_used: used_gas,
             logs: res_logs,
             result,
         };
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -371,7 +371,7 @@ impl Engine {
         let res_logs = logs.into_iter().map(Into::into).collect();
 
         let res = match status {
-            ExitReason::Succeed(_) | ExitReason::Revert(_) => Some( SubmitResult {
+            ExitReason::Succeed(_) | ExitReason::Revert(_) => Some(SubmitResult {
                 gas_used: used_gas,
                 logs: res_logs,
                 result: result.0.to_vec(),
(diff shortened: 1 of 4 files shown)


src/engine.rs
```diff
@@ -251,37 +251,77 @@ impl Engine {
...
+        let used_gas = executor.used_gas();
         let (values, logs) = executor.into_state().deconstruct();
-        self.apply(values, logs, true);
-        (status, result)
+        self.apply(values, Vec::<Log>::new(), true);
+
+        let mut res_logs = Vec::new();
+        for log in logs {
+            res_logs.push(log.into());
+        }
```
birchmd, line 268:
```suggestion
        let res_logs = logs.into_iter().map(Into::into).collect();
```

src/engine.rs
```diff
@@ -251,37 +251,77 @@ impl Engine {
...
+
+        let (res_status, used_gas) = match status {
+            ExitReason::Succeed(_) => (true, used_gas),
+            ExitReason::Revert(_) => (false, used_gas),
+            ExitReason::Error(_) | ExitReason::Fatal(_) => (false, 0),
```
birchmd, line 273: Is it true that the gas used is 0 in the case of `Error`? That doesn't seem right to me. For example in the case of an out-of-gas error all gas was spent.

src/engine.rs
```diff
@@ -251,37 +251,77 @@ impl Engine {
...
+            status: res_status,
+            gas_used: used_gas,
+            logs: res_logs,
+            result: result.0.to_vec(),
+        };
+        (status, res)
     }
 
-    pub fn call_with_args(&mut self, args: FunctionCallArgs) -> (ExitReason, Vec<u8>) {
+    pub fn submit_with_args(&mut self, args: FunctionCallArgs) -> (ExitReason, SubmitResult) {
```
birchmd, line 285: I don't think I have a strong opinion either way, but I'm just curious, what is better about the name `submit` vs `call`?

src/engine.rs
```diff
@@ -251,37 +251,77 @@ impl Engine {
...
+            ExitReason::Revert(_) => (false, used_gas),
+            ExitReason::Error(_) | ExitReason::Fatal(_) => (false, 0),
+        };
+
+        let res = SubmitResult {
+            status: res_status,
+            gas_used: used_gas,
+            logs: res_logs,
+            result,
+        };
```
birchmd, line 323: Looks like a pretty big section of code duplication with `deploy_code`. Maybe it could be extracted out into a function and used in both places.

src/lib.rs
```diff
@@ -139,9 +141,12 @@ mod contract {
     pub extern "C" fn deploy_code() {
         let input = sdk::read_input();
         let mut engine = Engine::new(predecessor_address());
-        let (status, address) = Engine::deploy_code_with_input(&mut engine, &input);
+        let (status, result) = Engine::deploy_code_with_input(&mut engine, &input);
         // TODO: charge for storage
-        process_exit_reason(status, &address.0)
+        process_exit_reason(
+            status,
+            &result.try_to_vec().expect("serialize can not fail"),
```
birchmd, line 148: I think we should try to move away from the built-in `expect`/`unwrap` calls on `Result` and `Option` in favour of ones that use the `sdk::panic` call.

It doesn't make a big difference right now because our binary is largely bloated by secp256 data, but if we want to optimize code size in the future then avoiding the rust panic is a good step.

Also, in the spirit of https://github.com/aurora-is-near/aurora-engine/pull/42#discussion_r620259522 this should probably be something like `ERR_SERIALIZE` or something like that. And it should probably be factored out as a constant since it is repeated in multiple places.

birchmd (changes requested):

src/engine.rs
joshuajbouw: The `ExitReason::Error` and `ExitReason::Fatal` are non-opcode related, and the `used_gas` is returned to signify how much gas was spent as for both `Succeed` and `Revert` it is consumed.

src/engine.rs
joshuajbouw: @artob mentioned that the name `call` is wrong, and technically `submit` is a better choice of words. I do agree with that.

src/engine.rs
joshuajbouw: Yeah, easily enough.

src/lib.rs
joshuajbouw: Sure, it shouldn't be able to fail regardless, hence the expect but, I can change it to a `sdk_unwrap` as a "just in case" kind of thing.

src/engine.rs
birchmd: Ok, so `executor.used_gas()` may not be zero (and probably isn't I think) in the `Error` case, but we are choosing to have the semantics that `SubmitResult.gas_used` is always zero in the case of an error. Could you elaborate more on why this is a good semantic to have?

Assuming this is a property we want to enforce then I think we should do it at the type level. I like the guideline of trying to "make invalid states unrepresentable". For example here it should be impossible for `SubmitResult.gas_used > 0` and `SubmitResult.status == false`, but that is not the case as the struct exists currently. Maybe a better representation would be to drop the `status` field and have `gas_used: Option<u64>`, where the `None` case would then represent error. I'm open to other ways of factoring this as well if you think the `Option` is a little awkward, for example `Result<u64, ErrorSingleton>`.

joshuajbouw: I do agree with that philosophy. So, looking at the code of the EVM, there would be a return value on either `Succeed` or `Revert`. `Fatal` and `Error` return a `Vec::new()`. So, it could just simply return `gas_used: Option<u64>` and `result: Option<Vec<u8>>`. But, if they both are going to be optional fields, I tend to like to group those together, perhaps as `SubmitReturn`. Then, you would either have that, or not. I think I'll do that, given what I saw with the fact that `result` too would be returning an empty `Vec` per inner operations.

src/lib.rs
joshuajbouw: Second though, I can't make this into a `sdk_unwrap` as expected as the `Error` returned doesn't implement `ToStr`.

What I would like to do instead is to define our own `ExitError` which may be 1:1, or nearly, what we are currently using, then implement `Display` on that which would then be compatible with every error for `sdk_unwrap()`. I'd say, lets go with it as is for now, and then I'll do a separate PR where I would do `sdk_unwrap()` with the new `Error` type that implements `Display`. Unusually, EVM `ExitError` _-doesn't-_ implement `Display` which I would consider a requirement for any kind of Error to implement both `Debug` and `Display`.

src/parameters.rs
```diff
@@ -32,6 +33,36 @@ pub struct MetaCallArgs {
...
+/// Borsch-encoded log for use in a `RawCallResult`.
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub struct ResultLog {
+    pub topics: Vec<RawU256>,
+    pub data: Vec<u8>,
+}
+
+impl From<Log> for ResultLog {
+    fn from(log: Log) -> Self {
+        let mut res_topics = Vec::with_capacity(log.topics.len());
```
birchmd, line 45: Nit: this transformation could also be done with `map` + `collect` instead of with a mutable Vec and for loop.

I generally think the `map` version is easier to read and fewer lines of code. But it's not a PR blocker by any stretch.

birchmd (approved):
LGTM nice work!

src/parameters.rs
joshuajbouw: No idea how I keep missing these hah! Map + collect is one of my fav features of Rust.

joshuajbouw: Ok, this is resolved.

joshuajbouw:
@artob provided that the CI passes, this is good to go

src/parameters.rs
joshuajbouw: Sorted

artob (changes requested):
A couple of nits.

src/engine.rs
artob: @joshuajbouw That concerned the external interface, not this function. I do think this rename here is a little dubious, I don't really see how this is `submit` other than that you're now returning that result type.

@birchmd In short, `submit` is the name of the method that takes and executes a raw, signed transaction.

src/lib.rs
artob: Addressed by https://github.com/aurora-is-near/aurora-engine/pull/49?

src/engine.rs
artob: As I recall previous discussion on this, the idea was that in the third case we've hit an error not in the user contract being executed, but in the Aurora Engine itself. In that case, we shouldn't charge EVM gas. However, is it really the case that only internal errors will end up in the third case here and everything else will cleanly be caught and returned as a revert?

src/lib.rs
joshuajbouw: Yes, this is addressed by #49 which is in the description of it.

joshuajbouw:
Should GTG provided CI passes.

src/engine.rs
artob: Fixed in 3658d4f.

src/engine.rs
artob: @joshuajbouw Is this something you still plan to do?

src/engine.rs
artob: @joshuajbouw Thoughts? Still planning on doing something with regards to this?

src/engine.rs
joshuajbouw: Not needed now with later changes that shortened that code drastically enough that it no longer is warranted.

src/engine.rs
joshuajbouw: Looked quite deeply into this. It looks like we are required to handle all these ourselves. The disappointing thing is that all the returns are not organised. I'm going to have to submit a PR urgently that will handle which are failures that cost gas, and which do not.

PR #49, opened 2021-05-02
feat: Better errors

joshuajbouw:
This PR came out out of this issue outlined here: https://github.com/aurora-is-near/aurora-engine/pull/48#discussion_r624676151.

Do note, that this PR may look intimidating with amount of files changed + lines, but most of it is refactoring.

This is in general just a much better error structure. I really hated how the underlying EVM for some bizarre reason mixed together BOTH result types. It just can be awkward to work with in general. That is fixed with this PR.

As a bonus, all error returns are now all consistent too as `ERR_`.

Point of discussion: As far as I'm aware, Reverts shouldn't be considered an "ERR" as they are part of everyday EVM life? As in, the contract executes correctly, but due to some circumstance, it'll back out without actually applying changes.

Closes: https://github.com/aurora-is-near/aurora-engine/issues/33

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1,14 +1,106 @@
 use borsh::{BorshDeserialize, BorshSerialize};
 use evm::backend::{Apply, ApplyBackend, Backend, Basic, Log};
 use evm::executor::{MemoryStackState, StackExecutor, StackSubstateMetadata};
-use evm::{Config, CreateScheme, ExitError, ExitReason, ExitSucceed};
+use evm::ExitFatal;
+use evm::{Config, CreateScheme, ExitError, ExitReason};
 
 use crate::parameters::{FunctionCallArgs, NewCallArgs, ViewCallArgs};
 use crate::precompiles;
+use crate::prelude::fmt;
+#[cfg(feature = "std")]
+use crate::prelude::Error;
 use crate::prelude::{Address, Borrowed, Vec, H256, U256};
 use crate::sdk;
 use crate::storage::{address_to_key, storage_to_key, KeyPrefix};
-use crate::types::{bytes_to_hex, log_to_bytes, u256_to_arr, AccountId, NonceError};
+use crate::types::{bytes_to_hex, log_to_bytes, u256_to_arr, AccountId};
+
+/// Errors involving the nonce
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum NonceError {
+    /// Attempted to increment the nonce, but overflow occurred
+    NonceOverflow,
+    /// Account nonce did not match the transaction nonce
+    IncorrectNonce,
+}
+
+impl fmt::Display for NonceError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use NonceError::*;
+        match self {
+            NonceOverflow => write!(f, "ERR_NONCE_OVERFLOW"),
+            IncorrectNonce => write!(f, "ERR_INCORRECT_NONCE"),
+        }
+    }
+}
+
+/// A result for nonces.
+pub type NonceResult<T> = Result<T, NonceError>;
+
+/// Errors with the EVM engine.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum EngineError {
+    /// Normal EVM errors.
+    EvmError(ExitError),
+    /// Fatal EVM errors.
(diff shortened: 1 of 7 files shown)


src/engine.rs
```diff
@@ -144,14 +230,12 @@ impl Engine {
...
-    fn check_increase_balance(address: &Address, amount: &U256) -> Result<U256, ExitError> {
+    fn check_increase_balance(address: &Address, amount: &U256) -> Result<U256, EngineError> {
         let balance = Self::get_balance(address);
         if let Some(new_balance) = balance.checked_add(*amount) {
             Ok(new_balance)
         } else {
-            Err(ExitError::Other(Borrowed(
-                "balance is too high, can not increase",
-            )))
+            Err(ExitError::Other(Borrowed("balance is too high, can not increase")).into())
```
artob, line 238: Missed this one (not an error code)

src/engine.rs
```diff
@@ -162,14 +246,12 @@ impl Engine {
...
-    fn check_decrease_balance(address: &Address, amount: &U256) -> Result<U256, ExitError> {
+    fn check_decrease_balance(address: &Address, amount: &U256) -> Result<U256, EngineError> {
         let balance = Self::get_balance(address);
         if let Some(new_balance) = balance.checked_sub(*amount) {
             Ok(new_balance)
         } else {
-            Err(ExitError::Other(Borrowed(
-                "balance is too low, can not decrease",
-            )))
+            Err(ExitError::Other(Borrowed("balance is too low, can not decrease")).into())
```
artob, line 254: Missed this one (not an error code)

src/engine.rs
```diff
@@ -1,14 +1,100 @@
...
+            EvmError(ExitError::InvalidRange) => "ERR_INVALID_RANGE",
+            EvmError(ExitError::DesignatedInvalid) => "ERR_DESIGNATED_INVALID",
+            EvmError(ExitError::CallTooDeep) => "ERR_CALL_TOO_DEEP",
+            EvmError(ExitError::CreateCollision) => "ERR_CREATE_COLLISION",
+            EvmError(ExitError::CreateContractLimit) => "ERR_CREATE_CONTRACT_LIMIT",
+            EvmError(ExitError::OutOfOffset) => "ERR_OUT_OF_OFFSET",
+            EvmError(ExitError::OutOfGas) => "ERR_OUT_OF_GAS",
+            EvmError(ExitError::OutOfFund) => "ERR_OUT_OF_FUND",
+            EvmError(ExitError::Other(m)) => m,
+            EvmError(_) => "", // unused misc
```
artob, line 63: Is this reachable?

src/engine.rs
```diff
@@ -1,14 +1,100 @@
...
+            EvmFatal(ExitFatal::NotSupported) => "ERR_NOT_SUPPORTED",
+            EvmFatal(ExitFatal::UnhandledInterrupt) => "ERR_UNHANDLED_INTERRUPT",
+            EvmFatal(ExitFatal::Other(m)) => m,
+            EvmFatal(_) => "", // unused misc
```
artob, line 67: Is this reachable?

artob (changes requested):
LGTM generally :+1: Please address the two nits and clarify reachability of the two other comments.

src/engine.rs
joshuajbouw: Yes, there are some errors as part of `EvmError` that are flagged as unused. Not sure why they left them in as its totally dead code and nothing handles it.

src/engine.rs
joshuajbouw: Same as the above response but for Fatal.

artob:
> As far as I'm aware, Reverts shouldn't be considered an "ERR" as they are part of everyday EVM life? As in, the contract executes correctly, but due to some circumstance, it'll back out without actually applying changes.

Yes, EVM reverts aren't an Aurora Engine error condition, just another ordinary result from our point of view. But that needs us to extend the result types for `call`, `view`, as we already did for `submit` (nee `raw_call`). So, better handled in #48, and the way this PR handles revert is fine for the scope here.

src/engine.rs
artob: Can we put an `unreachable` guard here instead of an empty string?

joshuajbouw: Sure.

artob (approved):
LGTM :+1: Nice work!

PR #51, opened 2021-05-04
Introduce precompiles to manage ERC20 connector

mfornet:
This creates two new precompiles:
- `exit_to_near`
- `exit_to_ethereum`

Their main role is scheduling a promise call outside of the EVM to another Near contract.
Check [this diagram](https://miro.com/app/board/o9J_lJyoV8I=/) for reference.

## Testing Plan

This are a set of basic tests that should be implemented before merging. This tests only include happy path for expected workflows, but it is important to cover with tests that other interaction are behaving as expected (tokens can't get lost or stolen).

- [X] (Rust Test) Can deploy ERC20 tokens from the EVM
- [X] (Rust Test) Can mint some tokens from admin access function
- [X] (Rust Test) ft_on_transfer from ERC20
- [X] (Rust Test) Register relayer (and check fee is ok from ft_on_transfer)
- [X] (Rust Test) Check that relayer gets payed
- [X] (Rust Test) Exit to Near
- [X] (Rust Test) Exit to Ethereum

Integration tests should spinup Ganache, and Localnet and run there, to make sure each workflow is behaving correctly in a environment closer to production.

- [x] (~~Integration~~ Manual Test) Send tokens from Near -> EVM
- [x] (~~Integration~~ Manual Test) Send tokens from EVM -> Near
- [x] (~~Integration~~ Manual Test) Send tokens from Ethereum -> EVM
- [x] (~~Integration~~ Manual Test) Send tokens from EVM -> Ethereum

Test contract_call is only scheduled appropriately:

Create a solidity with these functions:
- [x] A -> Call the precompile (which in turns schedule a contract_call on success)
- [x] B -> Call A as an external function and run require(false)
- [x] C -> Call B and catch the error (so the function doesn’t fail)
- [x] D -> Call C and Call A

Scenarios:

- [x] Call A and check contract call was emitted
- [x] Call B and check no contract call was emitted and tx failed
- [x] Call C and check no contract call was emitted and tx succeeded
- [x] Call D and check that ONLY one contract call was emitted and tx succeded
- [x] Previos set of tests should be executed with ExitToNear and ExitToEthereum


src/precompiles.rs
```diff
@@ -1,8 +1,22 @@
...
 type PrecompileResult = Result<(ExitSucceed, Vec<u8>, u64), ExitError>;
 
+// Computed as: near_account_to_evm_address("exitToEthereum".as_bytes()).to_low_u64_be()
+const EXIT_TO_ETHEREUM_ID: u64 = 17176159495920586411;
+
+// Computed as: near_account_to_evm_address("exitToNear".as_bytes()).to_low_u64_be()
+const EXIT_TO_NEAR_ID: u64 = 11421322804619973199;
+
 #[allow(dead_code)]
 pub fn no_precompiles(
```
joshuajbouw, line 21: Maybe this should be renamed to `no_externel_precompiles` or something along those lines?

src/precompiles.rs
```diff
@@ -42,11 +56,156 @@ pub fn istanbul_precompiles(
...
+            Some(Ok((ExitSucceed::Returned, Vec::new(), 0)))
+        }
         // Not supported.
         _ => None,
     }
 }
 
+#[allow(dead_code)]
+fn exit_to_near(input: &[u8], context: &Context) {
+    // TODO: Determine the correct amount of gas
```
joshuajbouw, line 74: Note TODO

src/precompiles.rs
```diff
@@ -42,11 +56,156 @@ pub fn istanbul_precompiles(
...
+    const GAS_FOR_FT_TRANSFER: Gas = 50_000;
```
joshuajbouw, line 75: I would put this in a mod called `costs` like the other precompiles.

src/precompiles.rs
```diff
@@ -42,11 +56,156 @@ pub fn istanbul_precompiles(
...
+    // if Self::required_gas(input)? > target_gas {
+    //     return Err(ExitError::OutOfGas);
+    // }
```
joshuajbouw, line 78: This is required.

src/precompiles.rs
```diff
@@ -42,11 +56,156 @@ pub fn istanbul_precompiles(
...
+        // ERC20 transfer
+        //
+        // This precompile branch is expected to be called from the ERC20 burn function\
+        //
+        // Input slice format:
+        //      sender (20 bytes) - Eth address of the address that burned his erc20 tokens
+        //      amount (U256 le bytes) - the amount that was burned
+        //      recipient_account_id (bytes) - the NEAR recipient account which will receive NEP-141 tokens
+
+        //TODO: add method in Aurora connector and call promise `get_near_account_for_evm_token(context.caller)`
```
joshuajbouw, line 104: Note TODO

src/precompiles.rs
```diff
@@ -42,11 +56,156 @@ pub fn istanbul_precompiles(
...
+        //
+        // Input slice format:
+        //      sender (20 bytes) - Eth address of the address that burned his erc20 tokens
+        //      amount (U256 le bytes) - the amount that was burned
+        //      eth_recipient (20 bytes) - the address of recipient which will receive ETH on Ethereum
+
+        //TODO: add method in Aurora connector and call promise `get_near_account_for_evm_token(context.caller)`
+        let nep141address = context.caller.to_string();
+
+        let mut input = input;
```
joshuajbouw, line 177: Probably a good idea to make this a new variable like, `account_bytes`

src/precompiles.rs
```diff
@@ -42,11 +56,156 @@ pub fn istanbul_precompiles(
...
+
+        let mut sender = [0u8; 20];
+        sender.copy_from_slice(&input[..20]);
+        input = &input[20..];
+
+        let amount = U256::from_little_endian(&input[..32]).as_u128();
+        input = &input[32..];
```
joshuajbouw, line 184: Always very good practice to indicate the starting and ending byte.

src/types.rs
```diff
@@ -12,6 +12,7 @@ pub type AccountId = String;
 pub type RawAddress = [u8; 20];
 pub type RawU256 = [u8; 32]; // Little-endian large integer type.
 pub type RawH256 = [u8; 32]; // Unformatted binary data of fixed length.
+pub type Gas = u64;
```
joshuajbouw, line 15: Just a note here, this should be applied to other precompiles and elsewhere. A Newtype likely may be preferred to ensure that there is no accidental mixing with other u64. Not recommended for now.

joshuajbouw (changes requested):
Just some small changes, and also to have it fit with the other precompiles. I believe it also needs to be added to the other precompile functions.

src/types.rs
artob: Yeah, a good idea to have a distinct type for gas.

src/precompiles.rs
artob: Also, naming conventions for `nep141address`: `nep141_address`.

sept-en:
@joshuajbouw thanks for your help, LGTM!
@mfornet please take a look at the updated code and approve if everything looks good for you as well.

sept-en (approved):

src/precompiles.rs
mfornet: Solved in #52

src/precompiles.rs
mfornet: Done!

src/precompiles.rs
mfornet: It is unbounded, so it is not possible to select ending byte in a useful way.

src/types.rs
mfornet: > Not recommended for now.

👀 
It is very easy change, and it is ok to introduce it from the beginning, why not now?

src/parameters.rs
```diff
@@ -89,6 +89,40 @@ pub struct BeginBlockArgs {
     pub gaslimit: RawU256,
 }
 
+// TODO: Use proper serializer
```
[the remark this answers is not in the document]
mfornet: Updated description with remaining TODO, and converted to draft while it is not implemented.

src/types.rs
joshuajbouw: Just out of scope to add them to all precompiles. You can do it, but probably after this PR is merged.

src/precompiles.rs
joshuajbouw: I see.

src/types.rs
artob: Exactly, let's keep PRs atomic. That will also help avoid merge conflicts given active work by others.

src/precompiles.rs
mfornet: I'm solving all TODO(#51) before merging, will mark as resolved this comment though the TODO itself is not resolved.

mrLSD (changes requested):

src/engine.rs
```diff
@@ -1,17 +1,42 @@
...
+    ($e:expr, $output:expr) => {
+        match $e {
+            Ok(v) => v,
+            Err(_) => {
+                sdk::return_output($output);
+                return;
+            }
+        }
+    };
+}
```
birchmd, line 29: I don't love this style of macro. It effectively does the same thing as the `?` operator, but more implicitly. It makes me think the need for this macro is a bit of a code smell that indicates some function should be returning a `Result` (and then could use the `?` operator), but isn't. If the function did return a result then the callsite would be the only place where this match is needed and so the macro would not be necessary.

src/engine.rs
```diff
@@ -397,6 +430,134 @@ impl Engine {
...
+    /// The output will be serialized as a String
+    /// https://github.com/near/NEPs/discussions/146
+    ///
+    /// IMPORTANT: This function should not panic, otherwise it won't
+    /// be possible to return the tokens to the sender.
+    ///
+    /// TODO(#51): Carefully review that this function can't fail,
+    ///     and returns appropriate value
+    pub fn receive_erc20_tokens(&mut self) {
+        let input = sdk::read_input();
```
birchmd, line 479: I prefer the core logic being in `Engine`, and the IO happening in `lib.rs`. In this example

```rust
// engine.rs

pub fn receive_erc20_tokens(&mut self, sender_id: AccountId, amount: Balance, msg: String) -> Result<Balance, Error> { ... }
```

``` rust
// lib.rs

    /// Allow receiving NEP141 tokens to the EVM contract
    #[no_mangle]
    pub extern "C" fn ft_on_transfer() {
        #[allow(clippy::if_same_then_else)]
        if sdk::predecessor_account_id() == sdk::current_account_id() {
            // TODO(#59) ETH transfer
            todo!();
        } else {
            let result = match NEP141FtOnTransferArgs::try_from_slice(input.as_slice()) {
                Ok(args) => engine.receive_erc20_tokens(args.sender_id, args.amount, args.msg).ok_or(ags.amount),
                Err(_) => 0,
            }
            sdk::return_output(result.to_string.as_bytes());
        }
    }
```

The reason I think this is better is because 

1. It is more clear from the type signature what `receive_erc20_tokens` is trying to do
2. `receive_erc20_tokens` becomes easier to test (input is passed directly, reducing reliance on `sdk`)

src/lib.rs
```diff
@@ -276,11 +282,42 @@ mod contract {
...
+            DeployErc20TokenArgs::try_from_slice(&sdk::read_input()).sdk_expect("ERR_ARG_PARSE");
+
+        let mut engine = Engine::new(predecessor_address());
+
+        let erc20_contract = include_bytes!("../etc/eth-contracts/res/EvmErc20.bin");
+        let deploy_args = ethabi::encode(&[
+            ethabi::Token::String("TestToken".to_string()),
+            ethabi::Token::String("TT".to_string()),
+            ethabi::Token::Uint(ethabi::Uint::from(0)),
+            ethabi::Token::Address(current_address().into()),
```
birchmd, line 304: These values look suspicious

src/precompiles/mod.rs
```diff
@@ -3,58 +3,82 @@ mod bn128;
...
 /// Exit to Ethereum precompile address (truncated to 8 bytes)
 ///
 /// Address: `0xb0bd02f6a392af548bdf1cfaee5dfa0eefcc8eab`
 /// This address is computed as: `&keccak("exitToEthereum")[12..]`
 const EXIT_TO_ETHEREUM_ID: u64 = 17176159495920586411;
 
+#[allow(dead_code)]
+pub fn exit_to_ethereum_address() -> Address {
+    Address::from_slice(
+        hex::decode("b0bd02f6a392af548bdf1cfaee5dfa0eefcc8eab")
```
birchmd, line 35: Just curious, where does this value come from?

src/state.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn reset_balance(&mut self, address: H160) {
+        self.memory_stack_state.reset_balance(address)
+    }
+
+    fn touch(&mut self, address: H160) {
+        self.memory_stack_state.touch(address)
+    }
+}
+
+struct Stack<T> {
```
birchmd, line 179: I'd love to see more tests with this around `enter`, `commit`, `discard` (especially with multiple nesting of `enter`).

birchmd (commented):
This is a big PR! Thanks for all your work on this!

Generally looks good to me, but I have a some comments. I didn't look too closely at the business logic yet, I can do that in another pass.

Personally, I think the type generics on the precompiles look fine (even _good_ 😄 )

src/precompiles/mod.rs
sept-en: @birchmd take a look [there](https://github.com/aurora-is-near/aurora-engine/blob/060f31033880eaa7bc03b64e0237d7476db624db/src/precompiles/mod.rs#L23)

src/lib.rs
sept-en: I think this is here just for testing purposes as for now.

src/engine.rs
sept-en: Please note that we are using the same approach for many functions (e.g. eth-connector related as well).
But I really like the suggestion. Especially the second item is really crucial.

src/precompiles/mod.rs
birchmd: Thanks @sept-en ! If we have a constant for it already, maybe the address could be derived using that instead of the hex string here?

src/engine.rs
birchmd: Yes, I know other eth-connector uses the style of reading input together with business logic. I think I mentioned this in a previous review of eth-connector code. I would like to see us move away from that pattern if possible though.

To be clear, I don't think this is a blocking concern (especially with timelines being what they are), but it would be nice to separate IO from logic where we can, and go back to fix all the places where we didn't in a future tech debt cycle.

sept-en: Totally agree with you!

src/precompiles/mod.rs
sept-en: Yes, I think so. IMO I think even better is to have the address defined as the constant separately as well.

birchmd: I agree it shouldn't need to be computed at runtime if it is derived from a constant. Maybe a `const fn` could be used?

I think we should avoid having two separate constants defined independently though because otherwise in the future we may update one but forget to update the other.

sept-en: I like the idea of using the `const fn` approach!

src/engine.rs
```diff
@@ -108,6 +133,10 @@ pub struct EngineState {
     pub upgrade_delay_blocks: u64,
     /// Mapping between relayer account id and relayer evm address
     pub relayers_evm_addresses: LookupMap<{ KeyPrefix::RelayerEvmAddressMap as KeyPrefixU8 }>,
+    /// Mapping between erc20 tokens in the EVM to nep141 in NEAR
+    pub nep141_erc20: LookupMap<{ KeyPrefix::Nep141Erc20Map as KeyPrefixU8 }>,
+    /// Mapping between nep141 in NEAR to erc20 tokens in the EVM
+    pub erc20_nep141: LookupMap<{ KeyPrefix::Erc20Nep141Map as KeyPrefixU8 }>,
```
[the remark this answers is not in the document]
mfornet: Created bijection map!

src/engine.rs
```diff
@@ -397,6 +430,134 @@ impl Engine {
...
             .map(|result| Address(result.as_slice().try_into().unwrap()))
     }
+
+    pub fn register_token(&mut self, erc20_token: &[u8], nep141_token: &[u8]) {
+        // Check that this nep141 token was not registered before, they can only be registered once.
+        assert!(self.state.nep141_erc20.get_raw(nep141_token).is_none());
+
+        self.state
+            .erc20_nep141
+            .insert_raw(erc20_token, nep141_token);
```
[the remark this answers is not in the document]
mfornet: done!

mfornet:
Addressed comments by @abacabadabacaba 

1. Concatenate bytes in solidity using `abi.encodePacked`
2. Prevent calling precompiles from staticcall functions. For this, a modification to the EVM was required to pass `is_static` flag to the precompile. See more in this PR: https://github.com/aurora-is-near/sputnikvm/pull/2
3. Revert the effect of a call in the relevant places, (i.e when the function that calls the precompile fails and is reverted)
4. Use flag to distinguish between exitToNear/exitToEthereum (with Erc20/Eth). There are two precompiles, and the first byte they receive is a flag to distinguish between eth and erc20.
5. Fix json injection, by checking the input format is correct (when the format is correct it is not possible to inject json).

Cargo.toml
```diff
@@ -41,8 +41,10 @@ rpath = false
 blake2 = { git = "https://github.com/near/near-blake2.git", version = "0.9.1", default-features = false }
 borsh = { version = "0.8.2", default-features = false }
 bn = { package = "aurora-bn", git = "https://github.com/aurora-is-near/aurora-bn.git", default-features = false }
-evm = { version = "0.26.0", default-features = false }
-evm-core = { version = "0.26.1", default-features = false }
+evm = { path = "../evm", default-features = false }
+#evm = { version = "0.26.0", default-features = false }
+evm-core = { path = "../evm/core", default-features = false }
+#evm-core = { version = "0.26.1", default-features = false }
```
mfornet, line 47: I will point to the right dependency when everything is settled. For now this is depending on this changes: https://github.com/aurora-is-near/sputnikvm/pull/2

src/engine.rs
mfornet: I tried to change the output to Result, but couldn't find a natural way to do it.

I tried using `Result<&[u8], &[u8]>`, but the output should be handled in the same way for `Ok` and `Err`. So handled the output with:

```
let output = match engine.receive_erc20_tokens(args) {
                Ok(output) => output,
                Err(output) => output,
};
```

The more complicated part is that the output is built inside the function, so we can't return &[u8], and should return Vec<u8>.

I will not handle this for now, and we can address it later to not block on this.

src/engine.rs
mfornet: done!

src/lib.rs
mfornet: This is "correct". Anyone can deploy an erc20 token and we populate metadata with default values. Will change name from `TestToken` to `Empty`, and symbol from `TT` to `EMPTY`.  Decimals are `0`, and last field is admin of the contract.

mfornet: We need to implement a way for trustlessly bridge metadata to the contract. For now, I'll add a trusted setup, where user with FullAccessKey will be able to set the metadata, so we can populate metadata properly for initial ERC20

src/engine.rs
birchmd: Sure, we can address this separately, no need to block on it.

Returning `Result<&[u8], &[u8]>` seems odd though. I would have expected something like `Result<&[u8], Error>` and then at the callsite `call_to_fn(...).unwrap_or(default_value)`. If the value to return changes depending on where the error happened then `unwrap_or` doesn't work and it would have to be an explicit match where the different error cases map to different default values.

I haven't looked very closely at where this macro is used to know if there is a complexity I am missing. But again, I'm ok with not blocking the PR and coming back to this later.

src/precompiles/native.rs
```diff
@@ -65,190 +66,295 @@ pub fn is_valid_account_id(account_id: &[u8]) -> bool {
...
+        input_mut = &input_mut[1..];
+
+        let (nep141_address, args) = match flag {
+            0x0 => {
+                // ETH transfer
+                //
+                // Input slice format:
+                //      recipient_account_id (bytes) - the NEAR recipient account which will receive NEP-141 ETH tokens
+
+                if is_valid_account_id(input) {
```
sept-en, line 134:
```suggestion
                if is_valid_account_id(input_mut) {
```

src/precompiles/native.rs
```diff
@@ -65,190 +66,295 @@ pub fn is_valid_account_id(account_id: &[u8]) -> bool {
...
+                    (
+                        String::from_utf8(crate::sdk::current_account_id()).unwrap(),
+                        // There is no way to inject json, given the encoding of both arguments
+                        // as decimal and valid account id respectively.
+                        crate::prelude::format!(
+                            r#"{{"receiver_id": "{}", "amount": "{}", "memo": null}}"#,
+                            String::from_utf8(input.to_vec()).unwrap(),
```
sept-en, line 141:
```suggestion
                            String::from_utf8(input_mut.to_vec()).unwrap(),
```

sept-en (changes requested):

src/precompiles/native.rs
mfornet: Nice catch! to avoid this error, I replaced all `input_mut` with `input`, and shadowed previos `input` with a mutable version of itself.

Cargo.toml
mfornet: It is pointing right now, to fixed github commit.

src/state.rs
```diff
@@ -0,0 +1,178 @@
...
+        self.memory_stack_state.exit_commit()
+    }
+
+    fn exit_revert(&mut self) -> Result<(), ExitError> {
+        self.promises.discard();
+
+        self.memory_stack_state.exit_revert()
+    }
+
+    fn exit_discard(&mut self) -> Result<(), ExitError> {
```
birchmd, line 129: Just curious, what is the difference between `exit_revert` and `exit_discard` in terms of semantics of the EVM?

src/test_utils/contract_call.rs
```diff
@@ -0,0 +1,171 @@
+use crate::parameters::{FunctionCallArgs, SubmitResult};
+use crate::prelude::Address;
+use crate::test_utils::{solidity, AuroraRunner, Signer};
+use crate::transaction::EthTransaction;
+use borsh::BorshSerialize;
+use std::convert::TryInto;
+
+pub(crate) struct TesterConstructor(pub solidity::ContractConstructor);
```
birchmd, line 8: All the naming here is very generic; it's a little hard for me to know exactly what this is testing. I don't necessarily think any of these names need to change if the module name changes though. For example `exit_precompiles.rs`.

src/types.rs
```diff
@@ -198,6 +236,84 @@ mod tests {
...
+
+    fn check_stack(stack: &Stack<i32>, expected: Vec<Option<i32>>) {
+        if let Some(&last) = stack.boundaries.last() {
+            assert!(last <= stack.stack.len());
+        }
+        assert_eq!(view_stack(stack), expected);
+    }
+
+    #[test]
+    fn test_stack() {
```
birchmd, line 268: ❤️

src/types.rs
```diff
@@ -156,6 +156,44 @@ pub fn near_account_to_evm_address(addr: &[u8]) -> Address {
...
+
+    pub fn commit(&mut self) {
+        self.boundaries.pop().unwrap();
+    }
+
+    pub fn discard(&mut self) {
+        let boundary = self.boundaries.pop().unwrap();
+        while self.stack.len() > boundary {
+            self.stack.pop().unwrap();
+        }
```
birchmd, line 186: Could use [`truncate` method](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#1011) here instead.

birchmd (commented):
Great work, this is a big PR with lots of tricky stuff in it. Most of my comments here are minor, but a major issue for me is that I cannot compile the tests locally because I get a linker error (`multiple definition of __jit_debug_descriptor`). Is this something you are encountering as well, or have I somehow misconfigured something?

mfornet:
> I cannot compile the tests locally because I get a linker error (`multiple definition of __jit_debug_descriptor`). Is this something you are encountering as well, or have I somehow misconfigured something?

I can run tests with `cargo test` from the root normally (some of them are failing)

src/state.rs
mfornet: `exit_revert` is to signal a problem with the code on the given input (i.e. Runtime error triggered by an `assert/require`).

AFAIU `exit_discard` is called when some invariant about the EVM are violated. For example trying to [deploy two contracts to the same address](https://github.com/aurora-is-near/sputnikvm/blob/master/src/executor/stack/mod.rs#L387) or [nonce of some account is non zero](https://github.com/aurora-is-near/sputnikvm/blob/master/src/executor/stack/mod.rs#L392) while deploying a contract.

src/sdk.rs
```diff
@@ -262,6 +262,31 @@ pub fn remove_storage(key: &[u8]) {
     }
 }
 
+pub fn register_len(register_id: u64) -> Option<u64> {
+    let len = unsafe { exports::register_len(register_id) };
+
+    if len == u64::MAX {
```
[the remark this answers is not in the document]
mfornet: Notice this is interacting with the low level NEAR blockchain API. I took [this code from `near-sdk`](https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/environment/env.rs#L135-L145) and adapted it.

src/sdk.rs
```diff
@@ -262,6 +262,31 @@ pub fn remove_storage(key: &[u8]) {
...
+/// Register used to record evicted values from the storage.
+const EVICTED_REGISTER: u64 = u64::MAX - 1;
+
+// TODO(MarX): Test this function
+/// Reads the most recent value that was evicted with `storage_write` or `storage_remove` command.
+pub fn storage_get_evicted() -> Option<Vec<u8>> {
+    let len = register_len(EVICTED_REGISTER)?;
+    let bytes: Vec<u8> = vec![0u8; len as usize];
+    unsafe {
+        exports::read_register(1, bytes.as_ptr() as *const u64 as u64);
```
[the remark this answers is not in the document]
mfornet: Will do that, similar to how it is [already done for other registers](https://github.com/aurora-is-near/aurora-engine/blob/45481254c44793573730867c105700a7a62414be/src/sdk.rs#L6-L7).

As a side note, what is happening is that when you call a low level function in NEAR blockchain (the result is stored in some register, that you can read from). Register `1` is not unique for `EVICTED_REGISTER`. For example: [here](https://github.com/aurora-is-near/aurora-engine/blob/45481254c44793573730867c105700a7a62414be/src/sdk.rs#L321) and [here](https://github.com/aurora-is-near/aurora-engine/blob/45481254c44793573730867c105700a7a62414be/src/sdk.rs#L331).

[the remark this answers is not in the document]
mfornet: Good catch 👌

src/precompiles/native.rs
```diff
@@ -45,95 +44,149 @@ impl ExitToNear {
...
-                return Err(ExitError::Other(Cow::from(
-                    "ERR_INVALID_RECEIVER_ACCOUNT_ID",
-                )));
+            0x1 => {
+                // ERC20 transfer
+                //
+                // This precompile branch is expected to be called from the ERC20 burn function\
+                //
+                // Input slice format:
+                //      amount (U256 le bytes) - the amount that was burned
```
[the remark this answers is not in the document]
mfornet: Great catch!

[Code never lies, comments sometimes do](https://brunocapuano.files.wordpress.com/2017/09/code-never-lies.jpg)

src/precompiles/native.rs
```diff
@@ -45,95 +44,149 @@ impl ExitToNear {
         super::make_address(0xe9217bc7, 0x0b7ed1f598ddd3199e80b093fa71124f);
 }
 
-impl Precompile for ExitToNear {
+fn get_nep141_from_erc20(erc20_token: &[u8]) -> AccountId {
+    AccountId::from_utf8(
+        crate::sdk::read_storage(bytes_to_key(KeyPrefix::Erc20Nep141Map, erc20_token).as_slice())
+            .expect(ERR_TARGET_TOKEN_NOT_FOUND),
+    )
+    .unwrap()
```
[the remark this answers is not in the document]
mfornet: Absolutely, actually I think we should find a way to totally disallow calling `crate::sdk::read_storage` at all (by making this function non-public) and only available for `storage.rs` module. Otherwise building abstraction and doing migrations will be crazy.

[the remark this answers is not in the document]
mfornet: Well, after a second thought, this is tricky here, because inside the precompile we don't have access to the Engine internal state. We can recreate it by calling Engine::new() and then using proper abstraction, but that would be a waste of resources loading the engine state again.

That makes me think, that I'll change implementation a bit. Instead of storing bijection map as a field, I'll have static function that builds the interface for the bijection map (no state need to be touched for this operation), here it is very helpful to have it implemented with keys as generic consts.

src/lib.rs
```diff
@@ -309,6 +314,67 @@ mod contract {
...
+            EthConnectorContract::get_instance().ft_on_transfer(&engine)
+        } else {
+            let input = sdk::read_input();
+
+            let args = if let Ok(args) = NEP141FtOnTransferArgs::try_from_slice(input.as_slice()) {
+                args
+            } else {
+                // At this point the amount is not known yet.
+                // It is responsibility of the NEP141 to provide correct arguments.
+                sdk::return_output(b"0");
```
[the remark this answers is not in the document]
mfornet: Good point, that would make more sense actually, so we give the NEP141 contract the opportunity to know about the error and recover from that.

src/engine.rs
```diff
@@ -414,6 +448,138 @@ impl Engine {
...
+                    recipient.clone(),
+                    relayer_address.clone(),
+                    Wei::new_u64(fee.as_u64()),
+                    gas_limit
+                ),
+                output_on_fail
+            );
+        }
+
+        let selector = &sdk::keccak("mint(address,uint256)".as_bytes())[..4];
```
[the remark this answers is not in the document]
mfornet: To do properly we need something like this:

```
const fn selector() -> Vec<u8> {
    let mut hasher = sha3::Keccak256::new();
    hasher.update("mint(address,uint256)".as_bytes());
    hasher.finalize()[..4].to_vec()
}
```

But this fails to compile with: 

```
error[E0015]: calls in constant functions are limited to constant functions, tuple structs and tuple variants
  --> src/engine.rs:26:22
   |
26 |     let mut hasher = sha3::Keccak256::new();
   |                      ^^^^^^^^^^^^^^^^^^^^^^

error[E0658]: mutable references are not allowed in constant functions
  --> src/engine.rs:27:5
   |
27 |     hasher.update("mint(address,uint256)".as_bytes());
   |     ^^^^^^
   |
   = note: see issue #57349 <https://github.com/rust-lang/rust/issues/57349> for more information
   = help: add `#![feature(const_mut_refs)]` to the crate attributes to enable

error[E0015]: calls in constant functions are limited to constant functions, tuple structs and tuple variants
  --> src/engine.rs:27:5
   |
27 |     hasher.update("mint(address,uint256)".as_bytes());
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0015]: calls in constant functions are limited to constant functions, tuple structs and tuple variants
  --> src/engine.rs:28:5
   |
28 |     hasher.finalize()[..4].to_vec()
   |     ^^^^^^^^^^^^^^^^^

error[E0015]: calls in constant functions are limited to constant functions, tuple structs and tuple variants
  --> src/engine.rs:28:5
   |
28 |     hasher.finalize()[..4].to_vec()
   |     ^^^^^^^^^^^^^^^^^^^^^^

error[E0015]: calls in constant functions are limited to constant functions, tuple structs and tuple variants
  --> src/engine.rs:28:5
   |
28 |     hasher.finalize()[..4].to_vec()
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

An alternative is hardcode this value (since it will not change), add a comment about how it was created and add test that confirms it is the expected value.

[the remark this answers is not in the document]
birchmd: > An alternative is hardcode this value (since it will not change), add a comment about how it was created and add test that confirms it is the expected value.

Yes, this is what I meant. Sorry for the ambiguity.

src/engine.rs
```diff
@@ -414,6 +448,138 @@ impl Engine {
...
+            output_on_fail
+        );
+
+        // Everything succeed so return "0"
+        sdk::return_output(b"0");
+    }
+
+    fn schedule_promises(promises: impl IntoIterator<Item = PromiseCreateArgs>) {
+        for promise in promises {
+            sdk::log_utf8(
```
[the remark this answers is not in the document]
birchmd: Agreed. A macro seems like it would be appropriate to do that.

src/precompiles/native.rs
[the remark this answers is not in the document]
birchmd: > Instead of storing bijection map as a field, I'll have static function that builds the interface for the bijection map

Yeah, this is a cool idea. I like it better than having the raw lookup here. I'm a little worried about whether this will be an abuse of the environment (it relies on the existence of ambient storage) which might be hard to reproduce in unit tests. But maybe that is a necessary evil and these functions will only ever be covered by integration tests.

src/sdk.rs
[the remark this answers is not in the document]
birchmd: I think duplicating the comment from `nearcore` on `exports::register_len` could be helpful for future devs (so that we don't have to go back to the low-level documentation all the time)

```
    /// Returns the size of the blob stored in the given register.
    /// * If register is used, then returns the size, which can potentially be zero;
    /// * If register is not used, returns `u64::MAX`
```

birchmd:
> > I cannot compile the tests locally because I get a linker error (`multiple definition of __jit_debug_descriptor`). Is this something you are encountering as well, or have I somehow misconfigured something?
> 
> I can run tests with `cargo test` from the root normally (some of them are failing)

Weird that it is working for you. The CI checks are also failing with the same `multiple definition of __jit_debug_descriptor` that I have locally.

.github/workflows/lints.yml
```diff
@@ -34,8 +34,23 @@ jobs:
...
+
+  contracts:
+    name: eth-contracts
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone the repository
+        uses: actions/checkout@v2
+      - name: Run yarn lint
+        working-directory: etc/eth-contracts
+        run: yarn && yarn lint
```
joshuajbouw, line 56: Add newline here

Cargo.toml
```diff
@@ -41,8 +41,8 @@ rpath = false
 blake2 = { git = "https://github.com/near/near-blake2.git", version = "0.9.1", default-features = false }
 borsh = { version = "0.8.2", default-features = false }
 bn = { package = "aurora-bn", git = "https://github.com/aurora-is-near/aurora-bn.git", default-features = false }
-evm = { version = "0.27.0", default-features = false }
-evm-core = { version = "0.27.1", default-features = false }
+evm = { git = "https://github.com/aurora-is-near/sputnikvm", default-features = false }
+evm-core = { git = "https://github.com/aurora-is-near/sputnikvm", default-features = false }
```
joshuajbouw, line 45: I don't think we should be using our own sputnikvm here. We should only stick with upstream...?

Makefile
```diff
@@ -1,6 +1,6 @@
 CARGO = cargo
 NEAR  = near
-FEATURES = contract,log
```
joshuajbouw, line 3: Just a sidenote, shouldn't this only be contract + exit-precompiles? We should be using log, right?

src/connector.rs
```diff
@@ -312,8 +312,11 @@ impl EthConnectorContract {
         // Check promise results
         let data0: Vec<u8> = match sdk::promise_result(0) {
             PromiseResult::Successful(x) => x,
-            _ => sdk::panic_utf8(b"ERR_PROMISE_INDEX"),
+            PromiseResult::Failed => sdk::panic_utf8(b"ERR_???"),
```
joshuajbouw, line 315: No descriptive enough.

src/connector.rs
```diff
@@ -312,8 +312,11 @@ impl EthConnectorContract {
...
+            // This shouldn't be reachable
+            PromiseResult::NotReady => sdk::panic_utf8(b"ERR_PROMISE_NOT_READY"),
```
joshuajbouw, line 317: Shouldn't we use `unreachable!` instead?

src/engine.rs
```diff
@@ -1,19 +1,53 @@
...
+                #[cfg(feature = "log")]
+                sdk::log(crate::prelude::format!("{:?}", _e).as_str());
+                sdk::return_output($output);
+                return;
+            }
+        }
+    };
+}
+
+macro_rules! assert_or_finish {
```
joshuajbouw, line 42: Just a nit, these macro names feel a bit off. Technically just `unwrap` conveys that meaning that the app will unwrap, or panic. Also, don't we have these already? `sdk_unwrap` etc.?

src/engine.rs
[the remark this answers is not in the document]
joshuajbouw: Sounds good to me.

joshuajbouw (changes requested):
I feel uncertain why we need an `AuroraStackState`, and why the precompiles now need the state implemented? I almost went down this path, but found solutions around not having to.

src/engine.rs
birchmd: See also previous discussion: https://github.com/aurora-is-near/aurora-engine/pull/51#discussion_r637229881

Essentially my opinion is that we should be able to remove these macros altogether, but we won't do it now since apparently it's non-trivial and this PR is important.

birchmd:
> I feel uncertain why we need an `AuroraStackState`, and why the precompiles now need the state implemented? I almost went down this path, but found solutions around not having to.

The reason is because we need to know what promises were asked to be created (from calls to exit precompiles), but we can't have the precompiles eagerly create the promises because then they cannot be rolled back if the transaction (or sub-call in a transaction) reverts.

The `AuroraStackState` keeps track of the data needed to construct the promises and only once the entire transaction is finished do we actually create them.

Cargo.toml
mfornet: While https://github.com/rust-blockchain/evm/pull/34 is not merged, we need to use ours, to keep track of the state in precompiles?

src/connector.rs
mfornet: I missed this, wrote it while debugging and never went back to it.

src/connector.rs
mfornet: It is better to fail with some significant error in case something doesn't work as we expect. In this case, we can reach this branch if something was missunderstood by us, or third party violates some invariant. I'd keep it as it is right now, rather than converting into unreachable!()

Makefile
sept-en: We need logging to know what happens inside the contract. Otherwise we won't have any chance for that

src/engine.rs
mfornet: Notice that we are not panicking, so we can't use `sdk_unwrap` there. For the name, tbh it is clear to me, we are making emphasis that if unwrap is not possible we will be finishing. 

It is very similar to using `?`. It wasn't done this way, since we should've returned an object of type `Result<Vec<u8>, Vec<u8>>` which is more inefficient, and doesn't look like a nice API.

joshuajbouw:
> > I feel uncertain why we need an `AuroraStackState`, and why the precompiles now need the state implemented? I almost went down this path, but found solutions around not having to.
> 
> The reason is because we need to know what promises were asked to be created (from calls to exit precompiles), but we can't have the precompiles eagerly create the promises because then they cannot be rolled back if the transaction (or sub-call in a transaction) reverts.
> 
> The `AuroraStackState` keeps track of the data needed to construct the promises and only once the entire transaction is finished do we actually create them.

Right, I get the context now. I just saw this PR https://github.com/rust-blockchain/evm/pull/34/files as well. Makes sense. The design still feels a little off to me, but this is critical for our deliverable for the end of the week, so let's just roll with how it is for now. Eventually, maybe something will come up where it can be improved.

Cargo.toml
joshuajbouw: Ah! I missed this. Ok yeah, makes perfect sense.

Makefile
joshuajbouw: I meant more so, on our mainnet deploys, we shouldn't be using `log`. So we need to be careful about that, yes?

src/connector.rs
joshuajbouw: I'm in the other camp if we are for sure that it can't be reachable. Of course, future logic could state otherwise. Perhaps, set the error message as unreachable to give a hint that this is basically unexpected?

src/engine.rs
joshuajbouw: Sure, not critical for this PR anyways.

Makefile
mfornet: If we remove log, debugging errors in production will be hell.

src/precompiles/native.rs
```diff
@@ -144,91 +190,138 @@ impl ExitToEthereum {
...
+            0x0 => {
+                // ETH transfer
+                //
+                // Input slice format:
+                //      eth_recipient (20 bytes) - the address of recipient which will receive ETH on Ethereum
+
+                let eth_recipient: String = hex::encode(input);
+
+                if eth_recipient.len() == 20 {
+                    (
```
sept-en, line 256: The `input` is 20 bytes long, but the `eth_recipient` is 40 bytes long, as this is a hex-encoded address.

```suggestion
                if input.len() == 20 {
                    let eth_recipient: String = hex::encode(input);
                    (
```

mfornet: Good catch, this is why we need end 2 end test before merging! I'm finishing them.

src/precompiles/native.rs
```diff
@@ -144,91 +190,138 @@ impl ExitToEthereum {
...
+                        String::from_utf8(crate::sdk::current_account_id()).unwrap(),
+                        // There is no way to inject json, given the encoding of both arguments
+                        // as decimal and hexadecimal respectively.
+                        crate::prelude::format!(
+                            r#"{{"amount": "{}", "recipient": "{}"}}"#,
+                            context.apparent_value.as_u128(),
+                            eth_recipient
+                        ),
```
sept-en, line 264: `withdraw()` in Aurora accepts Borsh-serialized args, so you should pass something like this to the function (`eth_recipient` bytes + `value` as LE bytes):

```rust
let mut borsh_args: Vec<u8> = Vec::new();
borsh_args.extend_from_slice(input);
borsh_args.extend_from_slice(&context.apparent_value.as_u128().to_le_bytes());
```

artob:
@mfornet Please fix the merge conflicts and re-request reviews as needed.

src/engine.rs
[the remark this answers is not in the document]
birchmd: Created https://github.com/aurora-is-near/aurora-engine/pull/142

src/json.rs
```diff
@@ -65,13 +65,13 @@ impl JsonValue {
...
-            JsonValue::Object(o) => match o.get(key).ok_or(())? {
-                JsonValue::String(s) => s.parse::<u128>().map_err(|_| ()),
-                _ => Err(()),
+            JsonValue::Object(o) => match o.get(key).ok_or(JsonError::MissingValue)? {
+                JsonValue::String(s) => s.parse::<u128>().map_err(|_| JsonError::InvalidU128),
+                _ => Err(JsonError::InvalidU128),
             },
-            _ => Err(()),
+            _ => Err(JsonError::NotJsonType),
         }
```
sept-en: This function should be removed. `u128` already handles this thing indeed.

src/lib.rs
```diff
@@ -630,49 +630,49 @@ mod contract {
...
+    // }
+
+    // impl<T, E: AsRef<[u8]>> SdkUnwrap<T> for Result<T, E> {
+    //     fn sdk_unwrap(self) -> T {
+    //         match self {
+    //             Ok(t) => t,
+    //             Err(e) => sdk::panic_utf8(e.as_ref()),
+    //         }
+    //     }
+    // }
```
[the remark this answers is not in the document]
mfornet: removed

src/json.rs
mfornet: removed

Cargo.toml
mfornet: The change is already upstream, so poiting to evm 0.28.0 now

src/precompiles/native.rs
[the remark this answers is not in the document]
mfornet: Implemented.

src/engine.rs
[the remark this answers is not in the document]
mfornet: Done.

artob (approved):
@mfornet Please note you have a lot of Clippy warnings and address those in a follow-up PR.

src/engine.rs
```diff
@@ -426,6 +462,151 @@ impl Engine {
...
+    /// with the input, or 0 if tokens were minted successfully.
+    ///
+    /// The output will be serialized as a String
+    /// https://github.com/near/NEPs/discussions/146
+    ///
+    /// IMPORTANT: This function should not panic, otherwise it won't
+    /// be possible to return the tokens to the sender.
+    pub fn receive_erc20_tokens(&mut self, args: &NEP141FtOnTransferArgs) {
+        let str_amount = args.amount.to_string();
+        let output_on_fail = str_amount.as_bytes();
```
sept-en, line 504: Any reason why we use this everywhere as the output on fail?
I wonder if the idea was to have 0 output on success and all other values signalize failure. What happens if the amount is zero?

src/lib.rs
```diff
@@ -312,6 +317,60 @@ mod contract {
...
+        let args: DeployErc20TokenArgs =
+            DeployErc20TokenArgs::try_from_slice(&sdk::read_input()).sdk_expect("ERR_ARG_PARSE");
+
+        let mut engine = Engine::new(predecessor_address()).sdk_unwrap();
+
+        let erc20_contract = include_bytes!("../etc/eth-contracts/res/EvmErc20.bin");
+        let deploy_args = ethabi::encode(&[
+            ethabi::Token::String("Empty".to_string()),
+            ethabi::Token::String("EMPTY".to_string()),
+            ethabi::Token::Uint(ethabi::Uint::from(0)),
```
sept-en, line 356: How these values are expected to be updated later?

src/deposit_event.rs
```diff
@@ -62,4 +61,20 @@ impl DepositedEvent {
...
+            DepositedEvent::event_params(),
+            self.eth_custodian_address,
+            vec![self.sender.to_vec()],
+            vec![
+                ethabi::Token::String(self.recipient.clone()),
+                ethabi::Token::Uint(self.amount.into()),
+                ethabi::Token::Uint(self.fee.clone()),
+            ],
+        )
+    }
```
sept-en, line 79: Any reason to have this? I don't see it's used anywhere. Probably should be removed along with `EthEvent::to_log_entry_data`.

sept-en (commented):
LGTM!

There are some minor comments that I think need to be addressed and resolved.

src/engine.rs
mfornet: If the function fails for any reason we should return the full amount indicating the sender that this amount must be refunded. We return `0` on success indicating that nothing must be refunded.

>  What happens if the amount is zero?
In this case it is a void transfer, so either way (failure or success) nothing should be refunded.

src/lib.rs
mfornet: For now using admin privileges! Later we should create a trustless way to bridge metadata.

src/deposit_event.rs
mfornet: It was useful for me to craft data for testing. Let's keep this here.

I think most of the tests that uses Proofs as very long string objects (that contains a json) should be rewritten creating the actual objects and serializing later. The main reason for this, is that if a test starts failing, debugging is a nightmare for those unfamiliar with this test, given that this data is almost a black box.

src/lib.rs
sept-en: What's the interface for upgrading the `name` and `symbol`?
Are we going to use `adminSstore` each time to change metadata until we create a new trustless way?

src/deposit_event.rs
sept-en: Got it, then sure, let's keep this!

src/lib.rs
mfornet: yes, for now admin Sstore

PR #53, opened 2021-05-05
Add custom erc20 contract to the EVM.

mfornet:
To compile the EVM it is required to compile ERC20 first if this change is used. Do this using the following steps:

```
cd eth-contracts
yarn
yarn compile
```

This will generate relevant binary in:
`eth-contracts/res/EvmErc20.bin`


artob:
@mfornet Don't introduce new top-level directories, put `eth-contracts` under either `src` or `etc`.

mfornet:
> @mfornet Don't introduce new top-level directories, put `eth-contracts` under either `src` or `etc`.

Sounds good to me. Actually I was thinking if we should even keep this solidity contract here or in separate repository and only pull binary.

I figured out it is better to have it here, since it is a small contract designed for be deployed as template by the evm.

Note that this includes those bytes in the evm binary, so the size will increase considerably.

joshuajbouw:
Looks like CI needs to be updated to accommodate this?

joshuajbouw:
> @mfornet Don't introduce new top-level directories, put `eth-contracts` under either `src` or `etc`.
I wouldn't put it under `src` as those are separate compiles, `etc` sounds better to me.

src/lib.rs
```diff
@@ -262,8 +262,7 @@ mod contract {
 
         let mut engine = Engine::new(predecessor_address());
 
-        // TODO(#51): Use proper erc20_contract
-        let erc20_contract = Default::default();
+        let erc20_contract = include_bytes!("../eth-contracts/res/EvmErc20.bin");
```
joshuajbouw, line 265: I believe that our CI needs to be updated too for this?

joshuajbouw (changes requested):

src/lib.rs
mfornet: Done

joshuajbouw:
Check tests

joshuajbouw:
And lints :D

birchmd:
I created an issue for myself to make use of this contract in the benchmarks after it is merged instead of the generic ERC20. We might as well measure the implementation we are actually going to use.

artob:
@mfornet Looking forward to merging this. Please reassign to me once CI is showing green and it's ready to go.

mfornet:
> @mfornet Looking forward to merging this. Please reassign to me once CI is showing green and it's ready to go.

@artob notice I'm merging this into #51 which is not ready.

artob:
> And lints :D

Let's be clear that these precompiles are blockers for implementing and validating the connectors and hence blockers the release, and nice-to-haves like lints are a low priority right now.

joshuajbouw (approved):
Looks good.

PR #59, opened 2021-05-05
ETH connector

mrLSD:
ETH connector contains fugible token logic for specific flow.
To verify logic:
* [DIagram](https://hackmd.io/Ttg3QQktRc2-5NHUAGIrow?view) - represent Diagram of algorithm and process flow
* [Formal specification](https://hackmd.io/q7uV5_EKSC6YQ753YmC3_w?edit) - represent algorithm in a formal way as functional style code with full type inference.


src/connector.rs
```diff
@@ -0,0 +1,630 @@
...
+        let predecessor_account_id = String::from_utf8(sdk::predecessor_account_id()).unwrap();
+        let current_account_id = String::from_utf8(sdk::current_account_id()).unwrap();
+        // Parse message with specific rules
+        let message_data = self.parse_on_transfer_message(&args.msg);
+
+        // Special case when current_account_id is predecessor
+        if current_account_id == predecessor_account_id {
+            self.burn_near(current_account_id, args.amount);
+            self.mint_eth(message_data.recipient, args.amount);
+        } else {
```
sept-en, line 553: We need to pay the relayer's fee even if we are making a deposit to nETH.

src/connector.rs
```diff
@@ -0,0 +1,630 @@
...
+            BalanceOfEthCallArgs::try_from_slice(&sdk::read_input()[..]).expect(ERR_FAILED_PARSE);
+        let balance = self.ft.internal_unwrap_balance_of_eth(args.address);
+        #[cfg(feature = "log")]
+        sdk::log(&format!(
+            "Balance of ETH [{}]: {}",
+            hex::encode(args.address),
+            balance
+        ));
+        sdk::return_output(&balance.to_string().as_bytes());
+    }
```
sept-en: As I mentioned earlier, this method duplicates the already existing [`get_balance`](https://github.com/aurora-is-near/aurora-engine/blob/95c80ff003f607f162c0b750f4551eb3a4b29aa4/src/lib.rs#L282) method. Probably we need to get rid of one of those.
@artob what do you think?

artob:
@evgenykuzyakov Please do a security review of the ETH connector if you have a chance.

src/connector.rs
```diff
@@ -537,7 +520,11 @@ impl EthConnectorContract {
             self.mint_eth(message_data.recipient, args.amount - fee);
             // Mint fee to relayer
             if fee > 0 {
-                let evm_relayer_addres = self.get_evm_relayer_address(&message_data.relayer);
+                let evm_relayer_addres: EthAddress = engine
```
joshuajbouw, line 526: Just noting spelling

src/connector.rs
```diff
@@ -537,7 +520,11 @@ impl EthConnectorContract {
...
+                    .get_relayer(&message_data.relayer.as_bytes())
+                    .expect("ERR_WRONG_RELAYER_ID")
+                    .0;
+                //let evm_relayer_addres = self.get_evm_relayer_address(&message_data.relayer);
```
joshuajbouw, line 530: dead code

joshuajbouw:
Just a general note, the structure of everything makes sense, as generally, you must require:
* Proof of transfer on ETH
* Ensure that it is indeed the right person making the transfer
* That it is the correct custodian
* Give value

Then on the withdraws:
* Ensure information is all correct to the extent we can on NEAR
* Burn value

This is just a very basic flow of how you need to do it. It was great that I gave myself some time to do something else over the weekend to clear my head up a bit. The second read through that I did on Sunday and today makes me feel more comfortable with this change. Besides the previous issues noted before, which can be fixed up during refactoring in the future, the functional logic looks quite fine to me.

Lets do that final call today or when you guys can, then I'll give it my thumbs up even though I'm happy enough with it as is.

sept-en:
> Just a general note, the structure of everything makes sense, as generally, you must require:
> 
>     Proof of transfer on ETH
>     Ensure that it is indeed the right person making the transfer
>     That it is the correct custodian
>     Give value
@joshuajbouw thanks for the summary! the only thing here worth mentioning is that we don't need to check for the second item - anyone can upload any proof from the Ethereum to Near. But only the person which account is encoded in the Proof will receive the appropriate amount specified in the Proof.

joshuajbouw:
@sept-en Sorry, yes. Was writing those off the top of my memory.

src/connector.rs
```diff
@@ -0,0 +1,563 @@
...
+pub struct OnTransferMessageData {
+    pub relayer: AccountId,
+    pub recipient: EthAddress,
+    pub fee: U256,
+}
+
+impl EthConnectorContract {
+    pub fn get_instance() -> Self {
+        Self {
+            contract: sdk::get_contract_data(CONTRACT_NAME_KEY),
```
birchmd, line 51: Much like https://github.com/aurora-is-near/aurora-engine/pull/62#discussion_r627449415 I suggest we use the `KeyPrefix` enum rather than ad-hoc strings to create unique storage keys.

For example we could add `KeyPrefix::ContractData` and then change the definition of `sdk::get_contract_data` to be something like

```rust
enum ContractId {
    EthConnector = 0x0,
    FungibleToken = 0x1,
}

fn get_contract_data<T: BorshDeserialize>(contract: ContractId) -> T {
    let key = [KeyPrefix::ContractData as u8, contract as u8];
    let data = read_storage(&key).expect("Failed read storage");
    T::try_from_slice(&data).unwrap()
}
```

birchmd (commented):
Some comments (I haven't gotten very far yet, so there will probably be more, but I wanted to get these out there so fixes can be pushed quickly)

src/connector.rs
mrLSD: @birchmd changed to your proposal

src/connector.rs
```diff
@@ -0,0 +1,593 @@
...
+            if fee > 0 {
+                let evm_relayer_address: EthAddress = engine
+                    .get_relayer(&message_data.relayer.as_bytes())
+                    .expect("ERR_WRONG_RELAYER_ID")
+                    .0;
+                self.mint_eth(evm_relayer_address, fee);
+            }
+        } else {
+            // Implement new scheme when PR #62 is merged
+            todo!();
```
birchmd, line 553: Just commenting to call this out. #62 is now merged, so this should be updated according to the comment.

src/connector.rs
birchmd: Thanks for getting that change in so quickly!

tests/test_connector.rs
```diff
@@ -0,0 +1,458 @@
...
+
+    let balance = total_supply_near(&master_account, CONTRACT_ACC);
+    assert_eq!(balance, DEPOSITED_AMOUNT - transfer_amount);
+
+    let balance = total_supply_eth(&master_account, CONTRACT_ACC);
+    assert_eq!(balance, transfer_amount);
+}
+
+#[test]
+#[should_panic]
```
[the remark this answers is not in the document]
mrLSD: Returned previous version for that test

src/connector.rs
```diff
@@ -0,0 +1,593 @@
...
+        T::try_from_slice(&data[..]).unwrap()
+    }
+
+    /// Init eth-connector contract specific data
+    pub fn init_contract() {
+        // Check is it already initialized
+        assert!(
+            !sdk::storage_has_key(&Self::get_contract_key(&EthConnectorKeyPrefix::Contract)),
+            "ERR_CONTRACT_INITIALIZED"
+        );
```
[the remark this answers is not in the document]
joshuajbouw: Doesn't hurt to add that.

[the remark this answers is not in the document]
sept-en: @artob @mrLSD please take a look, this is important.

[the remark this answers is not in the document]
mrLSD: @sept-en I removed the checker some time ago because init Aurora itself going in the same way.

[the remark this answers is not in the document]
sept-en: @mrLSD I think we need to have it here

birchmd:
> > Some comments about the tests. With a big PR like this I think having good test coverage is important because it's hard for me to see how all the pieces will fit together into the desired functionality, but tests help make this more clear and also show it working as intended.
> 
> Do you have concrete proposals? I can add additional tests.

Sorry, I didn't meant to imply the tests were inadequate (I don't necessarily think that is the case). I only wanted to justify why I was bothering to focus some reviewing attention specifically on the tests.

joshuajbouw:
IMPORTANT NOTE: All uses of KeyPrefix are through either bytes_to_key and address_to_key as I found no instance of manual key creation on master. This needs to be checked in this branch else there will be key storage issues!

See #77

Makefile
```diff
@@ -1,6 +1,6 @@
 CARGO = cargo
 NEAR  = near
-FEATURES = contract
+FEATURES = contract,log
```
joshuajbouw, line 3: This shouldn't be on by default.

src/connector.rs
joshuajbouw: I don't think this is resolved, as its still a todo?

Makefile
birchmd: There was a [discussion about this on Slack](https://near-protocol.slack.com/archives/C020JMC9U15/p1620667119070500?thread_ts=1620661443.048500&cid=C020JMC9U15).

Summary is: if something goes wrong in production there's basically no way to know what happened without logging and it's not that expensive to have it (@mrLSD measured it to be ~0.5 Tgas which is much much less than any operation currently).

src/connector.rs
```diff
@@ -0,0 +1,582 @@
...
+            self.burn_near(current_account_id, args.amount);
+            self.mint_eth(message_data.recipient, args.amount - fee);
+            // Mint fee to relayer
+            if fee > 0 {
+                let evm_relayer_address: EthAddress = engine
+                    .get_relayer(&message_data.relayer.as_bytes())
+                    .expect("ERR_WRONG_RELAYER_ID")
+                    .0;
+                self.mint_eth(evm_relayer_address, fee);
+            }
```
sept-en, line 546: This part works incorrectly and actually is critical. Currently, this works like that:

1) If the fee is zero, the user receives the full amount, and the transaction is "Succeeded".
2) If the fee is greater than thero:
    a) Query for the registered relayer, the relayer is registered. Then mint appropriate amount of fee to the relayer. The transaction is "Succeeded".
    b) Query for the registered relayer, the relayer is NOT registered. So we panic with "ERR_WRONG_RELAYER_ID", the transaction is "Failed". The user is not receiving his money, the proof was recorded as used on the previous steps so it's not possible to finish the deposit for this transaction any more. So basically, it's easy to burn any produced proof by not registering the relayer.

The suggested solution for that is the following:
* Make sure we never ever panic in the `ft_on_transfer` execution (at least for the ETH branch), as the proof was recorded on the previous steps so we won't be able to restore it.
* If the `fee > 0`, but there's no registered relayer for the predecessor account, mint the fee to the recipient as well. If we don't mint the fee for the recipient, this amount won't be minted and there's no sense in it.

sept-en: @mrLSD @joshuajbouw @artob

joshuajbouw: Note also the todo

joshuajbouw: Hugely important on what you said. If there is a proof, we need to be able to use it on the previous steps. They must ALWAYS be restorable, only once, of course.

sept-en: There's no way to straight-forwardly revert the chain of promises at the moment. So this is correct that the proof is recorded in the previous step and only after that the tokens are minted. This protects from proof reusage and possible reentrancy attempts. We just need to try making all possible checks before storing the proof and don't let it panic at a later stage.

mrLSD: Good point!
@joshuajbouw FYI it's not related to Proof. We can reproduce it without `Deposit` call because ft_transfer_call is a public method.

mrLSD: @sept-en @jakestutzman I fixed that issue. Also, I placed an early checker for all possible assertions in ft_on_transfer.

src/connector.rs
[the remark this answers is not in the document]
sept-en: @artob What do you think?

mrLSD:
#98 #104

src/fungible_token.rs
```diff
@@ -0,0 +1,452 @@
...
+
+    pub fn internal_storage_balance_of(&self, account_id: &str) -> Option<StorageBalance> {
+        if self.accounts_contains_key(account_id) {
+            Some(StorageBalance {
+                total: self.storage_balance_bounds().min,
+                available: 0,
+            })
+        } else {
+            None
+        }
```
sept-en, line 366: This is wrong as according to the NEP-145 specs: `If the account is unknown to the contract then the total account storage balance returned will be zero.`
So we need to always return `StorageBalance`, and zero for unknown accounts.

src/fungible_token.rs
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    pub fn storage_balance_of(&self, account_id: &str) -> Option<StorageBalance> {
+        self.internal_storage_balance_of(account_id)
```
sept-en, line 370: Return just `StorageBalance` instead of `Option` (see above).

sept-en (changes requested):

src/fungible_token.rs
mrLSD: Fixed at #98

src/fungible_token.rs
mrLSD: Fixed at #98

src/fungible_token.rs
sept-en: Please note the appropriate [issue](https://github.com/near/near-sdk-rs/issues/425) in `near-sdk-rs`.

src/storage.rs
```diff
@@ -16,7 +16,16 @@ pub enum KeyPrefix {
...
+    EthConnector = 0x6,
+}
+
+/// Enum used to differentiate different storage keys used by eth-connector
+#[derive(Clone, Copy, BorshSerialize, BorshDeserialize)]
+pub enum EthConnectorStorageId {
+    Contract = 0x0,
+    FungibleToken = 0x1,
+    UsedEvent = 0x2,
+    StatisticsAuroraAccountsCounter = 0x3,
```
sept-en, line 28: It makes sense to start statistics Id from something like `0xF` or `0x10`. So if other fields related to the contract itself are added in the future, they will come in order and not going to be mixed with Ids for Statistics.

I'll create a PR for that.

PR #60, opened 2021-05-05
Sanity tests

birchmd:
This PR introduces sanity tests which run as part of `cargo test` (and therefore CI). These should be helpful in ensuring we do not have correctness regressions as we go changing things (e.g. in https://github.com/aurora-is-near/aurora-engine/pull/57 ).

The PR looks large because I reorganized (with the help of matklad and [his blog post](https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html)) the benchmark code so that it could be reused in these sanity tests. The only new code is in `sanity.rs`, everything else is just moving things around.

This PR only covers basic eth transfers and nonces. In a future PR I will add contract calls, including standard precompile and ERC20 (both already used in benchmarks).


src/benches/eth_erc20.rs
```diff
@@ -1,20 +1,22 @@
...
+use crate::test_utils::{
+    address_from_secret_key, deploy_evm, sign_transaction, AuroraRunner, SUBMIT,
+};
 
 const INITIAL_BALANCE: u64 = 1000;
 const INITIAL_NONCE: u64 = 0;
 const TRANSFER_AMOUNT: u64 = 67;
 
-fn eth_erc20_benchmark(c: &mut Criterion) {
+pub fn eth_erc20_benchmark(c: &mut Criterion) {
```
matklad, line 19: This can now be `pub(crate)` I think? In general, I personally have two pronged approach to visibility in Rust:

* I try very hard not to make things `pub` if I can
* I don’t really care if things are private, pub(super) or pub(crate) otherwise


In large multi-crate projects, using pub solely for cross-crate interfaces improves readability a lot.

src/benches/mod.rs
```diff
@@ -0,0 +1,19 @@
+use criterion::Criterion;
+
+mod eth_deploy_code;
+mod eth_erc20;
+mod eth_standard_precompiles;
+mod eth_transfer;
+
+#[test]
+#[ignore] // We don't want to run in CI, so ignore. To run locally use `cargo test -- --ignored`
```
matklad, line 9:
```suggestion
#[ignore] // We don't want to run in CI, so ignore. To run locally use `cargo test --release -- --ignored`
```

matklad: I”m totally going to steal the ignore trick for rust-analyzer!

artob (approved):
LGTM :+1: Great work! Merging this right away as we urgently need this.

For your eventual follow-up PR note also @matklad's review comments.

Also, note that a number of source files' modes changed from 100644 to 100755 (executable bit set). I imagine that was unintentional, in which case please correct that in the follow-up PR.

PR #61, opened 2021-05-06
Fix transaction decoding

joshuajbouw:
This is an issue that came up during discovering a problem that @artob  found when decoding an RLP encoded signed transaction.


artob (approved):
LGTM, after [fixing](https://github.com/aurora-is-near/aurora-engine/pull/61/commits/224d5af10a5c0a9d5cf9bacb02fe89b087a6b050) one nit :+1: Thanks for this.

PR #62, opened 2021-05-06
Add common code between #51 and #59

mfornet:


artob:
@birchmd @joshuajbouw Please prioritize review of this PR, we need to merge it soonest.

src/engine.rs
```diff
@@ -138,6 +141,8 @@ pub struct EngineState {
     pub bridge_prover_id: AccountId,
     /// How many blocks after staging upgrade can deploy it.
     pub upgrade_delay_blocks: u64,
+    /// Mapping between relayer account id and relayer evm address
+    pub relayers_account: LookupMap,
```
sept-en, line 145: `pub relayers_evm_addresses: LookupMap,`

src/engine.rs
```diff
@@ -147,6 +152,7 @@ impl From<NewCallArgs> for EngineState {
             owner_id: args.owner_id,
             bridge_prover_id: args.bridge_prover_id,
             upgrade_delay_blocks: args.upgrade_delay_blocks,
+            relayers_account: LookupMap::new(RELAYERS_ACCOUNT_PREFIX.to_vec()),
```
sept-en, line 155: `relayers_evm_addresses: LookupMap::new(RELAYERS_EVM_ADDRESSES_PREFIX.to_vec()),`

src/engine.rs
```diff
@@ -452,6 +458,19 @@ impl Engine {
         let state = MemoryStackState::new(metadata, self);
         StackExecutor::new_with_precompile(state, &CONFIG, precompiles::istanbul_precompiles)
     }
+
+    pub fn register_relayer(&mut self, account_id: &[u8], evm_address: Address) {
+        self.state
+            .relayers_account
```
sept-en, line 464: `.relayers_evm_addresses`

src/engine.rs
```diff
@@ -452,6 +458,19 @@ impl Engine {
...
+            .insert_raw(account_id, evm_address.as_bytes());
+    }
+
+    pub fn get_relayer(&self, account_id: &[u8]) -> Option<Address> {
+        self.state
+            .relayers_account
```
sept-en, line 470: `.relayers_evm_addresses`

src/engine.rs
```diff
@@ -124,6 +125,8 @@ impl ExitIntoResult for ExitReason {
     }
 }
 
+pub(crate) const RELAYERS_ACCOUNT_PREFIX: &[u8] = b"r";
```
sept-en, line 128: `pub(crate) const RELAYERS_EVM_ADDRESSES_PREFIX: &[u8] = b"r";`

src/lib.rs
```diff
@@ -259,6 +261,29 @@ mod contract {
         result.map(|_f| Vec::new()).sdk_process();
     }
 
+    #[no_mangle]
+    pub extern "C" fn register_relayer() {
+        let relayer_address = sdk::read_input();
```
sept-en, line 266: We need to check this is exactly 20 bytes long

src/lib.rs
```diff
@@ -259,6 +261,29 @@ mod contract {
...
+    pub extern "C" fn ft_on_transfer() {
+        let predecessor_account_id = String::from_utf8(sdk::predecessor_account_id()).unwrap();
+        let current_account_id = String::from_utf8(sdk::current_account_id()).unwrap();
+
+        if current_account_id == predecessor_account_id {
+            // TODO(#59) ETH transfer
+        } else {
+            // TODO(#51) ERC20 transfer
+        }
+    }
```
joshuajbouw, line 285: Obviously, clippy is going to scream about this. Can you please instead add in `todo!()` in both?

src/engine.rs
```diff
@@ -452,6 +458,19 @@ impl Engine {
         let state = MemoryStackState::new(metadata, self);
         StackExecutor::new_with_precompile(state, &CONFIG, precompiles::istanbul_precompiles)
     }
+
+    pub fn register_relayer(&mut self, account_id: &[u8], evm_address: Address) {
+        self.state
+            .relayers_account
+            .insert_raw(account_id, evm_address.as_bytes());
+    }
+
```
joshuajbouw, line 467: Add `#[allow(dead_code)]` here, don't forget to remove it when used.

sept-en (commented):
LGTM! The most important thing to adjust is the address length verification in `register_relayer()`.

src/lib.rs
```diff
@@ -259,6 +261,29 @@ mod contract {
...
+            sdk::predecessor_account_id().as_slice(),
+            Address(relayer_address.as_slice().try_into().unwrap()),
+        );
+    }
+
+    /// Allow receiving NEP141 tokens to the EVM contract
+    #[no_mangle]
+    pub extern "C" fn ft_on_transfer() {
+        let predecessor_account_id = String::from_utf8(sdk::predecessor_account_id()).unwrap();
+        let current_account_id = String::from_utf8(sdk::current_account_id()).unwrap();
```
joshuajbouw, line 278: If these are 100% infallible, you can instead use `from_utf8_unchecked` as it is coming from the sdk. It'll reduce the cost of the tx.

joshuajbouw (changes requested):

src/engine.rs
birchmd, line 128: I worry a little bit that if we take an ad-hoc approach to key prefixes that we may end up with an accidental collision between keys one day (because the prefix+data of one key happens to match the prefix+data of another key, even if the prefixes and data are different).

I would prefer if we added a new variant to the `KeyPrefix` enum already used in `engine.rs`. If we take a unified approach to key prefixes then a collision will be impossible (because the prefixes will all be the same length and all distinct).

src/map.rs
```diff
@@ -0,0 +1,54 @@
...
+use crate::sdk;
+
+pub(crate) fn append_slice(id: &[u8], extra: &[u8]) -> Vec<u8> {
+    [id, extra].concat()
+}
+
+/// An non-iterable implementation of a map that stores its content directly on the trie.
+#[derive(BorshSerialize, BorshDeserialize, Default)]
+pub struct LookupMap {
+    key_prefix: Vec<u8>,
```
birchmd, line 13: Related to my comment above, you might want to make this a `KeyPrefix` type instead of arbitrary bytes.

src/map.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub fn contains_key_raw(&self, key_raw: &[u8]) -> bool {
+        let storage_key = self.raw_key_to_storage_key(key_raw);
+        sdk::storage_has_key(&storage_key)
+    }
+
+    /// Returns the serialized value corresponding to the serialized key.
+    #[allow(dead_code)]
+    pub fn get_raw(&self, key_raw: &[u8]) -> Option<Vec<u8>> {
+        let storage_key = self.raw_key_to_storage_key(key_raw);
+        sdk::read_storage(&storage_key)
```
birchmd, line 38: Related to all this key business, I wonder if we should change the `sdk` API to take a `KeyPrefix` and then arbitrary bytes. This ensure we always take the uniform approach to storage I mentioned above, and also means we don't need to duplicate concatenation logic in multiple places throughout the code.

src/lib.rs
birchmd, line 278: Nit: these probably don't need to be strings if all we are going to do is compare them on the next line.

birchmd (commented):
I have a suggestion about how we handle key prefixes. I won't block the PR because I understand there is a time crunch here, but I strongly recommend we make the suggested change because having to go back later to migrate the internal storage would be a real pain.

src/map.rs
mfornet: I will leave this outside of this PR, but agree with your idea.

src/lib.rs
mfornet: Good catch. Also notice that relayer can call this function any time and change it.

src/lib.rs
sept-en: @birchmd this is only a template function at the moment. The strings will be used in the implementation (when me merge this PR into our PRs)

mfornet: I skip this with a suggestion from @birchmd

src/engine.rs
joshuajbouw: Going to have to absolutely agree with that.

src/lib.rs
mfornet: I think we should do this inside `sdk::*_account_id()` given that we don't want to use unsafe for now here

mfornet: https://github.com/aurora-is-near/aurora-engine/pull/62#discussion_r627506198

joshuajbouw (approved):
Looks good to me.

sept-en (approved):

birchmd (approved):

src/map.rs
birchmd: Sounds good. If we use `KeyPrefix` throughout then changing the API later will be a trivial refactoring

src/engine.rs
```diff
@@ -147,6 +150,7 @@ impl From<NewCallArgs> for EngineState {
             owner_id: args.owner_id,
             bridge_prover_id: args.bridge_prover_id,
             upgrade_delay_blocks: args.upgrade_delay_blocks,
+            relayers_evm_addresses: LookupMap::new(KeyPrefix::RelayerEvmAddressMap),
```
mrLSD, line 153: We don't need to store the key in storage. It's constant, so it's redundant and read/store same information to Engine state without need.

mfornet: TIL that `const generic` is available in rust, and use it for the first time to solve this, it was neat.

src/storage.rs
```diff
@@ -1,12 +1,38 @@
...
     Balance = 0x2,
     Code = 0x3,
     Storage = 0x4,
+    RelayerEvmAddressMap = 0x5,
+}
+
+/// We can't use const generic over Enum, but we can do it over integral type
+pub type KeyPrefixU8 = u8;
+
+// TODO: Derive From<u8> using macro to avoid missing new arguments in the future
```
birchmd, line 18: If this doesn't get resolved in the larger connector effort then we should file an issue for it. I'm not a big fan of this brittle-looking match. Maybe we should have a unit test that round-trips all `KeyPrefix` variants through `u8` and back to make sure we don't hit the `unreachable`.

Would it be too much overhead to use borsh here instead of writing the match explicitly?

mrLSD (approved):
Great job!

birchmd (approved):
Using const generics here is super cool!

Though, to be honest, I'm not sure this is better than what we had before. As I comment below, I don't love the new match statement to map the const parameter back to an enum variant. I also am not sure whether we gained anything here with respect to not having things in storage (re https://github.com/aurora-is-near/aurora-engine/pull/62#discussion_r627600767 ).

Even though the parameter is a constant, it needs to be defined _somewhere_ and that either is in the contract state, or in the contract source. The former uses on-chain storage, which need to read from on each invocation to initialize the Engine object. The latter ends up in the contract bytecode, so we incur the cost on deploying the contract, and on loading it into the runtime. I'm not sure which is more expensive since my knowledge of the Near runtime is not deep enough (especially with contract caching which I think is being worked on and would mean each transaction would not necessarily re-load the bytecode into the runtime).

I also doubt if it matters much either way because we're talking about one byte of storage essentially (actually more in the version where we use const generics because now we have the match to convert `u8` to `KeyPrefix` which will need to exist in the bytecode).

Long story short, I think I preferred the way it was before (with a simple runtime value) because I think it is less complex and doesn't require a kind of hacky conversion, and I don't think we actually saved anything by adding the complexity. But I'll still approve the PR because I don't think it's worth fighting about either, whether we keep this version or the old one is up to you.

src/storage.rs
mfornet: > Would it be too much overhead to use borsh here instead of writing the match explicitly?

Using borsh might not be a good idea, since I believe borsh will see those fields as `isize` so the encoding will have 5 bytes, and will incur in some (not big but) unnecessary overhead.

I'll add the macro to derive this in separate PR to merge this now.

sept-en (approved):
LGTM! Well done @mfornet !

PR #63, opened 2021-05-07
More simple Engine tests

birchmd:
As a follow up to https://github.com/aurora-is-near/aurora-engine/pull/60 we add more tests. These are focused on contract execution, specifically ERC20 and standard pre-compiles.

Once again, I rearranged a bunch of code to facilitate reuse between benchmarks and tests, so this PR looks bigger than it is. The new tests are in

* `src/tests/erc20.rs`
* `src/tests/standard_precompiles.rs`

so I would recommend reviews focus on those two files.

This PR also fixes the comments that were made on #60 including the weird file mode change (not sure how that happened) and the handful of comments from matklad


artob (approved):
LGTM, and much needed :+1:

Issue #64, opened 2021-05-09
Understanding precompiles return value


mfornet:
It is hard to tell from [this tuple](https://github.com/aurora-is-near/aurora-engine/blob/master/src/precompiles/mod.rs#L19): `type PrecompileResult = Result<(ExitSucceed, Vec<u8>, u64), ExitError>;` what does the last value (`u64`) means. I'm assuming that second value `Vec<u8>` is the output.

Could you clarify what does the last value means?

Also consider changing this tuple to a named struct and/or add some comments documenting this.

joshuajbouw:
Yeah, so this is the interface from the Sputnik VM itself. We can not exactly change it. However, that last value should be `cost`, which is an issue, as we are returning 0 for now, but we also are not charging gas at the moment. The cost is returned, and then it is recorded.

This certainly should be fixed.

PR #65, opened 2021-05-10
Return precompile cost instead of 0.

joshuajbouw:
@mfornet wanted to know [what the last value of a return was](https://github.com/aurora-is-near/aurora-engine/issues/64) pertaining to the pre-compiles. This led to the understanding that a bug needs to be fixed with the correct cost of the precompile and not just a return of 0.

This closes https://github.com/aurora-is-near/aurora-engine/issues/64.


artob:
@joshuajbouw Does this pass all the sanity checks added lately?

PR #72, opened 2021-05-11
A more acceptable return for `Engine::block_hash`

joshuajbouw:
As per Ethereum VM opcodes, the block hash returns the block hash of the last 256 blocks, excluding the current one. Otherwise, it will return `0x0`. 

Unfortunately, we do not have the capability to return what the Ethereum VM wants, but we can at least give the best that we can for now. Simply, the requirement of returning `0x0` if we are on the current block or beyond 256 blocks. Otherwise, we can return the max value of an unsigned 256-bit integer.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -421,7 +421,7 @@ impl evm::backend::Backend for Engine {
     /// See: https://doc.aurora.dev/develop/compat/evm#blockhash
     fn block_hash(&self, number: U256) -> H256 {
         let idx = U256::from(sdk::block_index());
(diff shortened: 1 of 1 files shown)


birchmd:
I'm not so sure about this change. I think always returning zero and saying it's simply unsupported is better than saying "it's kinda supported -- no it doesn't work for the interesting inputs, but for out of range inputs it gives the right answer".

The latter is technically closer in compatibility, but a much more complicated story. And I don't think this small step towards more compatibility adds enough value to offset the awkward messaging.

PR #73, opened 2021-05-11
Base precompile code between connectors

mfornet:
Add precompiles, so eth-connector and erc20-connector can work from it.

### DIFF
--- a/src/precompiles/native.rs
+++ b/src/precompiles/native.rs
@@ -1,7 +1,7 @@
 use evm::{Context, ExitError, ExitSucceed};
 
 use super::{Precompile, PrecompileResult};
-use crate::prelude::{String, ToString, Vec, U256};
+use crate::prelude::{Cow, String, ToString, Vec, U256};
 use crate::types::AccountId;
 
 mod costs {
@@ -23,7 +23,7 @@ mod costs {
 /// Get the current nep141 token associated with the current erc20 token.
 /// This will fail is none is associated.
 fn get_nep141_from_erc20(_erc20_token: &[u8]) -> AccountId {
-    // TODO:
+    // TODO(#51): Already implemented
     "".to_string()
 }
 
@@ -103,9 +103,9 @@ impl Precompile for ExitToNear {
                     ),
                 )
             } else {
-                // TODO: Which is the expected result in case of failure
-                //  Stopped vs Returned vs Suicided
-                return Ok((ExitSucceed::Stopped, Vec::new(), 0));
+                return Err(ExitError::Other(Cow::from(
+                    "ERR_INVALID_RECEIVER_ACCOUNT_ID",
+                )));
             }
         } else {
             // ERC20 transfer
@@ -133,9 +133,9 @@ impl Precompile for ExitToNear {
                     ),
                 )
             } else {
-                // TODO: Which is the expected result in case of failure
-                //  Stopped vs Returned vs Suicided
-                return Ok((ExitSucceed::Stopped, Vec::new(), 0));
+                return Err(ExitError::Other(Cow::from(
+                    "ERR_INVALID_RECEIVER_ACCOUNT_ID",
+                )));
             }
         };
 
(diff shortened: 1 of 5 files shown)


sept-en (approved):
Added some comments (we already discussed some of them on the call with you), but I'm fine if we don't address them in the current PR to speed up the process. Noted here just for a future reference if this won't be fixed right now.

Looks good to me, great job!

src/sdk.rs
```diff
@@ -278,6 +278,26 @@ pub fn predecessor_account_id() -> Vec<u8> {
...
+
+#[allow(dead_code)]
+pub fn signer_account_pk() -> Vec<u8> {
+    unsafe {
+        exports::signer_account_pk(1);
+        let bytes: Vec<u8> = vec![0u8; exports::register_len(1) as usize];
+        exports::read_register(1, bytes.as_ptr() as *const u64 as u64);
+        bytes
+    }
+}
```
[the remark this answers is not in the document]
mfornet: We dont!

src/types.rs
```diff
@@ -10,6 +11,11 @@ pub type AccountId = String;
 pub type RawAddress = [u8; 20];
 pub type RawU256 = [u8; 32]; // Little-endian large integer type.
 pub type RawH256 = [u8; 32]; // Unformatted binary data of fixed length.
+pub type Gas = u64;
+pub type Balance = u128;
+
```
mrLSD, line 16: Just notice that we have the same type et `eth-connector`

mfornet: Let's have it here since they are common to both

src/precompiles/native.rs
```diff
@@ -0,0 +1,267 @@
...
+                    crate::prelude::format!(
+                        r#"{{"receiver_id": "{}", "amount": "{}", "memo": null}}"#,
+                        String::from_utf8(input.to_vec()).unwrap(),
+                        context.apparent_value.as_u128()
+                    ),
+                )
+            } else {
+                // TODO: Which is the expected result in case of failure
+                //  Stopped vs Returned vs Suicided
+                return Ok((ExitSucceed::Stopped, Vec::new(), 0));
```
mfornet, line 108: What is the expected PrecompileResult if it should fail?

Should we return a flag instead `1` on success, and `0` on failure?
```
return Ok((ExitSucceed::Returned, [flag].to_vec(), 0));
```

mfornet: From what I remember using `sha` precompile, when you called it, the result is stored in the range from [0, 32]
However, it also returns a boolean value, whether it succeeded or not. How can I return this boolean value from here?

sept-en: I think we should use `ExitError` for that purpose

src/precompiles/native.rs
```diff
@@ -0,0 +1,267 @@
...
+    pub(super) const FT_TRANSFER_GAS: Gas = 100_000_000_000_000;
+
+    // TODO(#51): Determine the correct amount of gas
+    pub(super) const WITHDRAWAL_GAS: Gas = 100_000_000_000_000;
+}
+
+/// Get the current nep141 token associated with the current erc20 token.
+/// This will fail is none is associated.
+fn get_nep141_from_erc20(_erc20_token: &[u8]) -> AccountId {
+    // TODO:
```
mrLSD, line 26: Where it will TODO? If not planned now we should create an issue.

src/precompiles/native.rs
```diff
@@ -0,0 +1,267 @@
...
+                (
+                    nep141_address,
+                    crate::prelude::format!(
+                        r#"{{"amount": "{}", "recipient": "{}"}}"#,
+                        amount,
+                        eth_recipient
+                    ),
+                )
+            } else {
+                // TODO: Which is the expected result in case of failure
```
mrLSD, line 230: Maybe for that cases, we can add log message?

src/precompiles/native.rs
mfornet: it is done in my pr, didnt remove for it to compile

artob:
@mfornet Is this ready to end up on chain on MainNet tomorrow? If not, please retarget it at the `develop` branch instead of `master`.

PR #74, opened 2021-05-12
Eugene audit 2021 05 11

evgenykuzyakov:
I've decided to comment inline.

Overall I didn't find anything major. I think the biggest issue is `remove_all_storage`. This is a protocol level decision that led to the removal of prefix deletion or iterations. The recent protocol change also restricted account deletion if the storage usage exceeds 10Kb (beyond the contract). This is done to prevent key iteration.

```rust
    /// Removes all storage for the given address.
    pub fn remove_all_storage(_address: &Address) {
        // FIXME: there is presently no way to prefix delete trie state.
        // NOTE: There is not going to be a method on runtime for this.
        //     You may need to store all keys in a list if you want to do this in a contract.
        //     Maybe you can incentivize people to delete dead old keys. They can observe them from
        //     external indexer node and then issue special cleaning transaction.
        //     Either way you may have to store the nonce per storage address root. When the account
        //     has to be deleted the storage nonce needs to be increased, and the old nonce keys
        //     can be deleted over time. That's how TurboGeth does storage.
    }
```

Without an ability to remove old storage from the account may result in a storage draining in a longer run that is not covered by the protocol yet. I'd recommend addressing this issue asap, which may require a hardfork of existing addresses/contracts.

Potential implementation is:
```rust
#[allow(dead_code)]
pub fn storage_to_key_nonced(address: &Address, storage_nonce: u32, key: &H256) -> [u8; 57] {
    let mut result = [0u8; 57];
    result[0] = KeyPrefix::Storage as u8;
    result[1..21].copy_from_slice(&address.0);
    result[21..25].copy_from_slice(&storage_nonce.to_le_bytes());
    result[25..].copy_from_slice(&key.0);
    result
}
```

### DIFF
--- a/src/sdk.rs
+++ b/src/sdk.rs
@@ -171,6 +171,7 @@ pub fn read_input_arr20() -> [u8; 20] {
     unsafe {
         exports::input(0);
         let bytes = [0u8; 20];
+        // TODO: Is it fine to not check the length of the input register here?
         exports::read_register(0, bytes.as_ptr() as *const u64 as u64);
         bytes
     }
@@ -210,6 +211,7 @@ pub fn read_u64(key: &[u8]) -> Option<u64> {
     unsafe {
         if exports::storage_read(key.len() as u64, key.as_ptr() as u64, 0) == 1 {
             let result = [0u8; 8];
+            // TODO: Are you sure the register length is correct?
             exports::read_register(0, result.as_ptr() as _);
             Some(u64::from_le_bytes(result))
         } else {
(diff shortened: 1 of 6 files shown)


joshuajbouw:
Thanks @evgenykuzyakov for looking into this.

artob:
@evgenykuzyakov Thanks for that! We are addressing your comments both for the current milestone as well as the roadmap going forward.

artob (approved):

src/sdk.rs
```diff
@@ -328,6 +330,8 @@ pub fn self_deploy(code_key: &[u8]) {
         // Remove code from storage and store it in register 1.
         exports::storage_remove(code_key.len() as _, code_key.as_ptr() as _, 1);
         exports::promise_batch_action_deploy_contract(promise_id, u64::MAX as _, 1);
+        // TODO: Call upgrade on the same promise to make sure the state has migrated successfully.
```
birchmd, line 333: @evgenykuzyakov I don't understand what you mean here. Could you elaborate further?

PR #77, opened 2021-05-12
Add storage version prefix.

joshuajbouw:
This is a bit straight forward. It adds a version prefix (starting from `0x7`, being `V1`) to addresses from the get-go.

We are not starting from `0x0` because `0x0` to `0x6` is already in use by test net and beta net. Of course, we could change it to start from `0x0` on the main net, but that would mean that there is no parity across all networks, which is always ideal and would prevent issues later down the line.

`VersionPrefix` doesn't exist in `KeyPrefix` for the reason that we can continue from `0x7` for any other storage prefixes that we may require. It would just be impossible to recover any current data from the `0x0` -> `0x6` prefixes.

Later this needs to be improved in the following ways:
- Current version from a single source.
- Logic to clean up old keys that are not versioned (0x0 -> 0x6) <- possibly?

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -1,6 +1,23 @@
 use crate::prelude::{Address, Vec, H256};
 use borsh::{BorshDeserialize, BorshSerialize};
 
+pub type VersionPrefixU8 = u8;
+
+// NOTE: We start at 0x7 as our initial value as our original storage was not
+// version prefixed and ended as 0x6.
+pub enum VersionPrefix {
+    V1 = 0x7,
+}
+
+impl From<VersionPrefixU8> for VersionPrefix {
+    fn from(value: VersionPrefixU8) -> Self {
+        match value {
+            0x7 => Self::V1,
+            _ => unreachable!(),
+        }
+    }
+}
+
 #[allow(dead_code)]
 #[derive(Clone, Copy, BorshSerialize, BorshDeserialize)]
 pub enum KeyPrefix {
@@ -10,6 +27,7 @@ pub enum KeyPrefix {
     Code = 0x3,
     Storage = 0x4,
     RelayerEvmAddressMap = 0x5,
+    // EthConnector = 0x6,
 }
 
 /// We can't use const generic over Enum, but we can do it over integral type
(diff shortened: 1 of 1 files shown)


artob (approved):
LGTM :+1:

src/storage.rs
```diff
@@ -25,40 +43,44 @@ impl From<KeyPrefixU8> for KeyPrefix {
...
 #[allow(dead_code)]
 pub fn address_to_key(prefix: KeyPrefix, address: &Address) -> [u8; 21] {
     let mut result = [0u8; 21];
-    result[0] = prefix as u8;
-    result[1..].copy_from_slice(&address.0);
+    result[0] = VersionPrefix::V1 as u8;
+    result[1] = prefix as u8;
+    result[2..].copy_from_slice(&address.0);
     result
 }
```
sept-en, line 64: I think we need to have the result 22 bytes long instead (1 byte for version + 1 byte for prefix + 20 bytes for address)

src/storage.rs
```diff
@@ -25,40 +43,44 @@ impl From<KeyPrefixU8> for KeyPrefix {
...
-    result[0] = KeyPrefix::Storage as u8;
-    result[1..21].copy_from_slice(&address.0);
-    result[21..25].copy_from_slice(&storage_nonce.to_le_bytes());
-    result[25..].copy_from_slice(&key.0);
+    result[0] = VersionPrefix::V1 as u8;
+    result[1] = KeyPrefix::Storage as u8;
+    result[2..22].copy_from_slice(&address.0);
+    result[22..26].copy_from_slice(&storage_nonce.to_le_bytes());
+    result[26..].copy_from_slice(&key.0);
     result
```
sept-en, line 84: Like the previous comment, I think In this case the result  should be 58 bytes long (1 byte for version, 1 byte for prefix, 20 bytes for address, 4 bytes for nonce, 32 bytes for key)

artob: @joshuajbouw Yup, needs updating.

src/storage.rs
joshuajbouw: This is why we should always `[2..21]` as I had warned before. We can turn the Clippy lint on for this post-launch.

PR #79, opened 2021-05-12
Implement versioned states.

joshuajbouw:
Implements versioned states which were missed in #77.

Thanks @mfornet.


mfornet (commented):
Don't we want to also have versioned state, like
`[version, KeyPrefix::Config, STATE_KEY]`
??

joshuajbouw:
In that method it adds the version already @mfornet

mfornet (approved):
Don't we want to also have versioned state, like
`[version, KeyPrefix::Config, STATE_KEY]`
??

artob (approved):

PR #81, opened 2021-05-12
Explicitly use `evm-core` 0.26.1.

joshuajbouw:
Just in case someone accidentally downgrades the `evm-core` as it is not explicitly directly used, we can specify the version in the cargo manifest to ensure we are using it. This wouldn't be needed if `evm` was updated to 0.26.1 as well.


artob:
Note @sorpaas.

artob (approved):

PR #83, opened 2021-05-12
Remove unused VersionPrefixU8

birchmd:
This code is not needed, so let's not have it.


joshuajbouw:
Yeah, remove it. It was part of the Lookup map at one point, but I recommend expanding this PR or doing another PR that removes the `KeyPrefixU8` as it is only used in that.

sept-en:
@birchmd great to see redundant & unused pieces are found and removed!

joshuajbouw (approved):
Looks good

artob (approved):

PR #84, opened 2021-05-13
Implement generational storage with `SELFDESTRUCT` tests

mfornet:
Check state is removed after selfdestruct is called following this description: 

`https://gist.github.com/mfornet/a8ffcedfffd22612e34cf0b3c3f5c9e3`

@joshuajbouw use this test, to check that reset state is working.

----

Also add a test (`test_self_destruct_with_submit`) where function that call `selfdestruct` is called using `engine.submit` rather than `engine.call` and it fails with:
```
`Result::unwrap()` on an `Err` value: InconsistentStateError(IntegerOverflow)'
```
This was found by chance. @birchmd @artob can you take a look at this.


joshuajbouw:
Hm, getting an error that I'm missing a file? Can you double-check that it's pushed?

```rust
    pub fn load() -> Self {
        Self(solidity::ContractConstructor::compile_from_extended_json(
            "etc/eth-contracts/artifacts/contracts/StateTest.sol/SelfDestructFactory.json",
        ))
    }
```

birchmd:
> Hm, getting an error that I'm missing a file?

Looks like it's supposed to be generated by `yarn` (see changes to `Makefile`). We'll just need to update the GitHub action to run that first.

birchmd:
By the way @mfornet is `yarn` (or hardhat or whatever) needed to compile this contract for some reason? If this existing `solidity::compile` function would work (as is used for the standard precompiles test) then I would prefer that to adding an additional tooling dependency.

Even if `yarn` is needed we could probably use a similar tactic to what I used with the standard precompiles contract to get rust to call `yarn` itself rather than have it go through the `Makefile`. This would be a better way than what I suggested about about changing the GitHub action.

src/test_utils/mod.rs
```diff
@@ -26,9 +26,36 @@ near_sdk_sim::lazy_static_include::lazy_static_include_bytes! {
 pub(crate) const SUBMIT: &str = "submit";
 
 pub(crate) mod erc20;
+pub(crate) mod self_destruct;
 pub(crate) mod solidity;
 pub(crate) mod standard_precompiles;
 
+pub(crate) struct Signer {
```
birchmd, line 33: Nice! I like this a lot. Might be worth refactoring other tests to use this in the future.

birchmd:
> called using `engine.submit` rather than `engine.call` and it fails

Noted. I'll look into it.

joshuajbouw:
Got it now @birchmd

birchmd:
@mfornet the error you encountered with `submit` is fixed in #85

artob:
@joshuajbouw @mfornet Please resolve the conflicts now after the merge of #87.

src/precompiles/mod.rs
```diff
@@ -150,17 +164,99 @@ pub fn berlin_precompiles(
...
+        y_bytes[8],
+        y_bytes[9],
+        y_bytes[10],
+        y_bytes[11],
+        y_bytes[12],
+        y_bytes[13],
+        y_bytes[14],
+        y_bytes[15],
+    ]
+}
```
joshuajbouw, line 213: Better version of this would be to use `copy_from_slice`

joshuajbouw:
Just merged `develop` into this branch carefully as there were at least 20 file conflicts.

joshuajbouw (approved):
Generally fine, I still have to wait for the CI to go through its run, but my only complaint is that there is obviously quite a bit of feature creep going on here. Do be very careful with that going forward. Would be best to have each of those as separate PRs. Also, do thoroughly document all changes that had been made in the description.

src/precompiles/mod.rs
mfornet: Is it possible to use `copy_from_slice` in `const fn`? I think that's why it was done this way.

birchmd: Indeed `copy_from_slice` is not allowed (because `&mut` pointers are not allowed). See [here](https://github.com/aurora-is-near/aurora-engine/pull/107#discussion_r640885291) for further discussion on an implementation which has less boilerplate, but it worse in my opinion.

mfornet:
Found a subtle bug in the [way keys were created](https://github.com/aurora-is-near/aurora-engine/pull/84/commits/27314b19303d10934a1cb00b8d9812ea6344ceec#diff-18c17f16d6477d10ed90360fe38e0e13fc117ecd9548f8f9374236c3ec7f8750L106-R107). Some tests are still failing (see CI), and is a regression introduced in PR #87, already checked on previous commit that it was working fine. 

I was looking into it, but still haven't figured out the issue.

[Simple test where it fails](https://github.com/aurora-is-near/aurora-engine/blob/27314b19303d10934a1cb00b8d9812ea6344ceec/src/tests/erc20.rs#L15)

1. A standard ERC20 implementation (form OpenZeppelin) is deployed.
2. Check balance from a user (it is 0)
3. Mint 10 tokens for the user
4. Check balance from a user (it is 10)

At step 3 the transaction fails.

src/engine.rs
```diff
@@ -558,23 +579,27 @@ impl ApplyBackend for Engine {
                         Engine::set_code(&address, &code)
                     }
 
-                    if reset_storage {
-                        Engine::remove_all_storage(&address)
-                    }
+                    // TODO: When does the storage needs to be wiped?
+                    // if reset_storage {
+                    // Engine::remove_all_storage(&address, generation)
+                    // }
```
mfornet, line 585: We need to understand what is the meaning of the flag `reset_storage`. From [this comment](https://github.com/aurora-is-near/sputnikvm/blob/master/src/backend/mod.rs#L38-L40) it seems that the storage needs to be wiped, however, as seen in some tests, this is not working as expected.

Maybe the comment in sputnikvm refers to a different storage 🤔

birchmd: I see the error. `remove_all_storage` increments the generation by 1, but then in the storage iterator we are using the previous generation. We should do the following:

```
if reset_storage {
    Engine::remove_all_storage(&address, generation);
    generation += 1;
}
```

birchmd: I confirmed all tests pass with that change, but actually, that solution is not so good because it interacts poorly with the `if delete_empty` block later. Need to think about it more carefully.

birchmd: Fixed in https://github.com/aurora-is-near/aurora-engine/pull/84/commits/dc49d44e224e5038bcf1a44bbd00c11fdbf80095

birchmd (approved):
Great work team!

PR #85, opened 2021-05-13
Fix(tests): Track storage usage in AuroraRunner

birchmd:
In #84 Marcelo encountered an error due to underflow resulting from releasing storage that was not properly tracked. This PR fixes that issue.

Note: the issue is only in the testing framework, it does not impact any production code.


mfornet (approved):

mfornet:
Did you check that test from #84 was working? in that case I'll rebase on top of `develop` once this is merged! :)

mfornet:
Changed target branch to `test_self_destruct` so we can merge them together with CI green. If you are ok with this approach, please merge.

joshuajbouw (approved):
Good here.

Issue #76, opened 2021-05-12
Remedy action items from #74


artob:
Please address the various code-quality `NOTE`  and `TODO` comments and suggests in @evgenykuzyakov's code review in #74.

birchmd:
Fixed by #86 and #103

PR #86, opened 2021-05-13
Address comments from Audit

birchmd:
Closes https://github.com/aurora-is-near/aurora-engine/issues/76

Addresses most of the minor TODOs from #74 
One still remains, but I'm unsure how to address it. I've [reached out for clarification](https://github.com/aurora-is-near/aurora-engine/pull/74#discussion_r632025663). That last one can be done as a separate PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -116,7 +116,8 @@ mod contract {
     pub extern "C" fn get_upgrade_index() {
         let state = Engine::get_state();
         let index = sdk::read_u64(&bytes_to_key(KeyPrefix::Config, CODE_STAGE_KEY))
-            .sdk_expect("ERR_NO_UPGRADE");
+            .sdk_expect("ERR_NO_UPGRADE")
+            .sdk_unwrap();
         sdk::return_output(&(index + state.upgrade_delay_blocks).to_le_bytes())
     }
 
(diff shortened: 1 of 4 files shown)


sept-en (approved):
LGTM!

Also, as you use constant for the `REGISTER_ID` in `read_input_arr_20()`, please do the same in `read_input()`.

src/lib.rs
```diff
@@ -161,9 +164,7 @@ mod contract {
     /// Call method on the EVM contract.
     #[no_mangle]
     pub extern "C" fn call() {
-        // TODO: Borsh input pattern is so common here. It worth writing sdk::read_input_borsh().
-        let input = sdk::read_input();
-        let args = FunctionCallArgs::try_from_slice(&input).sdk_expect("ERR_ARG_PARSE");
+        let args: FunctionCallArgs = sdk::read_input_borsh().sdk_unwrap();
```
joshuajbouw, line 167: Just a general note, I think I was implying the `sdk_unwrap`s to be not really used, unless we are using it for testing or some other purposes. For production, they should always return something clear that we can follow up on.

birchmd: `read_input_borsh` returns a `Result`, so `sdk_unwrap` will use the `AsRef[u8]` implementation on the error type, which returns `ERR_ARG_PARSE` as it did previously. We can change the error message to mention borsh if you think it should be more clear.

birchmd:
> Also, as you use constant for the REGISTER_ID in read_input_arr_20(), please do the same in read_input().

@sept-en done in https://github.com/aurora-is-near/aurora-engine/pull/86/commits/8ae64439b4481839e63f3156fd69b2a6ff75c235

artob (approved):

PR #87, opened 2021-05-14
Implement generational storage

joshuajbouw:
This implements the backwards compatible generational storage, which passes the tests.

How it works is that we have generational addresses, which start at 0, which increment by 1 every time a self destruct action, or more specifically, every time we `Apply::Delete`.

This feature is backwards compatible with the current storage structure and shouldn't cause conflicting issues as our keys are either a length of 54 or 58. Prior, our account storage kept a length of 54; however, to give us access to new storage, we must provide a new key. That is where the key length of 58 comes into play as the extra 4 bytes will allow us to tack on the generation value `u32`.

NOTE that the tests are -expectedly- failing. They were addressed in https://github.com/aurora-is-near/aurora-engine/pull/85.


src/engine.rs
```diff
@@ -222,16 +222,16 @@ impl Engine {
             .unwrap_or_else(U256::zero)
     }
 
-    pub fn remove_storage(address: &Address, key: &H256) {
-        sdk::remove_storage(&storage_to_key(address, key));
+    pub fn remove_storage(address: &Address, key: &H256, generation: u32) {
```
mfornet, line 225: What do you think about creating a separate struct:

```rust
AddressWithGeneration {
    address: Address,
    generation: u64,
}
```

You can use a different (and better) name for it. 

This way interface for all this method will be a little bit cleaner.

src/engine.rs
```diff
@@ -503,7 +522,8 @@ impl evm::backend::Backend for Engine {
 
     /// Get storage value of address at index.
     fn storage(&self, address: Address, index: H256) -> H256 {
```
mfornet, line 524: Not from this PR, but still

It is weird to use `index` for this variable name here and `key` elsewhere (see `get_storage`/`set_storage`), when they are referring to the same thing (have same semantics). I'd keep `key` which I believe is what ethereum is calling this value.

src/storage.rs
```diff
@@ -52,8 +54,31 @@ pub fn address_to_key(prefix: KeyPrefix, address: &Address) -> [u8; 22] {
...
+    if generation == 0 {
+        StorageKeyKind::Normal(normal_storage_key(address, key))
+    } else {
+        StorageKeyKind::Generation(generation_storage_key(address, key, generation))
+    }
+}
+
 #[allow(dead_code)]
-pub fn storage_to_key(address: &Address, key: &H256) -> [u8; 54] {
+pub fn normal_storage_key(address: &Address, key: &H256) -> [u8; 54] {
```
mfornet, line 81: Remove `pub` we should only be using `storage_to_key`

src/storage.rs
```diff
@@ -63,13 +88,13 @@ pub fn storage_to_key(address: &Address, key: &H256) -> [u8; 54] {
 }
 
 #[allow(dead_code)]
-pub fn storage_to_key_nonced(address: &Address, storage_nonce: u32, key: &H256) -> [u8; 58] {
+pub fn generation_storage_key(address: &Address, key: &H256, generation: u32) -> [u8; 58] {
```
mfornet, line 91: Remove `pub` we should only be using `storage_to_key`

src/engine.rs
```diff
@@ -503,7 +522,8 @@ impl evm::backend::Backend for Engine {
...
-        Engine::get_storage(&address, &index)
+        let generation = Self::get_generation(&address);
+        Engine::get_storage(&address, &index, generation)
```
mfornet, line 526: This doesn't feel as efficient as possible. We are making two read access per read in the evm.

What about we have some cache created on the fly on each call. Since the most common scenarios is contracts reading from the same address this should be a noticeable improve. Generation is only read once per call per address.

mfornet: We can discuss how to do this.

mfornet (changes requested):

src/engine.rs
joshuajbouw: Typically I do stuff like that if we get over 5 arguments. Here, I didn't see much of a point but, I get that grouping it together would be fine.

src/storage.rs
joshuajbouw: Good point, as there are all internal.

src/engine.rs
joshuajbouw: Yeah, it was kind of frustrating to really find a good way to do this and I was quite hesitant. I wanted to look into if we can do some early returns on some other parts. But, I'm not sure if that would be possible. The idea is to only read it once the entire time though.

birchmd: I agree that caching would likely be a good idea. It's a little unfortunate that the type signature of `fn storage` is `&self` instead of `&mut self`.

Maybe it's worth merging a version without caching and make an issue to do some optimization work in the future. This way we can get the core fix in faster and also spend more dedicated time on optimization. I think optimization should usually be a dedicated task so that measurements can be made as part of the process. E.g. we will know how much gas is saved by adding a caching layer, or whatever other clever solution we might come up with given more time to design something.

mfornet: Agree with Michael, let's delay this optimisation for future PR

artob: In principle, we could defer optimizations. However, given that we are already at (or over) any reasonable gas use with the current contract, the impact of everything like this ought to at least be quantified. Adding a few gas-guzzling tweaks here and there may impact our bottom line soon enough.

src/engine.rs
artob: Yes, `SSTORE` and `SLOAD` [use](https://ethervm.io) the nomenclature `key`.

src/engine.rs
joshuajbouw: I made some changes to this which stores the value in state upon creation.

src/engine.rs
```diff
@@ -501,9 +526,9 @@ impl evm::backend::Backend for Engine {
         Engine::get_code(&address)
     }
 
-    /// Get storage value of address at index.
-    fn storage(&self, address: Address, index: H256) -> H256 {
-        Engine::get_storage(&address, &index)
+    /// Get storage value of address at key.
+    fn storage(&self, address: Address, key: H256) -> H256 {
+        Engine::get_storage(&address, &key, self.generation)
```
birchmd, line 531: I think this is wrong. If a user account calls a contract the generation should be that of the contract, not the user account. As is, it looks like `self.generation` is set by the user account initiating the transaction.

Even if you were to change `self.generation` to match the contract that was being called by the user (if any), then it still wouldn't work because that contract could then call out to a different contract, and then its generation should be used.

What I am getting at is that `generation` is not a constant for the whole transaction, it scoped to a particular call, which means it cannot come from a field of `self` (since SputnikVM manages the call stack, not us). I think the only way to avoid multiple look ups is with caching (or some invasive change to SputnikVM).

birchmd (changes requested):

src/engine.rs
joshuajbouw: So, reverting the store generation on creation, all good?

joshuajbouw: Reverted

birchmd: @joshuajbouw you didn't revert the right commit.

joshuajbouw: Oh, wow. What in the world.

birchmd (approved):
LGTM
I filed https://github.com/aurora-is-near/aurora-engine/issues/105 to follow up on the performance question since it will not be addressed in this PR

src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![feature(array_methods)]
```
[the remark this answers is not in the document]
joshuajbouw: That could have been a Jetbrains Rust automatic addition.

src/engine.rs
```diff
@@ -542,17 +563,20 @@ impl ApplyBackend for Engine {
```
mfornet, line 563: This won't work since `remove_all_storage` is empty right now.

Out of curiosity, when is the flat `reset_storage` set to true?

src/engine.rs
```diff
@@ -256,17 +274,18 @@ impl Engine {
     }
 
     /// Removes an account.
-    pub fn remove_account(address: &Address) {
+    pub fn remove_account(address: &Address, generation: u32) {
         Self::remove_nonce(address);
         Self::remove_balance(address);
         Self::remove_code(address);
         Self::remove_all_storage(address);
+        Self::set_generation(address, generation + 1);
```
mfornet, line 282: Move this call inside `remove_all_storage`

mfornet (changes requested):

src/engine.rs
joshuajbouw: Its a possible flag that can be passed back by the EVM. Technically, we don't wipe it, but I just have it in there for prosperity sake. Maybe there will eventually be a use for it. Lots of notes and discussion on that already. See engine.rs L#287

src/engine.rs
```diff
@@ -542,17 +563,20 @@ impl ApplyBackend for Engine {
...
                     }
 
                     if delete_empty {
-                        Engine::remove_account_if_empty(&address)
+                        Engine::remove_account_if_empty(&address, generation)
                     }
                 }
-                Apply::Delete { address } => Engine::remove_account(&address),
+                Apply::Delete { address } => {
+                    let generation = Self::get_generation(&address);
```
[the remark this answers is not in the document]
joshuajbouw: We need the address from somewhere though that needs to be modified is the issue with that.

src/engine.rs
joshuajbouw: Good call, done.

joshuajbouw:
This should be good to go one tests pass.

artob:
@joshuajbouw Tests still failing

joshuajbouw:
> @joshuajbouw Tests still failing

As far as I was aware this was expected as it was fixed elsewhere?

birchmd:
@artob @mfornet this PR is targeting the branch with the new test (`test_self_destruct`) which #84 is based on. I suggest we should merge this PR into #84 (instead of the other way around) since it is easier and then we can debug any test failures there.

mfornet (approved):
LGTM, Approving to merge into #84

joshuajbouw:
Can someone merge this please?

mfornet:
Merging into #84 and will fix tests there

PR #88, opened 2021-05-14
EthConnector: make AdminControlled and pausable

sept-en:


src/admin_controlled.rs
```diff
@@ -0,0 +1,26 @@
+use crate::sdk;
+
+pub type PausedMask = u8;
```
birchmd, line 3: Might be worth using https://crates.io/crates/bitflags to make this (and constants `PAUSE_DEPOSIT` `PAUSE_WITHDRAW`) a little easier to read.

src/connector.rs
```diff
@@ -607,4 +629,32 @@ impl EthConnectorContract {
     fn check_used_event(&self, key: &str) -> bool {
         sdk::storage_has_key(&self.used_event_key(key))
     }
+
+    /// Get Eth connector paused flags
+    pub fn get_paused_flags(&self) {
+        let data = self.get_paused().try_to_vec().unwrap();
+        sdk::return_output(&data[..]);
```
birchmd, line 636: This is maybe cheating a bit by using the fact that `PauseMask` is a `u8`, but it seems unnecessary to go through borsh for data this simple.

```suggestion
        let mask = self.get_paused();
        sdk::return_output(&[mask]);
```

birchmd (commented):
This is looking pretty good to me so far!

src/admin_controlled.rs
sept-en: Good suggestion, I like the approach of using `bitflags`!
Though we need to think if we change this right now because the current implementation is almost the same that we have for the `rainbow-bridge` and `rainbow-token-connector`.

Also I think that using this crate can bloat the contract size up. So we need to carefully decide about it.

src/connector.rs
sept-en: I was thinking about it, but this will need to be changed if we change the type of `PauseMask` later. Also, we will need to have another check for the input length then (borsh verifies this, that there's no other input data left).
I'd prefer to use Borsh for consistency if you are ok with it.

sept-en:
Btw, the CI is failing because of the internal Github Actions issue reported [there](https://www.githubstatus.com/incidents/kdvly1jxxdpl).

src/connector.rs
birchmd: Yeah, makes sense to me. You're right that preferring consistency to efficiency is the better approach in the long run.

And if we are ever at the point we care about gas used by borsh vs directly returning the byte then we're probably in great shape anyway haha!

src/admin_controlled.rs
birchmd: Sure, I don't have a strong opinion either way. If the crate isn't providing much value and we've implemented direct bitflags in other places then it's probably fine to leave it as is. I just thought it would mention it in case it was something you didn't know about.

src/connector.rs
```diff
@@ -484,6 +502,8 @@ impl EthConnectorContract {
     /// We starting early checking for message data to avoid `ft_on_transfer` call panics
     /// But we don't check relayer exists. If relayer doesn't exist we simply not mint/burn the amount of the fee
     pub fn ft_transfer_call(&mut self) {
+        //TODO: perhaps need to add pausability functionality here as well?
```
birchmd, line 505: How will we decide whether this should be done or not?

tests/test_connector.rs
```diff
@@ -606,3 +611,259 @@ fn test_ft_transfer_call_fee_greater_than_amount() {
...
+        }
+        .try_to_vec()
+        .unwrap(),
+        DEFAULT_GAS,
+        1,
+    );
+    let promises = res.promise_results();
+    let p = promises[1].clone();
+    match p.unwrap().status() {
+        ExecutionStatus::Failure(_) => {}
```
birchmd, line 841: I wonder if it's worth trying to check specifically the error is raised because of the pause (e.g. by checking the error message), as opposed to some other reason. Just for the sake of completeness.

tests/test_connector.rs
```diff
@@ -606,3 +611,259 @@ fn test_ft_transfer_call_fee_greater_than_amount() {
...
+        _ => panic!(),
```
birchmd, line 842: Maybe worth adding a helpful message here like "Expected failure due to pause, but withdraw succeeded"

tests/test_connector.rs
```diff
@@ -12,6 +12,11 @@ use byte_slice_cast::AsByteSlice;
 use near_sdk_sim::transaction::ExecutionStatus;
 use primitive_types::U256;
 
+pub type PausedMask = u8;
+const UNPAUSE_ALL: PausedMask = 0;
+const PAUSE_DEPOSIT: PausedMask = 1 << 0;
```
birchmd, line 17: I still don't like all this code duplication between production and tests. Nothing to do about it in this PR, but we'll have to refactor this in the future.

sept-en: I agree!

src/connector.rs
sept-en: This is a public NEP-141 method, so we need to think if we want to add pausability for bridged Ether as a separate flag (i.e. `PAUSED_TRANSFER`).
@artob

tests/test_connector.rs
sept-en: Makes sense for me! It's worth adding the same for already existing tests there

src/lib.rs
```diff
@@ -434,6 +436,16 @@ mod contract {
         EthConnectorContract::get_instance().ft_on_transfer(&engine)
     }
 
+    #[no_mangle]
+    pub extern "C" fn get_eth_connector_paused_flags() {
+        EthConnectorContract::get_instance().get_paused_flags()
+    }
+
+    #[no_mangle]
+    pub extern "C" fn set_eth_connector_paused_flags() {
```
mrLSD, line 445: I think just `set_paused_flags`

src/admin_controlled.rs
mrLSD: I'm pretty sure that all logic with `mask` should be refactored with `enum`.
```
pub enum ContractPause {
    Deposit = 0x0,
    Withdraw = 0x1,
}
```

src/connector.rs
```diff
@@ -17,10 +18,15 @@ const GAS_FOR_FINISH_DEPOSIT: Gas = 50_000_000_000_000;
 const GAS_FOR_VERIFY_LOG_ENTRY: Gas = 40_000_000_000_000;
 const GAS_FOR_TRANSFER_CALL: Gas = 40_000_000_000_000;
 
+const UNPAUSE_ALL: PausedMask = 0;
+const PAUSE_DEPOSIT: PausedMask = 1 << 0;
+const PAUSE_WITHDRAW: PausedMask = 1 << 1;
```
mrLSD, line 23: We don't need that constants, because we can operate with `enum` representation as Imanited before.

mrLSD (changes requested):
We should change `PausedMask` logic to `enum`.

mrLSD (approved):
We discussed with @sept-en  directly, so for now current implementation is pretty enough.

src/lib.rs
sept-en: Ok for me if this won't produce ambiguity - which is paused, etc. What do you think?

src/admin_controlled.rs
joshuajbouw: Bitflag is awesome and lightweight. Very very light library. I second that.

src/admin_controlled.rs
```diff
@@ -0,0 +1,26 @@
...
+
+pub trait AdminControlled {
+    fn is_owner(&self) -> bool {
```
joshuajbouw, line 6: Lets try to keep documentation on new methods.

joshuajbouw, line 6:
```suggestion
    /// Returns true if the current account is the owner.
    fn is_owner(&self) -> bool {
```

src/connector.rs
joshuajbouw: Bitflags as mentioned above, should be done here.

src/admin_controlled.rs
```diff
@@ -0,0 +1,26 @@
...
+    fn set_paused(&mut self, paused: PausedMask);
+
+    /// Return if the contract is paused for the current flag and user
+    fn is_paused(&self, flag: PausedMask) -> bool {
+        (self.get_paused() & flag) != 0 && !self.is_owner()
+    }
+
+    fn check_not_paused(&self, flag: PausedMask) {
+        assert!(!self.is_paused(flag));
+    }
```
joshuajbouw, line 25: This method seems pretty pointless to me since it's just an assertion. I would remove it and replace all uses with `assert!(!self.is_paused(flag));` just to be clear what it is doing. I personally don't like to hide assertions like this.

src/connector.rs
joshuajbouw: NEP-141 has very specific requirements on what to add, I would just stick with those. I don't think it is a good idea to add plausibility. May lead to unintended consequences and could actually break compatibility.

src/connector.rs
```diff
@@ -551,6 +571,8 @@ impl EthConnectorContract {
     /// ft_on_transfer callback function
     #[allow(clippy::unnecessary_unwrap)]
     pub fn ft_on_transfer(&mut self, engine: &Engine) {
+        //TODO: perhaps need to add pausability functionality here as well?
+
```
joshuajbouw, line 575: Same comment as above re: shouldn't do this.

src/connector.rs
```diff
@@ -607,4 +629,33 @@ impl EthConnectorContract {
     fn check_used_event(&self, key: &str) -> bool {
         sdk::storage_has_key(&self.used_event_key(key))
     }
+
+    /// Get Eth connector paused flags
+    pub fn get_paused_flags(&self) {
+        let data = self.get_paused().try_to_vec().unwrap();
```
joshuajbouw, line 635: This should be an expect with `ERR_FAILED_PARSE`
```suggestion
        let data = self.get_paused().try_to_vec().expect(ERR_FAILED_PARSE);
```

src/lib.rs
joshuajbouw: I second `set_paused_flags` and `get_paused_flags` since nothing else will be needing it.

tests/test_connector.rs
joshuajbouw: Just adding to technical debt eh :D?

joshuajbouw (changes requested):

src/admin_controlled.rs
sept-en: I don't think we should explicitly use it like this, what if we change some rules there?
But I agree with the idea it's not obvious that `check_not_paused()` may assert. I'm thinking about renaming the function instead to `assert_not_paused()`. WDYT?

sept-en:
As we internally discussed with @joshuajbouw @birchmd, we won't use `bitflags` crate at the current implementation (at least now). The initial idea was to improve the readability and improve the development experience, but we discovered that using this approach will require having additional boilerplate code for Borsh serialization/deserialization, make the `AdminControlled` trait very specific and not reusable in other possible components/repos. We also agreed that each dependency increases the risk from a security perspective, and it's better not to add additional dependencies if it's not crucial or doesn't give us significant improvements.

src/connector.rs
```diff
@@ -612,4 +630,33 @@ impl EthConnectorContract {
     fn check_used_event(&self, key: &str) -> bool {
         sdk::storage_has_key(&self.used_event_key(key))
     }
+
+    /// Get Eth connector paused flags
+    pub fn get_paused_flags(&self) {
+        let data = self.get_paused().try_to_vec().expect(ERR_FAILED_PARSE);
```
birchmd, line 636: Seems like an abuse of `ERR_FAILED_PARSE` since we normally use it on input args, not output data I think. But it doesn't really matter because this will never fail anyways.

tests/test_connector.rs
```diff
@@ -586,6 +591,262 @@ fn test_ft_transfer_call_fee_greater_than_amount() {
...
+fn test_admin_controlled_only_admin_can_pause() {
+    let (master_account, contract) = init(CUSTODIAN_ADDRESS);
+    let user_account = create_user_account(&master_account);
+
+    // Try to pause from the user - should fail
+    let res = call_set_paused_flags(&user_account, CONTRACT_ACC, PAUSE_DEPOSIT);
+    let promises = res.promise_results();
+    let p = promises[1].clone();
+    match p.unwrap().status() {
+        ExecutionStatus::Failure(_) => {}
```
birchmd, line 643: Discussed privately that matching on the error message is currently not possible because of some linker error.

It is ok to merge as is, but please file and issue and add a TODO comment with the issue number so we remember to come back to it in the future.

src/lib.rs
```diff
@@ -447,6 +449,16 @@ mod contract {
         EthConnectorContract::get_instance().ft_on_transfer(&engine)
     }
 
+    #[no_mangle]
+    pub extern "C" fn get_paused_flags() {
+        EthConnectorContract::get_instance().get_paused_flags()
```
birchmd, line 454: As discussed elsewhere, I don't like this pattern of doing IO in business logic (outside `lib.rs`). It's fine to merge as is, but I'd like to avoid this in the future and go back to change these when we get a chance.

birchmd (approved):
LGTM nice work!

tests/test_connector.rs
sept-en: As @mrLSD introduced another way to check for the error message in #106 ([code](https://github.com/aurora-is-near/aurora-engine/blob/dc245406a6973b22baddd6ba2cae949a168c7b88/tests/test_connector.rs#L868)), the same way will be used when that PR is merged.

src/lib.rs
sept-en: I agree! I made the changes for this.

sept-en: I agree! I made changes for this.

tests/test_connector.rs
sept-en: Added explicit error message check.

PR #92, opened 2021-05-17
Use cache action to save cargo artifacts between runs

### LINKED ISSUE #91: Improve CI performance via caching
CI takes a long time to run presently. The majority of this time is spent downloading and compiling rust crates. If we cache cargo artifacts then we should be able to greatly improve the CI performance. The purpose of this issue is to investigate using caching in GitHub actions.

birchmd:
Closes #91 

Hopefully this works. It'll be a little difficult to test because it will only help future PRs (not this one since this one will create the cache I think). Currently the tests take ~20 minutes to run ([example](https://github.com/aurora-is-near/aurora-engine/runs/2578076872)), so we'll have to see if this time goes down.

Workflow based on the [official `cache` action documentation](https://github.com/actions/cache/blob/main/examples.md#rust---cargo).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,6 +8,14 @@ jobs:
     steps:
       - name: Clone the repository
         uses: actions/checkout@v2
+      - name: Cache Cargo artifacts
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - name: Install the toolchain
         uses: actions-rs/toolchain@v1
         with:


.github/workflows/tests.yml
```diff
@@ -8,6 +8,14 @@ jobs:
     steps:
       - name: Clone the repository
         uses: actions/checkout@v2
+      - name: Cache Cargo artifacts
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
```
artob, line 16: I wouldn't be surprised if tilde expansion wouldn't be performed on these paths.

artob (approved):
Let's give it a go...

matklad:
If ci caching is a concern, this can be improved a lot, that guideline leaves a lot of perf on the table (b/c cargo doesn't make caching easy).

Easy wins: 

* Disable incremental compilation on CI to reduce the size of the cache a lot: https://github.com/rust-analyzer/rust-analyzer/blob/e3d0d89d7e3e253234271008df9324ec1faf1746/.github/workflows/ci.yaml#L11 (that's default on 1.52.1 https://blog.rust-lang.org/2021/05/10/Rust-1.52.1.html,  :sad rofl: )
* Disable debug info, it is not useful on CI, and increases the size of target by a lot. Probably best to do this via environmental variable `CARGO_PROFILE_RELEASE_DEBUG=false`, `CARGO_PROFILE_DEV_DEBUG=false`

Hard win:

You want to cache compiled artifacts for dependencies, but not for your own code. You can do it yourself or use an action for this. See https://github.com/rust-analyzer/rust-analyzer/pull/8427, which changes from "do it yourself" to "use action".

PR #93, opened 2021-05-18
Ensure that Cargo.lock in the repo is valid

matklad:
At the moment, running `cargo check` locally for me:

* fails with some compilation error in the dependency
* updates Cargo.lock

That is, the Cargo.lock that is committed into the repo doesn't work. To
prevent this from happening in the future, let's forbid CI from
upgrading Cargo.lock.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,7 +28,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --verbose
+          args: --locked --verbose
 env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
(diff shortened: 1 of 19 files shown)


matklad:
This is more of an issue, rather than a pull-request -- to merge this, a working Cargo.lock needs to be supplied, and I won't be looking into fixing that.

birchmd (approved):
`cargo check` works for me. I'm not sure how `Cargo.lock` could have been corrupted since it is always autogenerated. That said, I don't see the harm in adding `--locked`.

joshuajbouw (approved):
Looks fine to me.

artob:
@matklad @joshuajbouw Please retarget this at the `develop` branch, as per the README:
https://github.com/aurora-is-near/aurora-engine#branches

PR #94, opened 2021-05-18
Remove unneeded nightly feature

matklad:
On wasm, intrinsic for trap unreachable is stable and exposed via
`core::arch`, no need for a nightly feature.

Wasn't able to verify if this compiles due to #93.


src/lib.rs
```diff
@@ -56,14 +55,14 @@ mod contract {
...
     #[no_mangle]
     pub unsafe fn on_panic(_info: &::core::panic::PanicInfo) -> ! {
-        ::core::intrinsics::abort();
+        ::core::arch::wasm32::unreachable();
     }
 
     #[cfg(target_arch = "wasm32")]
     #[alloc_error_handler]
     #[no_mangle]
     pub unsafe fn on_alloc_error(_: core::alloc::Layout) -> ! {
```
matklad, line 64: Haven't look deeply into this, but on the first glance seems like this isn't needed, and that we can actually build on stable.

birchmd: I get a compilation error if I remove `on_alloc_error`:

```
error: `#[alloc_error_handler]` function required, but not found.

note: Use `#![feature(default_alloc_error_handler)]` for a default error handler.
```

It suggests we could use a default one, but I don't know if that is on stable or not, or what the default implementation does.

birchmd (approved):

src/lib.rs
matklad: I guess this might be needed -- I missed the fact that we are no_std, and I think it's necessary for no_std + alloc.

joshuajbouw (approved):
Looks good.

artob:
@matklad @joshuajbouw Please retarget this at the `develop` branch, as per the README:
https://github.com/aurora-is-near/aurora-engine#branches

PR #95, opened 2021-05-18
Reduce size of cargo cache in CI

birchmd:
Some low-hanging fruit to reduce the size of the CI cache, as suggested in https://github.com/aurora-is-near/aurora-engine/pull/92#issuecomment-843087788

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,8 +3,6 @@ on: [push, pull_request]
 name: Tests
 env:
   CARGO_INCREMENTAL: 0
-  CARGO_PROFILE_RELEASE_DEBUG: false
-  CARGO_PROFILE_DEV_DEBUG: false
 jobs:
   test:
     name: Test suite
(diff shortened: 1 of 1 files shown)


birchmd:
cc @matklad (couldn't tag you as a reviewer for some reason)

matklad:
LGTM! Might be a good idea to verify locally that these env vars do what I think they should do, as cargo won't warn you if there's a typo

birchmd:
Locally adding these environment variables reduces the size of `target` (freshly compiled after `cargo clean`) from 2.5 GB to 2.4 GB. Not a huge saving, but probably worth doing.

matklad:
Aha, that's surprising (I'd expect something more like 3x reduction), but is explainable: 

* we disable `debug` in Cargo.toml already, so no we don't need to specifically do this for CI: https://github.com/aurora-is-near/aurora-engine/blob/master/Cargo.toml#L31
* the current rust compiled just has incremental compilation disabled https://blog.rust-lang.org/2021/05/10/Rust-1.52.1.html

So we can remove profile envs, and leave the incremental one for a bright future when incremental is back again.

artob (approved):

PR #96, opened 2021-05-18
Newtype for amounts in Wei

### LINKED ISSUE #67: Newtype for balances in EVM
It's easy to accidentally mess up amounts in EVM, even for experts like ourselves (e.g. https://github.com/aurora-is-near/aurora-engine/pull/57#discussion_r629353860 ). Also, `U256` represents lots of different (not interchangeable) quantities: account nonces, account balances (in Wei), ERC20 token balances (of various sorts!), etc.

To make sure we get account balances right we should have a newtype to represent amounts in Wei, and ensure all function arguments that represent balances use this type. It should also include convenience functions for unit conversion.

E.g.

```rust
pub struct Wei(U256);

impl Wei {
    const ETH_TO_WEI: U256 = U256::from(1_000_000_000_000_000_000);

    pub fn new(amount: U256) -> Self {
        Self(amount)
    }

    pub fn from_eth(amount: U256) -> Option<Self> {
        amount.checked_mul(Self::ETH_TO_WEI).map(Self)
    }
}
```

birchmd:
Closes #67 

This is a very safe refactoring since it does not change any functionality. But it is important because it will hopefully prevent us from making silly mistakes in the future when it comes to working with balances.

Notably this probably may impact the eth-connector work. Since that work is higher priority it might make sense to merge it first, then rebase this PR. On the other hand, if there is time to merge this first and then rebase the eth-connector work, then maybe that would be valuable as it may catch simple mistakes in that important work (if there are any).

### DIFF
--- a/src/benches/eth_deploy_code.rs
+++ b/src/benches/eth_deploy_code.rs
@@ -2,10 +2,11 @@ use criterion::{BatchSize, BenchmarkId, Criterion, Throughput};
 use secp256k1::SecretKey;
 
 use crate::test_utils::{address_from_secret_key, create_eth_transaction, deploy_evm, SUBMIT};
+use crate::types::Wei;
 
-const INITIAL_BALANCE: u64 = 1000;
+const INITIAL_BALANCE: Wei = Wei::new_u64(1000);
 const INITIAL_NONCE: u64 = 0;
-const TRANSFER_AMOUNT: u64 = 0;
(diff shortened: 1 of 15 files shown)


artob (approved):
LGTM :+1: Nice work!

sept-en (approved):
@birchmd thanks for adding this! This is really useful and helpful!

PR #97, opened 2021-05-18
Eugene's Eth connector review

evgenykuzyakov:
I didn't fully understand the flow. But here are a few findings:
- Use JSON instead of Borsh for FT standard methods. Since the goal is to be compatible with the rest of NEAR ecosystem with fungible tokens, this change has to use JSON as an input and output for methods like `ft_transfer` and `ft_transfer_call`, as well as view methods like `ft_balance_of`.
- I'm not sure if the internal `ft_transfer_call` is correctly addressing the balance and doesn't the rollback.
- It's very confusing to bundle NEAR and ETH into one FungibleToken struct. It probably worth to use 2 of them, but then have unified implementation. There is also a proposal for a multi-fungible token standard, but it's not on paper.
- I've commented in the code for the issues I found. 
- Also sorry for renaming `save_contract` to `save_ft_contract`, it's not necessary

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -14,6 +14,7 @@ use borsh::{BorshDeserialize, BorshSerialize};
 
 pub const NO_DEPOSIT: Balance = 0;
 const GAS_FOR_FINISH_DEPOSIT: Gas = 50_000_000_000_000;
+// Note: Is 40Tgas always enough?
 const GAS_FOR_VERIFY_LOG_ENTRY: Gas = 40_000_000_000_000;
 const GAS_FOR_TRANSFER_CALL: Gas = 40_000_000_000_000;
 
@@ -31,7 +32,8 @@ pub struct EthConnector {
 }
 
 /// Token message data
-#[derive(Debug, BorshSerialize, BorshDeserialize)]
+#[derive(BorshSerialize, BorshDeserialize)]
+#[cfg_attr(not(target_arch = "wasm32"), derive(Debug))]
 pub enum TokenMessageData {
     Near(AccountId),
     Eth { address: AccountId, message: String },
@@ -91,13 +93,15 @@ impl EthConnectorContract {
             contract: contract_data,
             ft,
         }
-        .save_contract();
+        .save_ft_contract();
     }
 
     /// Parse event message data for tokens
     fn parse_event_message(&self, message: &str) -> TokenMessageData {
+        // TODO: Why not use borsh here? Does message guarantee to not have `:` as part of inner fields
         let data: Vec<_> = message.split(':').collect();
         assert!(data.len() < 3);
+        // TODO: Do you need to verify validity of account ID in the data[0]?
         if data.len() == 1 {
             TokenMessageData::Near(data[0].into())
         } else {
@@ -115,9 +119,11 @@ impl EthConnectorContract {
 
         let msg = hex::decode(data[1]).expect(ERR_FAILED_PARSE);
         let mut fee: [u8; 32] = Default::default();
+        // TODO: Check msg length?
         fee.copy_from_slice(&msg[..32]);
         let mut recipient: EthAddress = Default::default();
         recipient.copy_from_slice(&msg[32..52]);
+        // TODO: Do you need to verify validity of account ID in the data[0]?
         OnTransferMessageData {
             relayer: data[0].into(),
             recipient,
@@ -133,7 +139,7 @@ impl EthConnectorContract {
         let relayer_account_id = String::from_utf8(sdk::predecessor_account_id()).unwrap();
         let mut data = fee.as_byte_slice().to_vec();
         let message = hex::decode(message).expect(ERR_FAILED_PARSE);
-        data.append(&mut message.to_vec());
+        data.extend(message);
         [relayer_account_id, hex::encode(data)].join(":")
     }
 
@@ -275,7 +281,7 @@ impl EthConnectorContract {
         };
         #[cfg(feature = "log")]
         sdk::log("Check verification_success");
-        let verification_success: bool = bool::try_from_slice(&data0).unwrap();
+        let verification_success = bool::try_from_slice(&data0).unwrap();
         assert!(verification_success, "ERR_VERIFY_PROOF");
 
         // Mint tokens to recipient minus fee
@@ -285,8 +291,12 @@ impl EthConnectorContract {
             // Store proof only after `mint` calculations
             self.record_proof(&data.proof_key);
             // Save new contract data
-            self.save_contract();
+            self.save_ft_contract();
 
+            // TODO: Why return a promise here instead of just call it here directly?
+            // Also passing 40Tgas out of total 50Tgas is a bit risky and close to the limit, because the
+            // a new promise cost is 4.5Tgas, but there are some extra data read/writes in this
+            // method, so it may not fit into the remaining 5.5Tgas
             let promise0 = sdk::promise_create(
                 &sdk::current_account_id(),
                 b"ft_transfer_call",
@@ -302,7 +312,7 @@ impl EthConnectorContract {
             // Store proof only after `mint` calculations
             self.record_proof(&data.proof_key);
             // Save new contract data
-            self.save_contract();
+            self.save_ft_contract();
         }
     }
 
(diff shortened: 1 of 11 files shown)


mrLSD:
* It has been Json. But we decided to remove JSON, and put Borsh.
That's compatible with Aurora CLI. And reduced contract size about ~100kb.
* Can you provide some cases for "rollback ft_transfer_call"?
 * In the case of a multi-token, I believe that this is not entirely correct. There is one token, but there are two types of account.

src/fungible_token.rs
```diff
@@ -57,6 +57,12 @@ impl FungibleToken {
                 .total_supply_near
                 .checked_add(amount)
                 .expect("ERR_TOTAL_SUPPLY_OVERFLOW");
+            // Note: Reusing total supply for a sum of ETH and NEAR is very weird. It doesn't give
+            // mean anything reasonable. I guess the goal is to just report changing numbers for the
+            // total supply when either ETH or NEAR balances are changing. In this case it's probably
+            // okay to report only one of them, e.g. ETH total_supply and not report NEAR total
+            // supply through `ft_total_supply`.
+            // TODO: Remove total_supply for NEAR
```
birchmd, line 65: I was really confused about this as well when I first read the code (see https://github.com/aurora-is-near/aurora-engine/pull/24#discussion_r616178696 ). But I think all the units are the same in the struct (all amounts are relating to $ETH) and the separation is about how much $ETH is owned by Aurora EVM addresses vs NEAR accounts. (@mrLSD please correct me if I'm wrong here).

That said, if @evgenykuzyakov made the same mistake then maybe we should really think about trying to express this more clearly in the code. A simple thing could be making a type alias like `type NEth = u128`. More comments describing the flow at a high level could also help.

src/connector.rs
```diff
@@ -502,6 +527,7 @@ impl EthConnectorContract {
 
         // Additional check overflow before process `ft_on_transfer`
         // But don't check overflow for relayer
+        // Note: It can't overflow because the total supply doesn't change during transfer.
```
mrLSD, line 530: Good point.

src/connector.rs
```diff
@@ -476,7 +497,8 @@ impl EthConnectorContract {
             args.sender_id, args.receiver_id
         ));
         // `ft_resolve_transfer` can change `total_supply` so we should save the contract
-        self.save_contract();
+        self.save_ft_contract();
+        // TODO: int -> str
```
mrLSD, line 501: Do you mean return int as bytes without string converting?

src/connector.rs
```diff
@@ -115,9 +119,11 @@ impl EthConnectorContract {
 
         let msg = hex::decode(data[1]).expect(ERR_FAILED_PARSE);
         let mut fee: [u8; 32] = Default::default();
+        // TODO: Check msg length?
         fee.copy_from_slice(&msg[..32]);
         let mut recipient: EthAddress = Default::default();
         recipient.copy_from_slice(&msg[32..52]);
+        // TODO: Do you need to verify validity of account ID in the data[0]?
```
mrLSD, line 126: It's 20-bytes of Eth address. Do we really need some additional checks?
In the other Aurora places, we check only slice length.

src/connector.rs
```diff
@@ -91,13 +93,15 @@ impl EthConnectorContract {
             contract: contract_data,
             ft,
         }
-        .save_contract();
+        .save_ft_contract();
     }
 
     /// Parse event message data for tokens
     fn parse_event_message(&self, message: &str) -> TokenMessageData {
+        // TODO: Why not use borsh here? Does message guarantee to not have `:` as part of inner fields
```
mrLSD, line 101: Because that's from Ethereum. So it should be a simple string.
":" is the separator for two cases:
1. Near account transfer (NEAR accounts can't contain ":")
2. Eth account transfer

src/connector.rs
```diff
@@ -91,13 +93,15 @@ impl EthConnectorContract {
...
         let data: Vec<_> = message.split(':').collect();
         assert!(data.len() < 3);
+        // TODO: Do you need to verify validity of account ID in the data[0]?
```
mrLSD, line 104: Good point.

src/connector.rs
```diff
@@ -115,9 +119,11 @@ impl EthConnectorContract {
 
         let msg = hex::decode(data[1]).expect(ERR_FAILED_PARSE);
         let mut fee: [u8; 32] = Default::default();
+        // TODO: Check msg length?
```
mrLSD, line 122: Good question. If the length is less than 52 bytes it will panics anymore. But if greater it will ignore the tail. For now, we don't have a concrete decision for that.

src/fungible_token.rs
mrLSD: Yes, we can simplify that. But do we really need that? It'll take 10 mins to remove that.

src/connector.rs
sept-en: We discussed that for the future we might extend the protocol for having several colons for the data separation. So as for me, as we parse&check these parts of the string, we should be fine. What issue do you see here?

sept-en:
>     * It has been Json. But we decided to remove JSON, and put Borsh.
>       That's compatible with Aurora CLI. And reduced contract size about ~100kb.
@mrLSD as we discussed recently (on the last week if I'm not wrong), we should have JSON for FT methods. Though it's not necessary to incorporate `serde` back for that. Some manual parsing should work.

src/connector.rs
```diff
@@ -285,8 +291,12 @@ impl EthConnectorContract {
             // Store proof only after `mint` calculations
             self.record_proof(&data.proof_key);
             // Save new contract data
-            self.save_contract();
+            self.save_ft_contract();
 
+            // TODO: Why return a promise here instead of just call it here directly?
```
sept-en, line 296: If I remember correctly this was made to follow the architecture by creating a transfer to Aurora via `ft_transfer_call`. If we don't create a promise and just make a regular function call, we won't be able to verify that this was called from the Aurora contract.

src/connector.rs
```diff
@@ -388,13 +398,15 @@ impl EthConnectorContract {
...
+        self.save_ft_contract();
         sdk::return_output(&res[..]);
     }
 
     /// Return total supply of NEAR + ETH
     pub fn ft_total_supply(&self) {
         let total_supply = self.ft.ft_total_supply();
+        // TODO: To make it match the standard, it has to be a string instead of an integer.
+        //    So we need to wrap the `100` into a string like `"100". This is due to JSON int limits.
         sdk::return_output(&total_supply.to_string().as_bytes());
```
sept-en, line 410: I thought to match the standard it has to return a string. If we use U256 integers internally and return the string (UTF-8 bytes), is there any problem with it?

src/precompiles/native.rs
```diff
@@ -122,12 +122,14 @@ impl Precompile for ExitToNear {
             let amount = U256::from_big_endian(&input_mut[..32]).as_u128();
             input_mut = &input_mut[32..];
 
+            // TODO: You have to charge caller's account balance for this transfer.
+
```
sept-en, line 126: What do you exactly mean? The expected workflow for this is that the user transfers his ERC20 tokens to the ERC20 contract by using its `burn()` function, which will burn the provided tokens and create a call to this precompile, which will create an appropriate promise call from the Aurora account. It's expected that the sender won't have the Near account at all, only the ETH account connected to the Aurora Web3 RPC.

src/precompiles/native.rs
```diff
@@ -212,6 +214,8 @@ impl Precompile for ExitToEthereum {
             let amount = U256::from_big_endian(&input_mut[..32]).as_u128();
             input_mut = &input_mut[32..];
 
+            // TODO: Charge the caller's account balance?
+
```
sept-en, line 218: I think this is similar to the comment I addressed above.

sept-en:
@evgenykuzyakov thanks for the detailed review! This is very helpful for us. Please, check our responses to some of your comments, some things require discussion, so please follow the thread until this is merged!

mrLSD:
> > ```
> > * It has been Json. But we decided to remove JSON, and put Borsh.
> >   That's compatible with Aurora CLI. And reduced contract size about ~100kb.
> > ```
> 
> @mrLSD as we discussed recently (on the last week if I'm not wrong), we should have JSON for FT methods. Though it's not necessary to incorporate `serde` back for that. Some manual parsing should work.

It was long story about that, because JSON was implemented and after that removed because Aurora itself didn't use it.

mfornet:
> It was long story about that, because JSON was implemented and after that removed because Aurora itself didn't use it.

[NEP-141 is using JSON](https://nomicon.io/Standards/FungibleToken/Core.html#reference-level-explanation). If we don't comply with that, it will not be possible to compose this contract with others that expect json.

PR #99, opened 2021-05-19
Refactor Engine::get_state

birchmd:
As pointed out by @joshuajbouw `Engine::get_state` returning a default value when the state is missing could obscure some critical issue if a state migration goes wrong. In this PR we refactor the function to return a `Result` with clear errors.

In future PRs we will address the broader issue of state migration, but this refactor seemed large enough to be done separately, before new code is added.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -92,6 +92,20 @@ impl ExitIntoResult for ExitReason {
     }
 }
 
+pub enum EngineStateError {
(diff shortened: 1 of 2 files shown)


sept-en (approved):
LGTM!

joshuajbouw (approved):

artob (approved):

PR #101, opened 2021-05-20
Ensure that Cargo.lock in the repo is valid

matklad:
At the moment, running `cargo check` locally for me:

* fails with some compilation error in the dependency
* updates Cargo.lock

That is, the Cargo.lock that is committed into the repo doesn't work. To
prevent this from happening in the future, let's forbid CI from
upgrading Cargo.lock.

Previously: https://github.com/aurora-is-near/aurora-engine/pull/93

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,7 +28,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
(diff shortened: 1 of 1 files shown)


artob (approved):
LGTM :+1: Thanks for this!

mrLSD (approved):

PR #103, opened 2021-05-20
Support state migration on upgrade

birchmd:
This PR is much smaller than it looks based on the diff line count. The majority of the changes come from adding a proper test for this functionality (the original `test_upgrade` was not sufficient).

The key differences in the main code are:
 1. The addition of `state_migration` as a new public method of the contract
 2. Changing `sdk::self_deploy` to call `state_migration` after it has deployed the new code.

To test this functionality I define a test contract and upgrade to it, checking that the upgrade succeed by calling a new function it exposes. To prevent duplicating logic, the test contract imports the production contract as a library, which required some small code reorganization in `lib.rs` and changing the crate features a little to expose the `sdk` and `engine` logic, but not the public entry points (if these are imported also then there is a linker error about `state_migration` begin defined twice because it is also defined in the test contract).


src/sdk.rs
```diff
@@ -4,6 +4,7 @@ use borsh::{BorshDeserialize, BorshSerialize};
 
 const READ_STORAGE_REGISTER_ID: u64 = 0;
 const INPUT_REGISTER_ID: u64 = 0;
+const GAS_FOR_STATE_MIGRATION: u64 = 100_000_000_000_000;
```
birchmd, line 7: I'm not 100% sure on the value for this constant. 100 Tgas seems like it should be enough any sort of state migration we want to do, but we could push it higher if we think this might not be enough.

We could also leave it as is and update the constant only if we actually write a migration that requires more gas (I assume we'll test all upgrades before we do them, so we'll know if we need more gas).

sept-en: I think it makes sense to set it as a max value (300 Tgas). As this is not user's function I don't think there's any problem that we will be required to pay some more gas for the state migration (which hopefully won't happen too frequently). So I don't think we need to put any max value constrain here.

birchmd: Well, we can't set it to 300 Tgas because that the way promises work is they take gas from the original transaction. So if the original `deploy_upgrade` transaction was given 300 Tgas then we could only give the promise it creates an amount of gas equal to 300 minus whatever amount is used in creating the promise.

We could set it higher, e.g. I think 250 Tgas would work, but I don't know if there is much value in figuring out exactly how high it can be when we don't know that 100 isn't enough.

sept-en: Oh, yeah sorry I forgot that we are talking about the inner promise. Then sure 250 Tgas could be an option.

sept-en (approved):
LGTM!

etc/state-migration-test/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "aurora-engine-state-migration-test"
+version = "1.0.0"
+authors = ["NEAR <hello@near.org>"]
```
joshuajbouw, line 4: Wouldn't this be changing to Aurora?

src/lib.rs
```diff
@@ -422,3 +414,21 @@ mod contract {
...
+pub unsafe fn on_panic(_info: &::core::panic::PanicInfo) -> ! {
+    ::core::arch::wasm32::unreachable();
+}
+
+#[cfg(target_arch = "wasm32")]
+#[alloc_error_handler]
+#[no_mangle]
+pub unsafe fn on_alloc_error(_: core::alloc::Layout) -> ! {
+    ::core::arch::wasm32::unreachable();
+}
```
joshuajbouw, line 434: Just a small nit, feels like this should be above the contract module.

joshuajbouw (approved):
Just a small nitpick about the authors, but generally, this is fine. Good work.

etc/state-migration-test/Cargo.toml
birchmd: Probably will be changing, but I'm not sure what it'll change to. cc @artob @djsatok 

The main `Cargo.toml` also still lists `NEAR` as the author (I started this `Cargo.toml` by copying the main one).

src/sdk.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/103/commits/208de2d0297e0658c680feea697bf5ff953bc25b

src/lib.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/103/commits/d14fe517cef05e41b22b85d32c6ff150fbeb7237

etc/state-migration-test/Cargo.toml
artob: Let's change it later, not important now

artob (approved):
LGTM :+1:

PR #104, opened 2021-05-20
Eth connector statistic

mrLSD:
Calculate total unique accounts.

Added method:
* get_accounts_counter

Covered with test

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -584,6 +584,11 @@ impl EthConnectorContract {
         sdk::return_output(0.to_string().as_bytes());
     }
 
+    /// Get accounts counter for statistics
+    pub fn get_accounts_counter(&self) {
+        sdk::return_output(self.ft.get_accounts_counter().to_string().as_bytes());
+    }
+
     /// Save eth-connector contract data
     fn save_ft_contract(&mut self) {
         sdk::save_contract(
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -583,7 +583,7 @@ impl EthConnectorContract {
     }
 
     /// Get accounts counter for statistics.
-    /// It represents total unique accounts.
+    /// It represents total unique accounts (all-time, including accounts which now have zero balance).
     pub fn get_accounts_counter(&self) {
         sdk::return_output(self.ft.get_accounts_counter().to_string().as_bytes());
     }
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -584,7 +584,8 @@ impl EthConnectorContract {
         sdk::return_output(0.to_string().as_bytes());
     }
 
-    /// Get accounts counter for statistics
+    /// Get accounts counter for statistics.
+    /// It represents total unique accounts.
     pub fn get_accounts_counter(&self) {
         sdk::return_output(self.ft.get_accounts_counter().to_string().as_bytes());
     }
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -585,7 +585,7 @@ impl EthConnectorContract {
     /// Get accounts counter for statistics.
     /// It represents total unique accounts (all-time, including accounts which now have zero balance).
     pub fn get_accounts_counter(&self) {
-        sdk::return_output(self.ft.get_accounts_counter().to_string().as_bytes());
+        sdk::return_output(&self.ft.get_accounts_counter().to_le_bytes());
     }
 
     /// Save eth-connector contract data
--- a/src/fungible_token.rs
+++ b/src/fungible_token.rs
@@ -466,7 +466,7 @@ impl FungibleToken {
     fn get_statistic_key() -> Vec<u8> {
         storage::bytes_to_key(
             storage::KeyPrefix::EthConnector,
-            &[storage::EthConnectorStorageId::Statistic as u8],
+            &[storage::EthConnectorStorageId::StatisticsAuroraAccountsCounter as u8],
         )
     }
 }
--- a/src/fungible_token.rs
+++ b/src/fungible_token.rs
@@ -436,7 +436,6 @@ impl FungibleToken {
     pub fn accounts_insert(&self, account_id: &str, amount: Balance) {
         if self.accounts_contains_key(account_id) {
             let key = Self::get_statistic_key();
-            sdk::log(&format!("accounts_insert: {}", account_id));
             let accounts_counter = sdk::read_u64(&key)
                 .unwrap_or(0)
                 .checked_add(1)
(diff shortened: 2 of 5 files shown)


src/fungible_token.rs
```diff
@@ -421,7 +421,20 @@ impl FungibleToken {
     }
 
     pub fn accounts_insert(&self, account_id: &str, amount: Balance) {
-        sdk::save_contract(&Self::account_to_key(account_id), &amount)
+        if self.accounts_contains_key(account_id) {
```
birchmd, line 424: Shouldn't this be if the account is _not_ known then increment the counter instead of if the account _is_ known then increment the counter?

Also, is it possible to delete an account? If so should the counter be decremented in this case?

src/fungible_token.rs
```diff
@@ -421,7 +421,20 @@ impl FungibleToken {
...
+            let accounts_counter = sdk::read_u64(&key)
+                .unwrap_or(0)
+                .checked_add(1)
+                .expect("ERR_ACCOUNTS_COUNTER_OVERFLOW");
+            sdk::write_storage(&key, &accounts_counter.to_le_bytes());
+        }
+        sdk::save_contract(&Self::account_to_key(account_id), &amount);
+    }
+
+    /// Get accounts counter for statistics
```
birchmd, line 435: I'd appreciate a little more detail in this comment. What does this counter actually measure? Is it the all-time number of distinct accounts? The number of accounts currently holding a non-zero balance? Something else?

I know someone could read the code to figure out what exactly the counter counts, but I think a high level understanding should be available without getting into the weeds of the code.

tests/test_connector.rs
```diff
@@ -585,3 +585,16 @@ fn test_ft_transfer_call_fee_greater_than_amount() {
...
+fn test_get_accounts_counter() {
+    let (master_account, contract) = init(CUSTODIAN_ADDRESS);
+    call_deposit_near(&contract, CONTRACT_ACC);
+
+    let counter = master_account.view(CONTRACT_ACC.into(), "get_accounts_counter", &[]);
+    let counter: u128 = String::from_utf8(counter.unwrap())
+        .unwrap()
+        .parse()
+        .unwrap();
+    assert_eq!(counter, 2);
```
birchmd, line 599: Why is the value 2? I only see one call to deposit so I would only expect there to be one account.

mrLSD: Please look to that test:
```
#[test]
fn test_near_deposit_balance_total_supply() {
    let (master_account, contract) = init(CUSTODIAN_ADDRESS);
    call_deposit_near(&contract, CONTRACT_ACC);

    let balance = get_near_balance(&master_account, DEPOSITED_RECIPIENT, CONTRACT_ACC);
    assert_eq!(balance, DEPOSITED_AMOUNT - DEPOSITED_FEE);

    let balance = get_near_balance(&master_account, CONTRACT_ACC, CONTRACT_ACC);
    assert_eq!(balance, DEPOSITED_FEE);

```

src/fungible_token.rs
mrLSD: Total unique accounts. I'll extend that comment and PR description.

src/fungible_token.rs
mrLSD: Fixed

src/fungible_token.rs
mrLSD: Done

src/storage.rs
```diff
@@ -25,6 +25,7 @@ pub enum EthConnectorStorageId {
     Contract = 0x0,
     FungibleToken = 0x1,
     UsedEvent = 0x2,
+    Statistic = 0x3,
```
sept-en, line 28: At the moment this is used only to represent the number of unique accounts, but the key used is too general.
Let's create another subkey for the `Statistics` items. I believe we agreed to have only distinct accounts counter so far. But still creating a subkey could be useful for the future and simplify the possible storage migration needed.

If we don't want to add the subkey, let's call it at least explicit:
`    StatisticsNumDistinctAuroraAccounts = 0x3,`

src/fungible_token.rs
```diff
@@ -420,8 +420,24 @@ impl FungibleToken {
         }
     }
 
+    /// Insert account.
+    /// Calculate total unique accounts
     pub fn accounts_insert(&self, account_id: &str, amount: Balance) {
-        sdk::save_contract(&Self::account_to_key(account_id), &amount)
+        if !self.accounts_contains_key(account_id) {
+            let key = Self::get_statistic_key();
```
sept-en, line 427: Let's use either the subkey or an explicit name for the accounts counter (See below for details).

sept-en (changes requested):
LGTM!
Still, let's put some attention to storage keys, otherwise, this could be quite confusing and hard to support later.

src/connector.rs
```diff
@@ -582,6 +582,12 @@ impl EthConnectorContract {
         sdk::return_output(0.to_string().as_bytes());
     }
 
+    /// Get accounts counter for statistics.
+    /// It represents total unique accounts.
```
birchmd, line 586:
```suggestion
    /// It represents total unique accounts (all-time, including accounts which now have zero balance).
```

tests/test_connector.rs
birchmd: Right, there is a fee as well, so one deposit creates two accounts. Thanks for clarifying; it might be worth saying this in a comment as well.

Also, I recommend extending the test to have a transfer between those two accounts after the deposit, and confirm that the count does not change (because the number of unique accounts is the same).

birchmd (approved):
Generally looks good to me. Please see my note about extending the test.

sept-en (approved):

src/connector.rs
```diff
@@ -582,6 +582,12 @@ impl EthConnectorContract {
         sdk::return_output(0.to_string().as_bytes());
     }
 
+    /// Get accounts counter for statistics.
+    /// It represents total unique accounts (all-time, including accounts which now have zero balance).
+    pub fn get_accounts_counter(&self) {
+        sdk::return_output(self.ft.get_accounts_counter().to_string().as_bytes());
```
joshuajbouw, line 588: Wouldn't this be more efficient to just have to `get_accounts_counter().to_le()`? Of course, the other end would need to read it as `le`. But, having them read it as a string feels odd to me.

mrLSD: Done

joshuajbouw (approved):

PR #106, opened 2021-05-26
Eth-connector extended proof tests

mrLSD:
Extended tests of proof data for `Deposit`.


tests/test_connector.rs
```diff
@@ -646,3 +658,218 @@ fn test_get_accounts_counter_and_transfer() {
...
+}
+
+fn check_execution_status_failure(es: ExecutionStatus, err_msg: &str) {
+    match es {
+        ExecutionStatus::Failure(err) => {
+            assert!(err.to_string().contains(err_msg));
+        }
+        _ => panic!(),
+    }
+}
```
sept-en, line 875: This is a pretty funny way to check the error message, though should work :) I'm ok with introducing this as the other way needs to include some errors from crates and have many nested error types checks (and also has some linking issues at the moment).

tests/test_connector.rs
```diff
@@ -457,12 +458,23 @@ fn test_deposit_with_same_proof() {
...
-    let l = promises.len();
-    let p = promises[l - 2].clone();
-    match p.unwrap().status() {
-        ExecutionStatus::Failure(_) => {}
-        _ => panic!(),
-    }
+    let promise = &promises[promises.len() - 2];
+    check_execution_status_failure(
+        promise.as_ref().unwrap().outcome().clone().status,
+        "ERR_PROOF_EXIST",
```
sept-en, line 464: As we have a constant defined for `ERR_NOT_ENOUGH_BALANCE_FOR_FEE`, please add another constant for this error as well.

tests/test_connector.rs
```diff
@@ -457,12 +458,23 @@ fn test_deposit_with_same_proof() {
...
+
+#[test]
+fn test_deposit_wrong_custodian_address() {
+    let custodian_address = "096DE9C2B8A5B8c22cEe3289B101f6960d68E510";
+    let (_master_account, contract) = init(custodian_address);
+    let promises = call_deposit_near(&contract, CONTRACT_ACC);
+    let promise = &promises[promises.len() - 2];
+    check_execution_status_failure(
+        promise.as_ref().unwrap().outcome().clone().status,
+        "ERR_WRONG_EVENT_ADDRESS",
```
sept-en, line 476: As we have a constant defined for `ERR_NOT_ENOUGH_BALANCE_FOR_FEE`, please add another constant for this error as well.

tests/test_connector.rs
```diff
@@ -457,12 +458,23 @@ fn test_deposit_with_same_proof() {
...
+    let promise = &promises[promises.len() - 2];
+    check_execution_status_failure(
+        promise.as_ref().unwrap().outcome().clone().status,
+        "ERR_PROOF_EXIST",
+    );
+}
+
+#[test]
+fn test_deposit_wrong_custodian_address() {
+    let custodian_address = "096DE9C2B8A5B8c22cEe3289B101f6960d68E510";
```
sept-en, line 470: Do we have any reason to set this specific address to ensure that the deposit fails if the proof is coming from `EthCustodian` that was not specified for this contract?
I'd prefer to set the initial value for the address as `0000000000000000000000000000000000000001`, so it's clear that we are not checking for some specific value.

sept-en (approved):

tests/test_connector.rs
sept-en: Btw, as the type of errors, we are checking there has the following format when converted to string:
`Action #0: Smart contract panicked: ERR_PROOF_EXIST [src/some_file.rs:LINE_NUMBER:COLUMN_NUMBER]`,
it makes sense to check that it contains not only the `err_msg` but `"Smart contract panicked: " + err_msg + " "`.
In this case, we are exhaustively guarded in case if we introduce errors that have the same matching prefix (e.g. `ERR_NOT_ENOUGH_BALANCE` and `ERR_NOT_ENOUGH_BALANCE_FOR_FEE`).

tests/test_connector.rs
mrLSD: It's just wrong address. I think we don't need zero account.

tests/test_connector.rs
mrLSD: I thnoks we don't need constants for one-time-used values.

tests/test_connector.rs
mrLSD: I think we don't need constants for one-time-used values.

tests/test_connector.rs
mrLSD: So it's a guaranty that ExecutionOutcome contain exact error because:
1. We fetch ExecutionStatus::Failure - so it contains a failure message
2. We fetch specific messages from failure content.
So it's pretty enough.

tests/test_connector.rs
sept-en: The purpose of this test is to check the mismatch for `EthCustodian` addresses. So it's not very clear from where the specified value is coming from. But for sure this should be wrong address, so I think it's ok if we specify the hardcoded account with `1` address (not zero)

tests/test_connector.rs
sept-en: I suggested this for consistency. But I'm ok if we keep it like this, as this is quite minor.

tests/test_connector.rs
sept-en: This works until we have no single error message that is the subset of any other.
I recommend adding this, though it's fine for me to keep it as is to save some time.

mrLSD: If we have the same error messages for the same process flow we don't have any possibilities to extract that message.

sept-en: Ok, I'm fine if we don't add this right now. Though we are adding another thing to the technical debt :)

PR #107, opened 2021-05-27
Fix(precompiles): Match on full address when dispatching on precompiles

### LINKED ISSUE #82: Potential collision with precompiles in address space
To detect if a call was directed to a precompile we are using the following code:

https://github.com/aurora-is-near/aurora-engine/blob/master/src/precompiles/mod.rs#L125

```
    match address.to_low_u64_be() {
        1 => Some(ECRecover::run(input, target_gas, context)),
        2 => Some(SHA256::run(input, target_gas, context)),
        3 => Some(RIPEMD160::run(input, target_gas, context)),
        4 => Some(Identity::run(input, target_gas, context)),
        5 => Some(ModExp::<Byzantium>::run(input, target_gas, context)),
        6 => Some(BN128Add::<Istanbul>::run(input, target_gas, context)),
        7 => Some(BN128Mul::<Istanbul>::run(input, target_gas, context)),
        8 => Some(BN128Pair::<Istanbul>::run(input, target_gas, context)),
        9 => Some(Blake2F::run(input, target_gas, context)),
        // Not supported.
        _ => None,
    }
```

Notice that we are truncating the address (20 bytes) to the last (8 bytes) and we check if it matches any of the precompiles. I think we should not truncate the address and check directly with the 20 bytes, otherwise we are exposing ourselves to collisions between real deployed addresses and precompiles. 64 bits has still some room so collisions should not happen naturally, but I think it is not unreasonable to believe that it is easy to create an address with last 8 bytes to match the one of our precompiles.

I haven't reason about security implications about it, but let's avoid hitting this issue anyway.

birchmd:
Fixes #82

### DIFF
--- a/src/precompiles/mod.rs
+++ b/src/precompiles/mod.rs
@@ -17,18 +17,6 @@ use crate::precompiles::secp256k1::ECRecover;
 use crate::prelude::{Address, Vec};
 use evm::{Context, ExitError, ExitSucceed};
 
-/// Exit to Ethereum precompile address (truncated to 8 bytes)
-///
-/// Address: `0xb0bd02f6a392af548bdf1cfaee5dfa0eefcc8eab`
-/// This address is computed as: `&keccak("exitToEthereum")[12..]`
-const EXIT_TO_ETHEREUM_ID: u64 = 17176159495920586411;
-
-/// Exit to NEAR precompile address (truncated to 8 bytes)
-///
-/// Address: `0xe9217bc70b7ed1f598ddd3199e80b093fa71124f`
-/// This address is computed as: `&keccak("exitToNear")[12..]`
-const EXIT_TO_NEAR_ID: u64 = 11421322804619973199;
-
 /// A precompile operation result.
 type PrecompileResult = Result<(ExitSucceed, Vec<u8>, u64), ExitError>;
 
@@ -88,12 +76,12 @@ pub fn homestead_precompiles(
         None => return Some(PrecompileResult::Err(ExitError::OutOfGas)),
     };
 
-    match address.to_low_u64_be() {
-        1 => Some(ECRecover::run(input, target_gas, context)),
-        2 => Some(SHA256::run(input, target_gas, context)),
-        3 => Some(RIPEMD160::run(input, target_gas, context)),
-        EXIT_TO_NEAR_ID => Some(ExitToNear::run(input, target_gas, context)),
-        EXIT_TO_ETHEREUM_ID => Some(ExitToEthereum::run(input, target_gas, context)),
+    match address.0 {
+        ECRecover::ADDRESS => Some(ECRecover::run(input, target_gas, context)),
+        SHA256::ADDRESS => Some(SHA256::run(input, target_gas, context)),
+        RIPEMD160::ADDRESS => Some(RIPEMD160::run(input, target_gas, context)),
+        ExitToNear::ADDRESS => Some(ExitToNear::run(input, target_gas, context)),
+        ExitToEthereum::ADDRESS => Some(ExitToEthereum::run(input, target_gas, context)),
         _ => None,
     }
 }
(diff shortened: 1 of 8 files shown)


src/precompiles/mod.rs
```diff
@@ -168,20 +156,50 @@ pub fn berlin_precompiles(
...
+        y_bytes[8],
+        y_bytes[9],
+        y_bytes[10],
+        y_bytes[11],
+        y_bytes[12],
+        y_bytes[13],
+        y_bytes[14],
+        y_bytes[15],
+    ]
+}
```
sept-en, line 205: This is quite simple, but I'd prefer if we have unit-tests for at least 0..9 addresses to make sure that we never ever accidentally break it. Otherwise, that would be hard to debug in the future.

sept-en (approved):
Great work! I really like this approach for handling addresses, very neat!

Added a minor comment regarding tests. Though if we are in hurry to merge it, feel free to ignore.

src/precompiles/mod.rs
birchmd: The function was being implicitly tested by the tests checking the `ExitToNear` and `ExitToEthereum` addresses, as well as the standard precompiles sanity test (since that relies on dispatching to the right addresses).

That said, you are right that more tests are always a good thing. I have added some tests specifically for this function in https://github.com/aurora-is-near/aurora-engine/pull/107/commits/fe37d3ba83270aa26541d1d8c9d7ce63b7a438c4

sept-en: @birchmd got it, thanks for adding this. LGTM!

src/precompiles/mod.rs
```diff
@@ -168,20 +156,99 @@ pub fn berlin_precompiles(
...
+        y_bytes[7],
+        y_bytes[8],
+        y_bytes[9],
+        y_bytes[10],
+        y_bytes[11],
+        y_bytes[12],
+        y_bytes[13],
+        y_bytes[14],
+        y_bytes[15],
+    ]
```
birchmd, line 204: After I wrote this, I learned that `loop` and `if` are allowed in `const fn`. This means an alternate implementation would be:

```rust
    let mut i = 0;
    let mut bytes = [0u8; 20];
    loop {
        if i < 4 {
            bytes[i] = x_bytes[i];
        } else if i < 20 {
            bytes[i] = y_bytes[i - 4];
        } else {
            break;
        }
        i += 1;
    }

    bytes
```

Note: `for` and `&mut` (and therefore `copy_from_slice`) are currently not allowed in `const fn`.

The `loop` implementation is less boilerplate-y than the current implementation, but arguably harder to read (more cognitive overhead). I think it takes a little longer to compile too (note this function is only used at compile-time, so its efficiency only impacts the compilation of the project), which makes sense because the loop has a lot more going on than just copying bytes from one place to another (the arithmetic of tracking the index, the conditional figuring out if it should take from `x_bytes` or `y_bytes`, etc.)

So ultimately I think the current implementation is better than the `loop` one, but I thought I would write this up for academic purposes.

sept-en: Yeah, I also noticed this part which looked a little bit strange for me, but I knew that if you did it, there was a reason for that. So immediately tried different approaches locally and ended up with the same conclusions, that I can't make it look prettier, rather harder to read :)

So I'm also for keeping it as the same. As approaches that work for `const fn` for this case are quite ugly.

artob (approved):
LGTM :+1:

An interesting approach; I'd love to know, though, how Rust compiles those `match` expressions: do they just turn into a sequence of `memcmp` branches?

birchmd:
I wasn't totally sure how it would compile down, so I checked in [Compile Explorer](https://godbolt.org/).

The result is here: https://gist.github.com/birchmd/44ffe28000662fd1f1f0abe7c528c76c

Since the byte arrays are known at compile time, it looks like the byte comparisons get hard coded into the instructions themselves.

artob:
> I wasn't totally sure how it would compile down, so I checked in Compile Explorer.

Thanks for that. Here's the corresponding WebAssembly output: https://godbolt.org/z/s3vK4zPdn

That's about 5,400 WebAssembly instructions, with 20 distinct blocks, for a `match` expression containing 9 branches. That probably compiles to about 10+ KB in the binary.

PR #109, opened 2021-05-28
Eth-connector: merge `develop` branch

sept-en:
Merge `develop` branch & resolve conflicts.


src/engine.rs
```diff
@@ -94,6 +94,22 @@ impl ExitIntoResult for ExitReason {
     }
 }
 
+//TODO: check as I added `Debug` trait there. Perhaps this wasn't added earlier intentionally.
```
birchmd, line 97: It wasn't needed previously. What do we need it for now?

src/engine.rs
```diff
@@ -523,9 +543,13 @@ impl ApplyBackend for Engine {
                     reset_storage,
                 } => {
                     Engine::set_nonce(&address, &basic.nonce);
+
+                    //TODO: check this part
                     // Apply changes for eth-connector
                     EthConnectorContract::get_instance()
                         .internal_set_eth_balance(&address, &basic.balance);
+                    Engine::set_balance(&address, &Wei::new(basic.balance));
```
birchmd, line 551: Looks like we need to refactor `internal_set_eth_balance` to take the `Wei` newtype. And we can remove the direct `set_balance` call.

src/fungible_token.rs
```diff
@@ -426,7 +434,9 @@ impl FungibleToken {
     pub fn accounts_insert(&self, account_id: &str, amount: Balance) {
         if !self.accounts_contains_key(account_id) {
             let key = Self::get_statistic_key();
+            //TODO: verify `read_u64` unwrapping
```
birchmd, line 437: I think the way you are unwrapping is fine. When I was writing the `read_u64` type signature I was a little unhappy with the nested `Option<Result<...>>` for exactly this reason. But also that return type corresponds to what is happening (there may or may not be a value, and that value may or may not be a `u64`), so I decided to leave it. If others feel like it is too cumbersome to use then we could switch to a less explicit return type in favour of some error handling inside `read_u64` itself.

birchmd (approved):
Nice job!

I think we can remove most of the TODOs you called out. The only thing I think needs additional attention is refactoring `internal_set_eth_balance` to accept `Wei`, but we could do that as a follow-up PR if we are in a hurry to merge this one.

src/engine.rs
sept-en: This is needed for the following piece of code: https://github.com/aurora-is-near/aurora-engine/blob/7ac9995edfa685dc4179f3ba11fe8d84ee2c112e/src/prover.rs#L133
The other approach is to use `sdk_unwrap()`. For that we need to move `SdkUnwrap` trait out of the struct and have it without `contract` feature. Then we need to introduce conditional compilation for its implementations (as the `sdk` module is not available without the `contract` feature.

So I think just adding the `Debug` and use `unwrap` just for that specific case is easier approach and is not that messy.

src/fungible_token.rs
birchmd: Another possibility would be to define a type like

```rust
enum MaybeResult<T, E> {
    Some(T),
    Err(E),
    None,
}

impl<T, E> MaybeResult {
    pub fn unwrap_or(self, default: T) -> T {
        match self {
            Self::Some(t) => t,
            Self::Err(_) | Self::None => default,
        }
    }
}
```

But this might be a little overkill if we're only using it once for `read_u64`.

sept-en: Yeah, I don't think we need to define a new type, because in most cases we use just `sdk_unwrap` for it. And there are only a couple of cases where we need to add a couple of sequential `unwrap_or`.

src/engine.rs
birchmd: Sure, it's ok to have `Debug` and `unwrap` for now.

I think some day (probably in the medium to long term) we should go through and eliminate all `unwrap` / `expect` calls in wasm code in favour of the sdk versions. This should make the binary smaller, which will be nice. But it won't make a difference right now because the majority of the size comes from secp256 stuff.

sept-en: I agree. For this specific case, I thought that `Debug` was not added intentionally to force everyone to use `sdk_unwrap()` instead.

src/fungible_token.rs
birchmd: Anotherother possibility, and this is probably what I should have done in the first place, would be to capture the possibility of the missing value in the error type.

```rust 
pub(crate) fn read_u64(key: &[u8]) -> Result<u64, ReadU64Error> {
    ...
}

pub(crate) enum ReadU64Error {
    MissingValue,
    InvalidU64,
}
```

sept-en: Yes, probably makes sense to add this as well. But Anyway I suggest doing this in a separate PR to not extend the scope of the current one.

etc/state-migration-test/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+overflow-checks = true
+lto = true
+panic = "abort"
+incremental = false
+codegen-units = 1
+rpath = false
+
+[dependencies]
+borsh = { version = "0.8.2", default-features = false }
+aurora-engine = { path = "../../", default-features = false, features = ["contract", "sha2"] }
```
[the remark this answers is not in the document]
sept-en: The `"contract"` feature is indeed needed to import the `crate::connector` module from the `engine.rs`.

src/fungible_token.rs
```diff
@@ -438,7 +447,9 @@ impl FungibleToken {
     /// Get accounts counter for statistics
     /// It represents total unique accounts.
     pub fn get_accounts_counter(&self) -> u64 {
-        sdk::read_u64(&Self::get_statistic_key()).unwrap_or(0)
+        sdk::read_u64(&Self::get_statistic_key())
+            .unwrap_or(Ok(0))
+            .unwrap_or(0)
```
mrLSD, line 452: It looks really strange.

src/lib.rs
```diff
@@ -43,6 +43,40 @@ mod test_utils;
 #[cfg(test)]
 mod tests;
 
+#[cfg(target_arch = "wasm32")]
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
+#[cfg(target_arch = "wasm32")]
```
mrLSD, line 50: Why did it move up top of `contract` feature?

src/sdk.rs
```diff
@@ -157,20 +162,28 @@ mod exports {
 
 pub fn read_input() -> Vec<u8> {
     unsafe {
-        exports::input(0);
-        let bytes: Vec<u8> = vec![0; exports::register_len(0) as usize];
-        exports::read_register(0, bytes.as_ptr() as *const u64 as u64);
+        exports::input(INPUT_REGISTER_ID);
```
mrLSD, line 165: @birchmd IMHO all those changes really redundant.
WASM has an infinity register stack. We don't need to unify as some standard `get/read/write` register flow. Most important read from the same register. And it was. it will be misleading that we should only use one given register.

src/sdk.rs
```diff
@@ -191,29 +204,42 @@ pub fn return_output(value: &[u8]) {
...
             None
         }
     }
 }
 
 /// Read u64 from storage at given key.
-pub fn read_u64(key: &[u8]) -> Option<u64> {
-    unsafe {
-        if exports::storage_read(key.len() as u64, key.as_ptr() as u64, 0) == 1 {
+pub(crate) fn read_u64(key: &[u8]) -> Option<Result<u64, InvalidU64>> {
```
mrLSD, line 233: @birchmd I think we should return only `Result` type and remove `Option`. It's not informative. We can return `Err(...)` for the first part of the contract too.

birchmd: Yes, agreed. See https://github.com/aurora-is-near/aurora-engine/pull/109#discussion_r641574473

src/sdk.rs
birchmd: I think giving constants a name is useful for two reasons:

1. It provides a semantic meaning to an otherwise ambiguous value (`0` could mean lots of things, but here it is a register ID)
2. It makes changing the value less error prone in the future because all occurrences of it now point back to one definition.  For example, if we decide to reserve register 0 for some purpose then we can change `INPUT_REGISTER_ID` to be some other value easily without carefully double-checking we changed all the right places.

src/lib.rs
birchmd: It's needed for a test contract which uses the production contract as a dependency, but I can't have the production public exports (i.e. `contract` feature) because I want to override the definition in the test contract and you can't export a two methods with the same name.

src/fungible_token.rs
artob: https://github.com/aurora-is-near/aurora-engine/pull/118

PR #113, opened 2021-05-28
Eth-connector: add the way to set contract data. Minor refactoring.

sept-en:
* Add `set_eth_connector_contract_data()` method.
* Minor refactoring: `init_contract()` method now accepts `InitCallArgs`
  instead of making IO operations in it.


birchmd (approved):
LGTM

PR #115, opened 2021-05-31
Remove dependency on Lunarity

### LINKED ISSUE #34: Upgrade logos dependency in lunarity
Our project has a dependency that does not compile on latest nightly

```
$ cargo +nightly check
   Compiling logos-derive v0.7.7
error[E0061]: this function takes 1 argument but 2 arguments were supplied
   --> /home/birchmd/.cargo/registry/src/github.com-1ecc6299db9ec823/logos-derive-0.7.7/src/lib.rs:55:20
    |
55  |             extras.insert(util::ident(&ext), |_| panic!("Only one #[extras] attribute can be declared."));
    |                    ^^^^^^ -----------------  ----------------------------------------------------------- supplied 2 arguments
    |                    |
    |                    expected 1 argument
    |
note: associated function defined here

error[E0061]: this function takes 1 argument but 2 arguments were supplied
   --> /home/birchmd/.cargo/registry/src/github.com-1ecc6299db9ec823/logos-derive-0.7.7/src/lib.rs:89:23
    |
89  |                 error.insert(variant, |_| panic!("Only one #[error] variant can be declared."));
    |                       ^^^^^^ -------  -------------------------------------------------------- supplied 2 arguments
    |                       |
    |                       expected 1 argument
    |
note: associated function defined here

error[E0061]: this function takes 1 argument but 2 arguments were supplied
   --> /home/birchmd/.cargo/registry/src/github.com-1ecc6299db9ec823/logos-derive-0.7.7/src/lib.rs:93:21
    |
93  |                 end.insert(variant, |_| panic!("Only one #[end] variant can be declared."));
    |                     ^^^^^^ -------  ------------------------------------------------------ supplied 2 arguments
    |                     |
    |                     expected 1 argument
    |
note: associated function defined here

error: aborting due to 3 previous errors

For more information about this error, try `rustc --explain E0061`.
error: could not compile `logos-derive`

To learn more, run the command again with --verbose.
```

This `logos-derive` dependency comes from our [dependency on `lunarity`](https://github.com/aurora-is-near/aurora-engine/blob/8b767e44a5378da090c3ceb9113af4d7378f9216/Cargo.toml#L49). Fortunately, we are already working off a fork which we control (it's on Illia's GitHub), and the latest version of `logos` (v0.12) _does_ compile with latest nightly. So we can resolve this issue by upgrading `logos` to the latest version in our fork of the `lunarity` project.

Unfortunately, this upgrade is non-trivial because there were [several breaking API changes](https://github.com/maciejhirsz/logos/releases/tag/v0.11.0) in `logos`, so some code in `lunarity` will need to be refactored to work with the latest version. This issue represents the work to do that refactoring and thus have a version of `lunarity` that compiles in latest nightly for our project to depend on.

This is probably a low priority item since it will likely not be blocking us until we decide to upgrade our `rust-toolchain` file to a later version of rust.

birchmd:
We do not want to depend on Lunarity any more because it uses a difference license than the rest of our project. It also doesn't compile on latest nightly because it depends on an old version of Logos.

In this PR we remove the dependency on Lunarity and instead  depend on (the most recent version of) Logos directly. This required adding a bit more code to `meta_parsing` because we were using Lunarity for parsing the type strings. I think the new code is a little better than the old code anyway because the state machine is a little simpler as the array bracket parsing can be delegated to the lexer.

With this change the project now compiles under latest nightly, though `cargo +nightly clippy` doesn't pass yet because there are some errors. I'll fix those in a follow-up PR.

Closes #34

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -86,9 +86,9 @@ version = "0.1.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8d3a45e77e34375a7923b1e8febb049bb011f064714a8e17a1a616fef01da13d"
 dependencies = [
- "proc-macro2 1.0.26",
- "quote 1.0.9",
- "syn 1.0.57",
+ "proc-macro2",
+ "quote",
+ "syn",
 ]
 
 [[package]]
@@ -129,7 +129,7 @@ dependencies = [
  "git2",
  "hex",
  "libsecp256k1",
- "lunarity-lexer",
+ "logos",
  "near-crypto 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "near-primitives-core 0.1.0 (git+https://github.com/near/nearcore?rev=3744f07e13bf43a9522fb39fa8f6f128396d0e1f)",
  "near-sdk",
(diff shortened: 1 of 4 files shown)


mrLSD (approved):

joshuajbouw (approved):
Good work!

artob (approved):
LGTM :+1: Nice work!

PR #116, opened 2021-06-01
Update to `evm_core` 0.27.1

joshuajbouw:
This updates the Aurora engine to the latest `evm-core`.

Do note that the merge to `master` is intentional given the nature of this update.

This also does include the logs which @mfornet had added to upstream so a further PR should be created to utilise logs.

### DIFF
--- a/src/precompiles/blake2.rs
+++ b/src/precompiles/blake2.rs
@@ -81,11 +81,7 @@ impl Precompile for Blake2F {
         let finished = input[212] != 0;
 
         let output = blake2::blake2b_f(rounds, h, m, t, finished).to_vec();
-        Ok(PrecompileOutput {
-            cost,
-            output,
-            ..Default::default()
-        })
+        Ok(PrecompileOutput::without_logs(cost, output))
     }
 }
 
--- a/src/precompiles/blake2.rs
+++ b/src/precompiles/blake2.rs
@@ -199,12 +199,12 @@ mod tests {
             01",
         )
         .unwrap();
-        Blake2F::run(&input, 12, &new_context()).unwrap().1
+        Blake2F::run(&input, 12, &new_context()).unwrap().output
     }
 
     fn test_blake2f_r_12() -> Vec<u8> {
         let input = hex::decode(INPUT).unwrap();
-        Blake2F::run(&input, 12, &new_context()).unwrap().1
+        Blake2F::run(&input, 12, &new_context()).unwrap().output
     }
 
     fn test_blake2f_final_block_false() -> Vec<u8> {
@@ -222,7 +222,7 @@ mod tests {
             00",
         )
         .unwrap();
-        Blake2F::run(&input, 12, &new_context()).unwrap().1
+        Blake2F::run(&input, 12, &new_context()).unwrap().output
     }
 
     #[test]
--- a/src/precompiles/blake2.rs
+++ b/src/precompiles/blake2.rs
@@ -1,6 +1,6 @@
-use crate::precompiles::{Precompile, PrecompileResult};
+use crate::precompiles::{Precompile, PrecompileOutput, PrecompileResult};
 use crate::prelude::{mem, Borrowed, TryInto};
-use evm::{Context, ExitError, ExitSucceed};
+use evm::{Context, ExitError};
 
 /// Blake2 costs.
 mod costs {
@@ -80,8 +80,12 @@ impl Precompile for Blake2F {
         }
         let finished = input[212] != 0;
 
-        let res = blake2::blake2b_f(rounds, h, m, t, finished).to_vec();
-        Ok((ExitSucceed::Returned, res, cost))
+        let output = blake2::blake2b_f(rounds, h, m, t, finished).to_vec();
+        Ok(PrecompileOutput {
+            cost,
+            output,
+            ..Default::default()
+        })
     }
 }
 
--- a/src/precompiles/mod.rs
+++ b/src/precompiles/mod.rs
@@ -129,7 +129,7 @@ pub fn homestead_precompiles(
         ExitToEthereum::ADDRESS => Some(ExitToEthereum::run(input, target_gas, context)),
         _ => None,
     };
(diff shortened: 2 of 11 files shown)


artob:
@joshuajbouw The test suite is failing, please fix.

src/precompiles/mod.rs
```diff
@@ -87,7 +118,8 @@ pub fn homestead_precompiles(
         #[cfg(feature = "exit-precompiles")]
         ExitToEthereum::ADDRESS => Some(ExitToEthereum::run(input, target_gas, context)),
         _ => None,
-    }
+    };
+    output.map(|res| res.map(|o| o.into()))
```
birchmd, line 122: Could avoid the closure and write `output.map(|res| res.map(Into::into))`. Probably doesn't matter much either way, but maybe slightly easier to read without the closure.

(same applies to Byzantium, Istanbul, etc.)

src/precompiles/blake2.rs
```diff
@@ -80,8 +80,12 @@ impl Precompile for Blake2F {
         }
         let finished = input[212] != 0;
 
-        let res = blake2::blake2b_f(rounds, h, m, t, finished).to_vec();
-        Ok((ExitSucceed::Returned, res, cost))
+        let output = blake2::blake2b_f(rounds, h, m, t, finished).to_vec();
+        Ok(PrecompileOutput {
+            cost,
+            output,
+            ..Default::default()
```
birchmd, line 87: This is cool syntax, I don't think I didn't know you could do this. I'm not sure if I like it though. It means if we add a new field to `PrecompileOutput` in the future and we do not want it to have a default value then the compiler will not help us find the places where the struct initialization needs to be updated.

Maybe a better way to do this would be to define a `new` function on `PrecompileOutput` that only takes `cost` and `output` arguments.

src/precompiles/secp256k1.rs
```diff
@@ -145,38 +155,46 @@ mod tests {
...
+            .unwrap()
+            .output;
         assert_eq!(res, expected);
 
         // Why is this test returning an address???
         // let input = hex::decode("47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad000000000000000000000000000000000000000000000000000000000000001b000000000000000000000000000000000000000000000000000000000000001bffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff").unwrap();
         // let expected = hex::decode("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff").unwrap();
         //
-        // let res = ecrecover_raw(&input, Some(500)).unwrap().1;
+        // let res = ecrecover_raw(&input, Some(500)).unwrap().output;
```
birchmd, line 197: Lol, updating the commented out code is very thorough! But maybe we don't need this block of commented out code any more?

birchmd (approved):
LGTM

Two small style comments, but non-blocking if this is sufficiently urgent.

birchmd:
Since this targets `master` we'll also need to have a PR merging `master` back into `develop` after this is merged.

src/precompiles/secp256k1.rs
joshuajbouw: `ctrl + R` and going through them one by one to make sure it changed what you want :P, but anyways, a similar commented out code was in another library, and there potentially is a quirk here that shouldn't be happening from what I recall. From looking at the implementation though, it should still be compatible.

joshuajbouw:
> Since this targets `master` we'll also need to have a PR merging `master` back into `develop` after this is merged.

Correct

src/precompiles/blake2.rs
joshuajbouw: I had considered that as I'm aware of the downfall of that. This also has been a bit of topic of hotly debate for use of those as arguments on public methods... For example, you get the benefit of methods being backwards compatible, but that -may- have undesirable consequences. Since this isn't meant to be public, I was putting faith in the team to update all arguments that would use this as it should be pretty straightforward to catch them all. That and I feel quite doubtful that we would add more to it.

src/precompiles/mod.rs
joshuajbouw: Ah thats right, I forgot you could do that now. I'm indifferent either way.

src/precompiles/blake2.rs
birchmd: Discussed with @joshuajbouw privately. We both agree the probability of this causing a problem in the future is low, but also that using a constructor function here is also ok and erring on the side of compile-time safety is preferred for internal structures like this. I'll push a change shortly.

birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/116/commits/d7e29374eef5aa5158dab02c09ce5638766b4b8f

src/precompiles/mod.rs
birchmd: I like the version without closure slightly better, so changed in https://github.com/aurora-is-near/aurora-engine/pull/116/commits/613982ca942cacb2784dcfb09209b2c55b513b8c

mfornet (commented):
LGTM

All changes I see are related to the update of precompiles interfaces re: logs.

Did the critical patch on the EVM had any noticeable impact in our engine that we can check? I couldn't see any write up, and it is hard to tell alone from [the code](https://github.com/rust-blockchain/evm/commit/ce4afc35f0ccd6a0afccfc610b8e1f2879b3804d).

joshuajbouw:
> LGTM
> 
> All changes I see are related to the update of precompiles interfaces re: logs.
> 
> Did the critical patch on the EVM had any noticeable impact in our engine that we can check? I couldn't see any write up, and it is hard to tell alone from [the code](https://github.com/rust-blockchain/evm/commit/ce4afc35f0ccd6a0afccfc610b8e1f2879b3804d).

Hence why I had to carefully go through it as there was a lack of release notes.

artob (approved):
LGTM :+1: Great work, @joshuajbouw, that's a nice clean-up.

PR #120, opened 2021-06-03
EthConnector: add view function to check whether the proof is used.

sept-en:
This feature was requested for both the aurora-frontend and the event-relayer components. The solution seems to be most solid and convenient. IMO this doesn't expose any vulnerability.


src/parameters.rs
```diff
@@ -145,6 +145,14 @@ pub struct DepositCallArgs {
...
 }
 
+/// Eth-connector isUsedProof arguments
+#[cfg(feature = "engine")]
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct IsUsedProofCallArgs {
+    /// Proof data
+    pub proof: Proof,
+}
+
```
mfornet, line 155: Why do we need this, instead of just using `Proof`

src/lib.rs
```diff
@@ -401,6 +402,15 @@ mod contract {
         EthConnectorContract::get_instance().finish_deposit_near();
     }
 
+    #[no_mangle]
+    pub extern "C" fn is_used_proof() {
+        let args = IsUsedProofCallArgs::try_from_slice(&sdk::read_input()).expect(ERR_FAILED_PARSE);
```
mfornet, line 407:
```suggestion
        let args = Proof::try_from_slice(&sdk::read_input()).expect(ERR_FAILED_PARSE);
```

src/parameters.rs
sept-en: I added this for the sake of consistency: e.g. we have `StorageBalanceOfCallArgs`, `RegisterRelayerCallArgs`, `PauseEthConnectorCallArgs`, and many others having only just one parameter in the struct.
So it seems for me having just `Proof` here is not complying with the approach we use for call args everywhere in the `aurora-engine`.
But if you think this is mandatory to change, please let me know and I will change.

birchmd: In my view, the trade-off here is: adding a little complexity now (having two structs that are functionally identical) in exchange for making it easier to add arguments to `is_used_proof` (would simply mean adding a new field to `IsUsedProofCallArgs`). The question is then whether we think the complexity is worth the benefit, i.e. whether we expect `is_used_proof` to someday take more than just the proof itself. If not then maybe we should use `Proof` directly. I don't really know one way or the other how likely it is, and I also don't think having this extra struct hurts the code much.

Long story short, I don't care much either way whether we have `IsUsedProofCallArgs` or use `Proof` directly. And maybe I lean slightly towards having it because I do like the "future-proofing" nature of it.

birchmd (approved):
LGTM! Nice work.

mfornet (approved):

artob (approved):
LGTM :+1: Good work!

PR #121, opened 2021-06-03
Add evm-bully=yes build to CI

frankbraun:
The contract build for the evm-bully (`make evm-bully=yes`) often breaks.
Add it to CI to prevent that from happening.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,6 +28,20 @@ jobs:
         with:
           command: test
           args: --locked --verbose
+  bully-build:
+    name: Bully build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone the repository
+        uses: actions/checkout@v2
+      - name: Install the toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly-2021-03-25
+          override: true
+      - run: make evm-bully=yes
+      - run: ls -lH release.wasm
 env:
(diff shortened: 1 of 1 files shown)


sept-en (approved):
LGTM!

I also think that we should use `develop` as a base branch having the release protocol we have. I.e. the `master` points to the latest deployed version and the `develop` branch is the latest stable one that will be uploaded into `master` and released after that.

.github/workflows/builds.yml
```diff
@@ -61,3 +61,31 @@ jobs:
...
+          name: bully.wasm
+          path: bully.wasm
+      - name: Update the latest release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          tag: latest
+          file: bully.wasm
+          overwrite: true
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Update the latest tag
```
artob, line 90: You need to remove the `Update the latest release` and `Update the latest tag` steps, lest you introduce a race condition for updating the `latest` tag.

In addition, it is also unlikely that you actually need the `Upload the release.wasm artifact` step (which also wasn't properly renamed).

artob (changes requested):
This should be targeted at `develop`, not `master`. And there are copy-paste artifacts to remove.

frankbraun:
I changed it to target `develop` instead of `master`. Also moved the tests over to `tests.yml`, because this should be checked on every PR, not just for releases.

artob (approved):

PR #123, opened 2021-06-03
Fix(engine): handle transaction gas limit properly

### LINKED ISSUE #119: Correctly handle gas limits on transactions
Presently we allow any gas limit to be specified on a transaction passed to `submit`. This is incorrect behaviour relative to Ethereum where there is a notion of "intrinsic gas".

The current behaviour is causing a problem because (a) it is an incompatibility with Ethereum, and we are always pushing for maximum compatibility; and (b) the relayer has trouble handling transactions with 0 gas limit (and these would be invalid transactions if we had the correct logic in the Engine).

This issue is completed when the following is done:

- [ ] Have a function which computes "intrinsic gas" for transactions
- [ ] Verify all transactions passed via `submit` have a gas limit greater than or equal to the intrinsic gas
- [ ] Respect the gas limit when processing transactions passed to `submit` (i.e. fail with OOG when computation exceeds limit)

References:

* https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L259
* https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L114
* https://github.com/ethereum/yellowpaper/blob/master/Paper.tex#L593

birchmd:
Fixes #119

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -145,7 +145,7 @@ pub struct Engine {
 }
 
 // TODO: upgrade to Berlin HF
-const CONFIG: &Config = &Config::istanbul();
+pub(crate) const CONFIG: &Config = &Config::istanbul();
 
 /// Key for storing the state of the engine.
 const STATE_KEY: &[u8; 5] = b"STATE";
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -294,7 +294,7 @@ impl Engine {
     pub fn deploy_code_with_input(&mut self, input: Vec<u8>) -> EngineResult<SubmitResult> {
         let origin = self.origin();
         let value = Wei::zero();
-        self.deploy_code(origin, value, input)
+        self.deploy_code(origin, value, input, u64::MAX)
     }
 
     pub fn deploy_code(
@@ -302,15 +302,19 @@ impl Engine {
         origin: Address,
         value: Wei,
         input: Vec<u8>,
+        gas_limit: u64,
     ) -> EngineResult<SubmitResult> {
-        let mut executor = self.make_executor();
+        let mut executor = self.make_executor(gas_limit);
         let address = executor.create_address(CreateScheme::Legacy { caller: origin });
         let (status, result) = (
-            executor.transact_create(origin, value.raw(), input, u64::MAX),
+            executor.transact_create(origin, value.raw(), input, gas_limit),
             address,
         );
         let is_succeed = status.is_succeed();
-        status.into_result()?;
+        if let Err(e) = status.into_result() {
+            Engine::increment_nonce(&origin);
+            return Err(e);
+        }
         let used_gas = executor.used_gas();
         let (values, logs) = executor.into_state().deconstruct();
         self.apply(values, Vec::<Log>::new(), true);
@@ -327,7 +331,7 @@ impl Engine {
         let origin = self.origin();
         let contract = Address(args.contract);
         let value = Wei::zero();
-        self.call(origin, contract, value, args.input)
+        self.call(origin, contract, value, args.input, u64::MAX)
     }
 
     pub fn call(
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -239,16 +239,30 @@ mod contract {
...
+                if signed_transaction.transaction.gas < intrinsic_gas.into() {
+                    sdk::panic_utf8(b"ERR_INTRINSIC_GAS")
+                }
+            }
+        }
+
         // Figure out what kind of a transaction this is, and execute it:
         let mut engine = Engine::new_with_state(state, sender);
         let value = signed_transaction.transaction.value;
+        let gas_limit = u256_to_u64(signed_transaction.transaction.gas);
```
birchmd, line 258: I'm not sure if this is correct behavior or not. Presently it is taking the gas limit mod `u64::MAX + 1` to make it fit into `u64` (since that is the type expected by SputnikVM). But should I panic if the gas limit is larger than `u64::MAX` instead of silently ignoring the larger number? Or maybe I should truncate to `u64::MAX` in the case it is larger? Is there an Ethereum spec for this?

src/engine.rs
```diff
@@ -382,15 +387,15 @@ impl Engine {
         value: Wei,
         input: Vec<u8>,
     ) -> EngineResult<Vec<u8>> {
-        let mut executor = self.make_executor();
+        let mut executor = self.make_executor(u64::MAX);
```
birchmd, line 390: Do view calls need a gas limit?

birchmd (commented):
Some clarifying questions:

src/engine.rs
artob: Yes, for compatibility with mutative calls, a gas limit is passed in: https://eth.wiki/json-rpc/API#eth_call

src/lib.rs
```diff
@@ -276,6 +290,7 @@ mod contract {
             meta_call_args.contract_address,
             meta_call_args.value,
             meta_call_args.input,
+            u64::MAX, // TODO: is there a gas limit with meta calls?
```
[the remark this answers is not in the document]
artob: We're not at all using the meta-transaction machinery at the moment, so it's fine to leave it as you did, with a `TODO` comment.

src/lib.rs
artob: Gas limits are `uint64` throughout the Geth code base, so there's no way it can handle a larger value. However, I'm unsure what Geth does if receiving a transaction with a larger value. I'll endeavor to find out and come back to you.

artob: Specifying the same max gas limit with another endpoint implementation:

`{'code': -32000, 'message': 'exceeds block gas limit'}`

artob: Looks like Geth specifies the error code `ErrGasUintOverflow` (`gas uint64 overflow`) for situations where the gas would overflow:

https://github.com/ethereum/go-ethereum/blob/master/core/error.go#L64

We probably ought to define a corresponding error code for this, so that we can return the corresponding error message in the relayer, for full Geth compatibility.

birchmd: Made to panic if gas limit is larger than `u64::MAX` in https://github.com/aurora-is-near/aurora-engine/pull/123/commits/57934d44cf42e8e2ecbbd4d32508c83175441d65

src/engine.rs
birchmd: Gas limit added to `Engine::view` in https://github.com/aurora-is-near/aurora-engine/pull/123/commits/b8879c8dd53cb2c8b542180432d55003158e69ab but it's not really used yet because the only way `view` is called is from `view_with_args` which does not have a gas limit (since it is Near-native). However, we will add an `eth_call` contract method in the future which will pass a user-input gas limit to `Engine::view`.

joshuajbouw (approved):
Straight forward

sept-en (approved):
LGTM!

artob (approved):
LGTM :+1: Great work, @birchmd. The only thing that stuck out is that we might better define a max gas constant instead of using `u64::MAX` all over the place, but we can address that later.

PR #124, opened 2021-06-03
EthConnector: use strings in JSON for amounts.

sept-en:
The integer values in JSON are limited by `u64`. So we need to use strings instead and then convert them to the needed integer type.


mrLSD:
Unfortunately that PR duplicate: #125

sept-en:
@mrLSD The previous PR can't duplicate the future one :) Though I didn't know that you were working on the same thing as well. So I'll close my PR as yours is basically the same, but contains a couple more tests.

sept-en:
Closed in favor of #125.

PR #125, opened 2021-06-03
Eth-connector: fixed u128 json parsing & tests

mrLSD:
**Eth-connector**: fixed u128 json parsing & tests.
* fixed issue at tests: `assert_execution_status_failure` pattern
* fixed all u128 tests for eth-connector
* Added tests:
  * `test_ft_transfer_max_value`
  * `test_ft_transfer_empty_value`


tests/test_connector.rs
```diff
@@ -1151,3 +1151,60 @@ fn assert_execution_status_failure(
...
+        DEFAULT_GAS,
+        1,
+    );
+    let promises = res.promise_results();
+    let promise = &promises[promises.len() - 3];
+    eprintln!("{:#?}", promise.as_ref().unwrap().outcome().clone().status);
+    assert_execution_status_failure(
+        promise.as_ref().unwrap().outcome().clone().status,
+        "ERR_NOT_ENOUGH_BALANCE",
+        "Expected failure as the amount too large",
```
sept-en, line 1180:
```suggestion
        "Expected failure as the amount is too large",
```

tests/test_connector.rs
```diff
@@ -1151,3 +1151,60 @@ fn assert_execution_status_failure(
...
+        1,
+    );
+    let promises = res.promise_results();
+    let promise = &promises[promises.len() - 3];
+    eprintln!("{:#?}", promise.as_ref().unwrap().outcome().clone().status);
+    assert_execution_status_failure(
+        promise.as_ref().unwrap().outcome().clone().status,
+        "ERR_FAILED_PARSE_U128",
+        "Expected failure as empty string can't be parsed to u128",
+    );
```
sept-en, line 1209: As this just checks the string->u128 conversion, I think it's better to have this in the unit test in `src/parameters.rs`. The integration test for this case seems to be redundant.

sept-en (changes requested):

tests/test_connector.rs
mrLSD: But that test covers empty parameter as well.

src/json.rs
```diff
@@ -45,7 +46,7 @@ impl JsonValue {
     pub fn u128(&self, key: &str) -> Result<u128, ()> {
         match self {
             JsonValue::Object(o) => match o.get(key).ok_or(())? {
-                JsonValue::Number(n) => Ok(*n as u128),
+                JsonValue::String(n) => Ok(n.parse::<u128>().expect_utf8(b"ERR_FAILED_PARSE_U128")),
```
sept-en, line 49: I don't think we should panic in this case. It's better to return the error instead.

tests/test_connector.rs
sept-en: The empty `amount` parameter will fail in the unit test as well.

src/json.rs
mrLSD: If we change Err result, it will be logically true to change it for all other `json parser` errors. And also for that, we should have more complicated Error kinds for that. I think it's really redundant for that simple case.

sept-en: We currently don't panic in any other function in `json` module. So if we don't introduce any new Error types in this PR, let's just have `n.parse::<u128>().map_err(|_| ())`. I think it makes no sense just to panic for `u128` and don't panic for all other types.

mrLSD: I disagree. We can check in tests that specific case. This is an exception to the rule.
Again I think that's redundant.

tests/test_connector.rs
mrLSD: But we can test all flow with all arguments. And that part of the integration tests.

birchmd: I think we could have both integration tests and unit test. I think it is generally the case that it is not redundant to have both because they do subtly different things. Unit test ensure a components works correctly in isolation, while integration tests ensure a feature (which may interact with multiple components) works correctly.

The additional advantage of unit tests in this particular case is that they are lighter-weight so we can test more input values in less time (indeed we could even randomly generate valid and invalid strings to try and parse).

src/json.rs
birchmd: Per [the Rust book](https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html), returning a `Result` instead of panicking should be the default choice because it give more control at the call-site on how to handle the error. The only time you `panic!` is when there is no possible way to recover.

In this particular case we are talking about a general JSON parsing function, so making the choice that it is impossible to recover from failed parsing seems wrong to me. I know in the current way it is used this is the case, but what if in the future we want to use this function in another context where it makes to have a fallback value in the case the parsing fails?

tl;dr I agree with @sept-en that we should use `Result` error variant instead of `panic` here.

birchmd: As a side note: why are we removing the `Number` match branch? Couldn't this function return a value if there is a `Number` or if there is a `String` which can be parsed into a number?

mrLSD: Number is an unreachable case.

mrLSD: As I mentioned we should introduce custorm errors for all parser cases if we want return Result with specific error.

joshuajbouw (approved):
Besides what was mentioned by others, this looks good to me.

I do agree, wait for #131 to be merged, then update this accordingly.

artob:
@mrLSD @sept-en @joshuajbouw The JSON custom errors in #131 are now merged, so this needs updating.

src/json.rs
sept-en: Solved after merging #131 in 0dae64c

sept-en (approved):
I updated this to have #131. Seems this ready to merge!
@artob @birchmd @joshuajbouw please review and approve!

src/json.rs
```diff
@@ -54,7 +54,9 @@ impl JsonValue {
     pub fn u128(&self, key: &str) -> Result<u128, JsonError> {
         match self {
             JsonValue::Object(o) => match o.get(key).ok_or(JsonError::MissingValue)? {
-                JsonValue::Number(n) => Ok(*n as u128),
```
birchmd, line 57: It's still seems strange to me that we do not accept numbers; only strings. What would be the issue with accepting both?

birchmd (approved):
Very nice!

src/json.rs
sept-en: I don't think that we need to accept numbers for `u128` as the `Number` is limited to `u64`. So we will have some undefined behaviour and thus it will be hard to track such use cases and get why do we have some errors in future.
When we limit it to strings we allow using only this invariant so we won't have overflow use cases.
Otherwise, we need to remember and choose if we are going to use `Number` or `String` for `u128`, for each use case, etc.

birchmd: There is not undefined behaviour, it's that the map only goes one way. So any `JsonValue::Number` can be converted safely into a `u128`, but not the other way around. But this function is only `JsonValue -> u128`, so supporting the `Number` variant is logically sound.

If someone tried to pass some json like `{ "amount" : 1_000_000_000_000_000_000_000 }` I would expect the error to be in the parsing step because the given number does not fit in a `u64`.

As is, I think people may get confused when then pass json like `{ "amount" : 10 }` and get back `ERR_FAILED_PARSE_U128` when `10` is a valid `u128`.

That said, after laying it out like this it would maybe be a little awkward to accept both numbers and strings for a field because then the schema for the API is a little complex (e.g. `amount: Number | String`). So I would be ok with enforcing the conditions that all numbers which _could_ be bigger than `u64::MAX` must be given as strings regardless of their size. Though in this case, I would say that we should distinguish a type error from a value error. I.e. we should have a separate error variant which can flag to the user they passed a number where they should have passed a string.

sept-en: I think we shouldn't allow using both numbers and strings. Nevertheless, I agree with you that we need to avoid confusion, that's exactly what I meant. Added another `JsonError` and test in 03ad248

mrLSD: @birchmd  If we should put u128 number, it always should be a String. Also, we return u128 as String. So it's a completely consistent approach.

mrLSD:
@birchmd @sept-en I think we should decide do we need more improvements or we can merge current PR.

sept-en:
> @birchmd @sept-en I think we should decide do we need more improvements or we can merge current PR.

@mrLSD I already implemented things that @birchmd asked, so I think it's ready to merge. Waiting for Michael's confirmation :)

birchmd:
Yes, please merge. Sorry for the delay.

Thanks for adding the additional error type @sept-en !

For future reference, if I have approved a PR then for me it means I am ok with it being merged because the remaining comments are minor/non-blocking.

PR #131, opened 2021-06-04
Add JSON custom error types. Refactoring.

sept-en:
* Add `JsonError` for `json` module.
* Use `JsonError` instead `()` for errors in `JsonValue`.
* Don't panic in `json` module on parsing.
* Move `SdkUnwrap`, `SdkExpect` traits from `lib.rs` to `types.rs`.
* Unwrap the error when parsing JSON for parameter args instead of failing with generic `ERR_FAILED_PARSE`.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -6,7 +6,7 @@ use borsh::{BorshDeserialize, BorshSerialize};
 #[cfg(not(feature = "contract"))]
 use sha3::{Digest, Keccak256};
 
-#[cfg(feature = "contract")]
+#[cfg(feature = "engine")]
 use crate::sdk;
 
 pub type AccountId = String;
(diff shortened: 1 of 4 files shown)


birchmd (approved):
LGTM! Nice job!

src/types.rs
```diff
@@ -186,6 +186,56 @@ impl<T, E> ExpectUtf8<T> for core::result::Result<T, E> {
...
+#[cfg(feature = "engine")]
+impl<T, E: AsRef<[u8]>> SdkUnwrap<T> for core::result::Result<T, E> {
+    fn sdk_unwrap(self) -> T {
+        match self {
+            Ok(t) => t,
+            Err(e) => sdk::panic_utf8(e.as_ref()),
+        }
+    }
+}
+
```
joshuajbouw, line 238: To be fair, this should be somewhere else. I wouldn't exactly call them types.

joshuajbouw (changes requested):

src/types.rs
sept-en: Perhaps, yes, I agree!. Please also note we have similar traits in the same `types` file.
So if you don't mind, I suggest merging this PR and leave this part for the next one with some refactoring and structuring changes.

birchmd: I agree with @sept-en that we should not block this PR over this sort of code organization. If it really is `P0` as the tag on the PR says then getting through and fixing minor comments later seems better than bikeshedding on it now.

sept-en: @joshuajbouw @birchmd created issue #134 for that. Please, let's merge this PR ASAP if possible :)

joshuajbouw (approved):
Given comments, yes we can my issue to another PR.

mrLSD (approved):

artob (approved):

PR #133, opened 2021-06-09
EthConnector: don't burn NEP-141 on deposit.

sept-en:
We used to have the following workflow for depositing ETH to Aurora:
* Deposit nETH (NEP-141) tokens to Aurora account
* Burn Aurora's nETH tokens
* Mint ETH to Ethereum address on Aurora.

When we call `exitToNear` precompile for ETH, we expect to burn ETH on Aurora and make `ft_transfer` of nETH tokens from Aurora to the recipient on NEAR. The problem using the previously mentioned deposit workflow is that we don't have nETH tokens available on Aurora account when we try to perform this exit.

Possible solutions:
## 1) Don't burn nETH tokens on Aurora account during the deposit
This solution is currently implemented in the PR. 

Pros: 
* We have needed liquidity and keep calling just a regular  `ft_transfer` when calling `exitToNear` precompile for ETH.
* We have consistency with the approach that we use for ERC20 connector for Aurora (User transfers some NEP-141 tokens to Aurora account and we mint appropriate ERC20 tokens without burning NEP-141)
* We don't need to add any additional functionality to the contract interface and this is quite straightforward.

Cons:
* The balance of `total_balance_near` (which should be renamed to `total_balance_nep141` or something similar) will grow and will represent both nETH and ETH tokens, which might be not the thing we want.
* The balance of `total_balance` will be the sum of `total_balance_near` and `total_balance_eth`. And in this way is not really something meaningful.

## 2) Add needed private functions to Aurora contract
Add the private function `self_mint_and_ft_transfer` (and `self_mint_and_withdraw` for `exitToEthereum`) that will be called only from the appropriate precompile for ETH. The function will first mint the needed amount of nETH to Aurora account and then perform a regular `ft_transfer` to the recipient.

Pros:
* `total_supply_near`, `total_supply_eth`, `total_supply` values remain meaningful and we don't have any redundant balance for nETH that we keep in the contract.
* nETH tokens are minted only when the appropriate precompile is executed and immediately transferred to the recipient.

Cons:
* We have a different approach comparing to that we use when bridging NEP141->ERC20 on Aurora (this was described in the 1st approach).
* We introduce another one point where we mint some tokens -> This is another point for failure and needs a thorough review.

I expect that the 1st approach more likely will be accepted thus I implemented this here. 
However, I'm more into 2nd approach so I think it worth discussing. 

NB: this needs to be solved for both ETH exits: `exitToNear`, `exitToEthereum`.

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -388,6 +388,7 @@ impl EthConnectorContract {
     }
 
     /// Burn NEAR tokens
+    #[allow(dead_code)]
     fn burn_near(&mut self, owner_id: AccountId, amount: Balance) {
         #[cfg(feature = "log")]
         sdk::log(&format!("Burn NEAR {} tokens for: {}", amount, owner_id));
@@ -593,7 +594,6 @@ impl EthConnectorContract {
         // Special case when predecessor_account_id is current_account_id
         if current_account_id == predecessor_account_id {
             let fee = message_data.fee.as_u128();
-            self.burn_near(current_account_id, args.amount);
             // Mint fee to relayer
             let relayer = engine.get_relayer(&message_data.relayer.as_bytes());
             if fee > 0 && relayer.is_some() {
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -387,14 +387,6 @@ impl EthConnectorContract {
         self.ft.internal_deposit_eth(owner_id, amount);
     }
 
-    /// Burn NEAR tokens
-    #[allow(dead_code)]
-    fn burn_near(&mut self, owner_id: AccountId, amount: Balance) {
-        #[cfg(feature = "log")]
-        sdk::log(&format!("Burn NEAR {} tokens for: {}", amount, owner_id));
-        self.ft.internal_withdraw(&owner_id, amount);
-    }
-
     /// Burn ETH tokens
     fn burn_eth(&mut self, address: EthAddress, amount: Balance) {
         #[cfg(feature = "log")]
--- a/tests/test_connector.rs
+++ b/tests/test_connector.rs
@@ -316,13 +316,13 @@ fn test_eth_deposit_balance_total_supply() {
     assert_eq!(balance, DEPOSITED_EVM_FEE);
 
     let balance = total_supply(&master_account, CONTRACT_ACC);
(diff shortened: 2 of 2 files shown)


birchmd:
The cons of the first approach seem to primarily be bookkeeping related. If we correct for the double-counting in `total_supply` could be corrected in any public getters. On the other hand, cons of the second approach seem to be more complexity related and I think it's better to reduce complexity where possible.

mfornet:
Strongly prefer first approach. Being consistent and having a "simple" solution seems a bigger win. The cons are related to the way balances are being displayed, here we have the freedom to update them to show expected values, or educated developers consuming this interface.

mfornet (approved):

src/connector.rs
```diff
@@ -388,6 +388,7 @@ impl EthConnectorContract {
     }
 
     /// Burn NEAR tokens
+    #[allow(dead_code)]
```
birchmd, line 391: If we agree on this solution then I think we should delete the function instead of allowing it to exist as dead code. If we end up needing it in the future we can always restore it from the git history.

birchmd (approved):
One question I have about changing the behaviour here: what happens to any transfers than have already happened? Can they never exit?

mrLSD (approved):
Implemented solution looks clear & simple.

src/connector.rs
sept-en: Done!

joshuajbouw (approved):
Good here. Straight forward.

artob (approved):

artob:
> One question I have about changing the behaviour here: what happens to any transfers than have already happened? Can they never exit?

@sept-en Please do remember to answer @birchmd's question here.

sept-en:
@artob Sorry, forgot to answer it here.

> One question I have about changing the behaviour here: what happens to any transfers than have already happened? Can they never exit?

@birchmd That's correct. This is not an issue for us because ETH connector hasn't been deployed to Mainnet yet, so there we will start from scratch. On Testnet I solved it just by bridging some Ropsten directly to `develop.aurora` account (that I use for testing). For sure on Mainnet this won't be an option, but it's good that we don't have ETH connector here and don't need to implement any specific balance migration there.

PR #135, opened 2021-06-10
Fix needless borrows

birchmd:
Clippy (on latest nightly) pointed out several places where we were borrowing for no reason. This PR could also be called "remove a bunch of useless ampersands from the code" since no functionality is changed.


sept-en (approved):
LGTM!

joshuajbouw (approved):
Sweet, just started getting that myself. Thanks!

Thats so awesome that Clippy is finally doing that.

PR #137, opened 2021-06-10
Contract Migration

mrLSD:
Universal migration mechanism for storage state.
Possible actions:
```
    // Add field with Value
    Add,
    // Rename key field
    RenameKey,
    // Update value for key field
    UpdateValue,
    // Rename key and update value
    UpdateKeyValue,
    // Remove key field
    Remove,
    // Remove only value for key field
    RemoveValue,
```

The migration function can be invoked only by the contract owner (for security reasons).


sept-en:
Nice, thanks for working on this!
Please, make all I/O-related operations in `src/lib.rs` from the very beginning.
This way it will be easier to separate from business logic and we try to follow this pattern everywhere.

src/migration.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+/// Migrate key fields and/or data value
+/// Can be executed only contract itself.
+pub fn migrate() {
+    sdk::assert_private_call();
+    let _args =
+        Migration::from(parse_json(&sdk::read_input()).expect_utf8(ERR_FAILED_PARSE.as_bytes()));
+    sdk::return_output(&"done".as_bytes());
```
sept-en, line 63: Accept `MigrationArgs` instead of parsing here.

```suggestion
pub fn migrate(args: MigrationArgs) {
```

src/lib.rs
```diff
@@ -519,6 +521,11 @@ mod contract {
         EthConnectorContract::get_instance().get_accounts_counter()
     }
 
+    #[no_mangle]
+    pub extern "C" fn migrate() {
+        crate::migration::migrate()
+    }
+
```
sept-en, line 528: Do all I/O-related here, please.

```suggestion
    pub extern "C" fn migrate() {
        sdk::assert_private_call();
        let args = Migration::from(parse_json(&sdk::read_input()).expect_utf8(ERR_FAILED_PARSE.as_bytes()));
        crate::migration::migrate(args);
        sdk::return_output(&"done".as_bytes());
    }
    
```

sept-en (changes requested):

mfornet:
Haven't looked into the PR yet, but have few questions regarding the options:

```rust
/// Rename key field
RenameKey,
// Rename key and update value
UpdateKeyValue,
````

What do you mean by rename key? Borsh serialisation doesn't save the name of the fields on the storage. Or you meant something like. Moving `value` stored at `key_1` to `key_2` (and deleting `key_1`)?

----

```rust
// Remove only value for key field
RemoveValue,
```

What does it do, replace this value with empty sequence of bytes?

----

I don't know if I'm missing a conversation from some place. But can you provide here high level details  (or even technical if necessary) about how are we planning to execute one migration using this changes, before jumping into the implementation.

mrLSD:
> Haven't looked into the PR yet, but have few questions regarding the options:
> 
> ```rust
> /// Rename key field
> RenameKey,
> // Rename key and update value
> UpdateKeyValue,
> ```
> 
> What do you mean by rename key? Borsh serialisation doesn't save the name of the fields on the storage. Or you meant something like. Moving `value` stored at `key_1` to `key_2` (and deleting `key_1`)?
It doesn't matter. We can set any kind of key. So we can rename storage key (and don't touch value) and some other useful cases.

> ```rust
> // Remove only value for key field
> RemoveValue,
> ```
> 
> What does it do, replace this value with empty sequence of bytes?
yes

> I don't know if I'm missing a conversation from some place. But can you provide here high level details (or even technical if necessary) about how are we planning to execute one migration using this changes, before jumping into the implementation.

@sept-en asks me to implement a migration mechanism for changing the storage state.

That logic currently covering all possible cases. Parameters set via json.

mrLSD:
@birchmd Fill free for suggestions and ideas.
The short story is: @sept-en asks me about how to do the migration.
I proposed an idea and started implementation.

birchmd:
This PR feels more like a design proposal and so I think it should be discussed separately from any actual implementation. Hopefully if we all agree on a design first then we can avoid churn on code itself which will save us all time in the long run.

I have enabled GitHub discussions for the Engine repo and created a section called `Design` where we can talk about designs for new features/functionality.

The discussion for this migration functionality is here: https://github.com/aurora-is-near/aurora-engine/discussions/140
Please feel free to add more details there (e.g. @sept-en if this is something you are pushing for please comment with some pros for having a migration DSL, as opposed to migrations in code itself).

As a general note I think we should be spending more time on design than we currently are. Spending some time on design up-front can hopefully prevent sudden pivots like we have seen recently (e.g. don't burn NEP-141 tokens in connector, generational storage, tracking promise creation in exit precompiles). I know that each of those examples were in response to problems that arose with the implementation, and that a design phase will not always catch all such problems, but at least it gives us a better chance of doing so.

mfornet:
> @sept-en asks me about how to do the migration. I proposed an idea and started implementation.

@mrLSD Where did you propose this idea, I can't find it anywhere, and the description of the PR doesn't help to it. 

> That logic currently covering all possible cases. Parameters set via json.

You didn't address the rest of my comments. So you are covering cases that are not described properly, and it is not straight forward to understand how they are going to be used.

Let's continue this conversation in https://github.com/aurora-is-near/aurora-engine/discussions/140 (thanks @birchmd for starting this discussion).

joshuajbouw:
This PR is stale, whats the status @mrLSD ?

mrLSD:
> This PR is stale, whats the status @mrLSD ?

Paused. See #140

PR #138, opened 2021-06-11
Add EIP-2565 support

joshuajbouw:
This is EIP 1 out of 4 for the Berlin hardfork.

For this one, it's the `modexp` gas cost changes. For more details, see the linked reference below.

Reference: https://eips.ethereum.org/EIPS/eip-2565


birchmd (approved):
LGTM Nice work!

src/precompiles/modexp.rs
```diff
@@ -3,87 +3,46 @@ use crate::precompiles::{
...
 pub(super) struct ModExp<HF: HardFork>(PhantomData<HF>);
 
-impl ModExp<Byzantium> {
-    fn adj_exp_len(exp_len: U256, base_len: U256, bytes: &[u8]) -> U256 {
-        let mut exp32_bytes = Vec::with_capacity(32);
-        for i in 0..32 {
+impl<HF: HardFork> ModExp<HF> {
+    fn calc_iter_count(exp_len: u64, base_len: u64, bytes: &[u8]) -> u64 {
+        let mut exp_bytes = Vec::with_capacity(32);
+        for i in 0..exp_len {
```
birchmd, line 15: The whole point of this `for` loop is to extract the exponent from the input data right? Maybe worth extracting this logic into a separate function so that we only need to parse the exponent once (note we need the exponent again in `run_inner`).

src/precompiles/modexp.rs
```diff
@@ -3,87 +3,46 @@ use crate::precompiles::{
...
             if U256::from(96) + base_len + U256::from(1) >= U256::from(bytes.len()) {
```
birchmd, line 16: Now that `base_len` is a `u64` probably don't need to convert all the other stuff in to `U56`

src/precompiles/modexp.rs
```diff
@@ -132,24 +89,96 @@ impl Precompile for ModExp<Byzantium> {
...
-        todo!()
+    fn required_gas(input: &[u8]) -> Result<u64, ExitError> {
+        let base_len = U256::from(&input[0..32]).as_u64();
+        let exp_len = U256::from(&input[32..64]).as_u64();
+        let mod_len = U256::from(&input[64..96]).as_u64();
+
+        let mul = Self::mul_complexity(base_len, mod_len);
+        let iter_count = Self::calc_iter_count(exp_len, base_len, &input) - 1;
+
+        Ok(core::cmp::max(200, mul * iter_count / 3))
```
birchmd, line 161: There is a subtle deviation from the spec here. It says `math.floor(multiplication_complexity * iteration_count / 3)`, which in general is not equal to `multiplication_complexity * math.floor(iteration+count / 3)` (what you have here).

I think it should be `(mul * iter_count) / 3`

birchmd (changes requested):
Actually, sorry, I found a small mistake, undoing my approve for now

src/precompiles/modexp.rs
birchmd: Actually, my bad 🤦 I had added brackets in my head which weren't really there. It's correct as is.

birchmd (approved):
Back to approval, sorry!

src/precompiles/modexp.rs
```diff
@@ -132,24 +89,96 @@ impl Precompile for ModExp<Byzantium> {
...
+}
+
+impl Precompile for ModExp<Byzantium> {
+    fn required_gas(input: &[u8]) -> Result<u64, ExitError> {
+        let base_len = U256::from(&input[0..32]).as_u64();
+        let exp_len = U256::from(&input[32..64]).as_u64();
+        let mod_len = U256::from(&input[64..96]).as_u64();
+
+        let mul = Self::mul_complexity(core::cmp::max(mod_len, base_len))?;
+        let iter_count = Self::calc_iter_count(exp_len, base_len, &input) - 1;
```
birchmd, line 120: Sorry, I'm all over the place today :(

Does this underflow in the case the exponent is zero?

birchmd (changes requested):

src/precompiles/modexp.rs
```diff
@@ -132,24 +89,96 @@ impl Precompile for ModExp<Byzantium> {
...
+}
+
+impl ModExp<Byzantium> {
+    fn mul_complexity(x: u64) -> Result<u64, ExitError> {
+        if x <= 64 {
+            Ok(x * x)
+        } else if x <= 1_024 {
+            Ok(x * x / 4 + 96 * x - 3_072)
+        } else {
+            let (sqroot, overflow) = x.overflowing_mul(x);
```
birchmd, line 103: It's not a square root, it's a square. Maybe `x_sq` would be a better name.

src/precompiles/modexp.rs
joshuajbouw: Good catch! Other implementations just subtracted the 1. So, hmm. Good question. I added all test case though. I think one of them is 0 exponent.

birchmd: The minus 1 is only in the non-zero cases in the spec. Also in the last case, the minus 1 is only on the bit length, not the whole expression.

```
def calculate_iteration_count(exponent_length, exponent):
    iteration_count = 0
    if exponent_length <= 32 and exponent == 0: iteration_count = 0
    elif exponent_length <= 32: iteration_count = exponent.bit_length() - 1
    elif exponent_length > 32: iteration_count = (8 * (exponent_length - 32)) + ((exponent & (2**256 - 1)).bit_length() - 1)
    return max(iteration_count, 1)
```

src/precompiles/modexp.rs
joshuajbouw: Honestly, I'm not even quite too happy with how it is. I'm going to revisit this.

joshuajbouw: This was resolved by using slices. I'm not sure why I hadn't just done this in the first place. Might've been a reason before, but whatever it was, no longer is there.

src/precompiles/modexp.rs
joshuajbouw: Irrelevant now with the changes to this as commented by https://github.com/aurora-is-near/aurora-engine/pull/138/commits/1f49da17bb2c6e660e191e85e157a8794b04111a.

src/precompiles/modexp.rs
joshuajbouw: Ah yes, good catch. Fixed now in https://github.com/aurora-is-near/aurora-engine/pull/138/commits/0494f937a2f5cd930846ecdca3278e6a64a14b05.

src/precompiles/modexp.rs
```diff
@@ -3,107 +3,41 @@ use crate::precompiles::{
...
-                    // Pad out the data if the byte is empty.
-                    exp32_bytes.push(0u8);
-                }
-            }
-        }
-        let exp32 = U256::from(exp32_bytes.as_slice());
+impl<HF: HardFork> ModExp<HF> {
+    fn calc_iter_count(exp_len: u64, base_len: u64, bytes: &[u8]) -> u64 {
+        let start = base_len as usize + 96;
+        let end = exp_len as usize + start;
```
birchmd, line 15:
```suggestion
        let end = cmp::min(32, exp_len as usize) + start;
```

We only want the first 32 bytes of the exponent for this function.

src/precompiles/modexp.rs
```diff
@@ -3,107 +3,41 @@ use crate::precompiles::{
...
-        } else if x <= U256::from(1_024) {
-            Ok(x * x / U256::from(4) + U256::from(96) * x - U256::from(3_072))
-        } else {
-            let (sqroot, overflow) = x.overflowing_mul(x);
-            if overflow {
-                Err(ExitError::OutOfGas)
-            } else {
-                Ok(sqroot / U256::from(16) + U256::from(480) * x - U256::from(199_680))
-            }
+            8 * (exp_len - 32) + exp.bits() as u64
```
birchmd, line 24:
```suggestion
            8 * (exp_len - 32) + exp.bits() as u64 - 1
```

src/precompiles/modexp.rs
```diff
@@ -132,24 +66,96 @@ impl Precompile for ModExp<Byzantium> {
...
+impl ModExp<Byzantium> {
+    fn mul_complexity(x: u64) -> Result<u64, ExitError> {
+        if x <= 64 {
+            Ok(x * x)
+        } else if x <= 1_024 {
+            Ok(x * x / 4 + 96 * x - 3_072)
+        } else {
+            let (x_sq, overflow) = x.overflowing_mul(x);
+            if overflow {
+                Err(ExitError::OutOfGas)
```
birchmd, line 82: In [the Go implementation](https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go#L295) they use big ints all the way through and then return a `u64` at the end (returning `u64::MAX` if the result would be larger than a `u64` can represent). Perhaps we should do the same.

src/precompiles/modexp.rs
```diff
@@ -132,24 +66,96 @@ impl Precompile for ModExp<Byzantium> {
...
+            Ok(PrecompileOutput::without_logs(cost, output))
+        }
+    }
+}
+
+impl ModExp<Berlin> {
+    fn mul_complexity(base_len: u64, mod_len: u64) -> u64 {
+        let max_len = core::cmp::max(mod_len, base_len);
+        let words = max_len.div_ceil(&8);
+        words.pow(2)
```
birchmd, line 125:
```suggestion
        words.saturating_mul(words)
```

Can use saturating multiplication to avoid overflow.

birchmd (changes requested):
Thanks for pushing the new changes! There are still a couple issues with the implementation I think. Maybe we need tests that use really big values for the inputs (bigger than could fit in a `U256`) because I think the issues would only appear from very large inputs.

src/precompiles/modexp.rs
```diff
@@ -132,24 +66,96 @@ impl Precompile for ModExp<Byzantium> {
...
-        todo!()
+    fn required_gas(input: &[u8]) -> Result<u64, ExitError> {
+        let base_len = U256::from(&input[0..32]).as_u64();
+        let exp_len = U256::from(&input[32..64]).as_u64();
+        let mod_len = U256::from(&input[64..96]).as_u64();
+
+        let mul = Self::mul_complexity(base_len, mod_len);
+        let iter_count = Self::calc_iter_count(exp_len, base_len, &input);
+
+        Ok(core::cmp::max(200, mul * iter_count / 3))
```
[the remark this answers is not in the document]
joshuajbouw: I had originally had done that, thinking that the gas wouldn't get that high. Double checked, and it could potentially. Will change it back.

src/precompiles/modexp.rs
joshuajbouw: That is what I had originally done, not sure why I opted to remove this in the end.

joshuajbouw:
@birchmd I'll finish the rest up once I get back from some sleep. Exhausted right now. Thanks for the review.

src/precompiles/modexp.rs
```diff
@@ -3,107 +3,41 @@ use crate::precompiles::{
...
-            if 96 + base_len + i >= input.len() {
-                exp_bytes.push(0u8);
-            } else {
-                exp_bytes.push(input[96 + base_len + i]);
-            }
-        }
-
-        let mod_len = mod_len.as_usize();
         let mut mod_bytes = Vec::with_capacity(32);
         for i in 0..mod_len {
```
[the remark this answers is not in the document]
joshuajbouw: Fixed in https://github.com/aurora-is-near/aurora-engine/pull/138/commits/367c8401307f4eb3fb3f918a51c8b6a086e61d0c

src/precompiles/modexp.rs
joshuajbouw: Good catch on that one. None of the tests hit it, interestingly enough.

src/precompiles/modexp.rs
joshuajbouw: Done here https://github.com/aurora-is-near/aurora-engine/pull/138/commits/29e4336bacd886b3818c6ba6f0be730e16e242f6

src/precompiles/modexp.rs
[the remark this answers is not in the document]
joshuajbouw: Done here. https://github.com/aurora-is-near/aurora-engine/pull/138/commits/29e4336bacd886b3818c6ba6f0be730e16e242f6

src/precompiles/modexp.rs
```diff
@@ -3,52 +3,80 @@ use crate::precompiles::{
...
-        }
-        let exp32 = U256::from(exp32_bytes.as_slice());
+impl<HF: HardFork> ModExp<HF> {
+    fn calc_iter_count(exp_len: U256, base_len: U256, bytes: &[u8]) -> U256 {
+        let start = base_len.as_usize() + 96;
+        let end = core::cmp::min(32, exp_len.as_usize()) + start;
+        let exp = U256::from(&bytes[start..end]);
 
-        if exp_len <= U256::from(32) && exp32 == U256::zero() {
+        if exp_len <= U256::from(32) && exp == U256::zero() {
```
birchmd, line 18:
```suggestion
        if exp_len <= U256::from(32) && exp.is_zero() {
```

src/precompiles/modexp.rs
```diff
@@ -3,52 +3,80 @@ use crate::precompiles::{
...
+
+impl ModExp<Byzantium> {
+    fn mul_complexity(x: U256) -> Result<U256, ExitError> {
         if x <= U256::from(64) {
             Ok(x * x)
         } else if x <= U256::from(1_024) {
             Ok(x * x / U256::from(4) + U256::from(96) * x - U256::from(3_072))
         } else {
-            let (sqroot, overflow) = x.overflowing_mul(x);
+            let (x_sq, overflow) = x.overflowing_mul(x);
```
birchmd, line 75: Overflow check is no longer necessary because the max value of `x` is `u32::MAX` and `(2^32)^2 = 2^64 < 2^256`.

src/precompiles/modexp.rs
```diff
@@ -60,96 +88,66 @@ impl Precompile for ModExp<Byzantium> {
...
-            / U256::from(20);
-        let (gas_val, overflow) = mul.overflowing_mul(adj);
-        if overflow {
-            Err(ExitError::OutOfGas)
-        } else {
-            Ok(gas_val.as_u64())
-        }
+        let mul = Self::mul_complexity(core::cmp::max(mod_len, base_len))?;
+        let iter_count = Self::calc_iter_count(exp_len, base_len, &input);
+        let maybe_gas = mul.checked_mul(core::cmp::max(iter_count, U256::from(1)));
```
birchmd, line 93: Overflow check not necessary since `mul` is bounded by `2^61` and `iter_count` is bounded by `2^40`, so the whole result is bounded by `2^101 < 2^256`.

src/precompiles/modexp.rs
```diff
@@ -60,96 +88,66 @@ impl Precompile for ModExp<Byzantium> {
...
+
+        maybe_gas
+            .map(|g| g.as_u64() / 20)
```
birchmd, line 96: This cast will panic if `g > u64::MAX`. We need to do something more explicit like

```rust
let gas = g / 20;
if gas > u64::MAX { u64::MAX } else { gas.as_u64() }
```

src/precompiles/modexp.rs
```diff
@@ -60,96 +88,66 @@ impl Precompile for ModExp<Byzantium> {
...
-                padded_result.extend(core::iter::repeat(0).take(diff));
-                padded_result.extend_from_slice(&computed_result);
-                padded_result
-            } else {
-                computed_result
-            }
-        };
-
-        Ok(PrecompileOutput::without_logs(cost, output))
+        Ok(core::cmp::max(200, mul * iter_count / 3))
```
birchmd, line 130: This arithmetic should be done in `U256` to prevent overflow (`mul` is bounded by `2^58` and `iter_count` is bounded by `2^40` so the whole expression is bounded by `2^98` which is too big for `u64`, but fine for `U256`). Then at the end we can round down to `u64`.

birchmd (changes requested):

birchmd (approved):
I made several changes around overflow handling as well wrote some functions for dealing with the slices to avoid index out of bounds panics (this behaviour is the same as geth). I also added some tests to make sure it all works. Should be good to go now.

joshuajbouw:
Thanks for the clean up.

artob:
@joshuajbouw Please fix the merge conflicts.

artob (approved):

PR #141, opened 2021-06-12
JSON: bugs fixes, add unit tests.

sept-en:


src/json.rs
```diff
@@ -196,3 +214,148 @@ pub fn parse_json(data: &[u8]) -> Option<JsonValue> {
...
+        let val = JsonValue::parse_u8(&json).ok().unwrap();
+        assert_eq!(val, 255);
+
+        let json = JsonValue::from(260f64);
+        let val = JsonValue::parse_u8(&json).ok().unwrap();
+        assert_eq!(val, 255);
+
+        let json = JsonValue::from(12345678f64);
+        let val = JsonValue::parse_u8(&json).ok().unwrap();
+        assert_eq!(val, 255);
```
sept-en, line 350: Probably this is wrong that values over `U8::MAX` are allowed for parsing and return 255 by default.
I think this should be treated as an error, so I will fix this.

sept-en: Fixed out-of-range allowed behaviour for `u8` and `u64` in 973c225.

mrLSD:
@sept-en what current status of PR?

joshuajbouw:
@sept-en This is getting stale, whats the status?

sept-en:
@mrLSD @joshuajbouw this was set as a draft because of dependency on #137. Now it needs to be finished without it as the mentioned PR won't be there in the nearest future. The current PR is a low priority though I will try to resolve it this week.
Thanks for reminding me!

mrLSD:
@sept-en I completed that PR. I can't request review from you. So just put your feedback.

src/json.rs
```diff
@@ -12,7 +16,7 @@ pub enum JsonValue {
     Object(BTreeMap<String, JsonValue>),
 }
 
-#[derive(Debug)]
+#[cfg_attr(test, derive(PartialEq))]
```
joshuajbouw, line 19: I would add more than just PartialEq. Also, does it only need to be for test? If it isn't used, then it'll not even be compiled so it is fine to add more than enough here than just what is needed. Typically for enums I would add like (off the top of my head here) `Debug, Ord, PartialOrd, Eq, PartialEq` which should cover all that we would use.

So, drop `cfg_attr`, add more derives

src/json.rs
```diff
@@ -23,6 +27,9 @@ pub enum JsonError {
     InvalidString,
     InvalidArray,
     ExpectedStringGotNumber,
+    OutOfRangeU8,
+    OutOfRangeU64,
+    OutOfRangeU128,
```
joshuajbouw, line 32: These should be condensed to be the same error `OutOfRange`.

src/json.rs
```diff
@@ -78,19 +92,31 @@ impl JsonValue {
...
             _ => Err(JsonError::InvalidU8),
         }
     }
 
     #[allow(dead_code)]
-    pub fn array<T, F>(&self, key: &str, call: F) -> Result<Vec<T>, JsonError>
+    pub fn array<T, F>(&self, key: &str, mut call: F) -> Result<Vec<T>, JsonError>
     where
-        F: FnMut(&JsonValue) -> T,
+        F: FnMut(&JsonValue) -> Result<T, JsonError>,
```
joshuajbouw, line 109: It feels "off" to me to take an input as a Result here. I do not recommend it.

src/lib.rs
```diff
@@ -323,7 +323,7 @@ mod contract {
         let args: NEP141FtOnTransferArgs = parse_json(sdk::read_input().as_slice())
             .sdk_unwrap()
             .try_into()
-            .map_err(|err| format!("ERR_JSON_{:?}", err))
+            .map_err(|err| err)
```
joshuajbouw, line 326: This is unnecessary as you are doing nothing to `err`.

joshuajbouw (changes requested):

src/lib.rs
mrLSD: I just generalized that part of the code. Also what you mean "doing nothing to err"?  The next line unwrapped it.

src/json.rs
mrLSD: Do you have another approach to catch errors from input functions with Result types?

src/json.rs
mrLSD: We should finalize Error details for aurora-engine. Someone wants more general errors, someone more detailed.
Actually, I think more detailed errors is a good way for an environment where impossible to debug.

src/json.rs
mrLSD: Redundant derives IMHO is a bad style. Also for the release version, it's completely wrong.

src/lib.rs
joshuajbouw: I mean, the `map_err` is unnecessary.

src/json.rs
joshuajbouw: I mean, you could have another `Enum` of each kind if you want.

src/json.rs
joshuajbouw: I don't necessarily agree as Rust compiler is smart enough to know which to compile. It won't add to the byte size of the resulting WASM.

src/json.rs
joshuajbouw: Well, do something with it before its passed to a method

src/lib.rs
birchmd: @joshuajbouw means this closure is the identity function, so it seems like it should not be needed. What happens if you simply remove this line?

src/json.rs
birchmd: If this function were called `parse_array` I would be more on-board with this function signature. Parsing usually involves an error case.

I would expect a function simply called `array` to look something like `fn array(&self, key: &str) -> Result<Vec<JsonValue>, JsonError>`.

src/json.rs
mrLSD: @sept-en that was from you. What do you think?

src/json.rs
birchmd: I think it makes sense to follow conventions of other projects. And in my experience usually errors are grouped together as underflow and overflow (regardless of container type size).

That said, I agree more granular errors make debugging much easier when all we get usually is the error message. So I'm fine either way.

birchmd (changes requested):
Thank you for adding lots of new tests! I greatly appreciate it ❤️

src/json.rs
joshuajbouw: But generally, these are all related to each other. So just add a field of which the max value is. That's pretty typical.

src/lib.rs
joshuajbouw: @birchmd Will be the same. You just are returning the same value in the `.map`. Removing `.map_err` yields the same result in this regard.

src/json.rs
```diff
@@ -49,7 +56,14 @@ impl JsonValue {
     pub fn u64(&self, key: &str) -> Result<u64, JsonError> {
         match self {
             JsonValue::Object(o) => match o.get(key).ok_or(JsonError::MissingValue)? {
-                JsonValue::Number(n) => Ok(*n as u64),
+                JsonValue::Number(n) => {
+                    // Upper bound is covered by the type limitation
```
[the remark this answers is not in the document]
joshuajbouw: Good catch

joshuajbouw:
In the future @sept-en please link other PRs and say that it is dependent on another else I wouldn't know without talking about it. Please fill out a description too, regardless of how self-explanatory.

src/lib.rs
birchmd: I know. I was trying to rephrase what you said for @mrLSD 's benefit.

src/json.rs
mrLSD: @joshuajbouw I extended OutOfRange to separated kind of errors

src/json.rs
mrLSD: Actually we don't use for release that function. So @birchmd I will try your solution.

src/json.rs
```diff
@@ -207,3 +261,216 @@ pub fn parse_json(data: &[u8]) -> Option<JsonValue> {
...
+        assert_eq!(err, JsonError::NotJsonType);
+    }
+
+    #[test]
+    fn test_json_type_u64() {
+        let json = parse_json(r#"{"foo": 123}"#.as_bytes()).unwrap();
+        let val = json.u64("foo").ok().unwrap();
+        assert_eq!(val, 123);
+
+        let json = parse_json(r#"{"foo": -123}"#.as_bytes()).unwrap();
```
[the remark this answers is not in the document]
mrLSD: Good point, thanks! I'll cover.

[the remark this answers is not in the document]
mrLSD: I tested and we have a problem. `rjson` initial implementation for f64, so we should implement some new functionality for that.

mrLSD:
@birchmd we should decide do we really need to implement new functionality for u64, u128. `rjson` parser takes the initial number only as f64.

src/json.rs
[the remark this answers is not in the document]
birchmd: Casting to `i128` does not fix the problem of the upper bound. I think the cast will silently succeed in a "saturating" kind of way, but that is not what we want.

I do not think there is a good way around including a runtime check that `n <= u64::MAX`.

birchmd:
> rjson parser takes the initial number only as f64.

That is unfortunate; going through `f64` has a possibility of some precision issue I think. We could test this by passing `u64::MAX`. If there is a precision problem with this (i.e. the value after parsing is not equal to the input value due to rounding in floating point), then maybe we need to consider forking `rjson`, or contributing upstream, or just writing our own parser to suit our needs.

None of those options is too much work I don't think, but probably would be outside the scope of this PR. We could for example, file an issue for unresolved items and merge this since it is an improvement even if it is not perfect.

joshuajbouw:
@mrLSD Why not use Serde with the JSON plugin?

mrLSD:
> @mrLSD Why not use Serde with the JSON plugin?

It'll be the ideal solution. But AFAIR it increases contract size.

src/json.rs
mrLSD: @birchmd I propose to delete that method. We do not use it.
Also implementation for `Vec<JsonValue>` has pitfalls.

birchmd: Works for me.

birchmd:
> Why not use Serde with the JSON plugin?

I think we were doing that previously and someone suggested we go for something more minimal because `serde` is very heavy-weight for our use case.

We could always measure what the impact is on code size (we should eliminate the dependency on secp256k1 first though because we already know the vast majority of the current contract size comes from that).

joshuajbouw:
> > Why not use Serde with the JSON plugin?
> 
> I think we were doing that previously and someone suggested we go for something more minimal because `serde` is very heavy-weight for our use case.
> 
> We could always measure what the impact is on code size (we should eliminate the dependency on secp256k1 first though because we already know the vast majority of the current contract size comes from that).

I see, good to know.

src/json.rs
```diff
@@ -207,3 +252,215 @@ pub fn parse_json(data: &[u8]) -> Option<JsonValue> {
...
+        let err = json.u64("foo").unwrap_err();
+        assert_eq!(
+            err,
+            JsonError::OutOfRange(JsonOutOfRangeError::OutOfRangeU64)
+        );
+
+        // let json = parse_json(r#"{"foo": 12.3}"#.as_bytes()).unwrap();
+        // let err = json.u64("foo").unwrap_err();
+        // assert_eq!(err, JsonError::InvalidU64);
```
birchmd, line 314: Please to do not leave commented out code

src/json.rs
```diff
@@ -207,3 +252,215 @@ pub fn parse_json(data: &[u8]) -> Option<JsonValue> {
...
+
+    #[test]
+    fn test_json_type_u64() {
+        let json = parse_json(r#"{"foo": 123}"#.as_bytes()).unwrap();
+        let val = json.u64("foo").ok().unwrap();
+        assert_eq!(val, 123);
+
+        let json = parse_json(r#"{"foo": 12.99}"#.as_bytes()).unwrap();
+        let val = json.u64("foo").ok().unwrap();
+        assert_eq!(val, 12);
```
birchmd, line 299: It's a little awkward to me that this is the result. I think it would be nice to file a new GH issue for this and mark it as a TODO

```
let val = json.u64("foo").ok().unwrap(); // TODO(<GH issue number here>): should fail since it is not a `u64`
```

src/json.rs
```diff
@@ -207,3 +252,215 @@ pub fn parse_json(data: &[u8]) -> Option<JsonValue> {
...
+
+        let json = parse_json(format!(r#"{{"foo": {} }}"#, u64::MAX).as_bytes()).unwrap();
+        let val = json.u64("foo").ok().unwrap();
+        assert_eq!(val, u64::MAX);
```
birchmd, line 303: Thanks for adding this!

Please also add a test showing what happens if you put in a number larger than `u64::MAX`.

src/json.rs
```diff
@@ -49,7 +59,14 @@ impl JsonValue {
     pub fn u64(&self, key: &str) -> Result<u64, JsonError> {
         match self {
             JsonValue::Object(o) => match o.get(key).ok_or(JsonError::MissingValue)? {
-                JsonValue::Number(n) => Ok(*n as u64),
+                JsonValue::Number(n) => {
+                    // Upper bound is covered by the type limitation
+                    if (*n as i128) < u64::MIN as i128 {
```
joshuajbouw, line 64: Why is this conversion `i128`? Shouldn't this just be `u64` ...?

src/json.rs
```diff
@@ -78,24 +95,16 @@ impl JsonValue {
     #[allow(dead_code)]
     pub fn parse_u8(v: &JsonValue) -> Result<u8, JsonError> {
         match v {
-            JsonValue::Number(n) => Ok(*n as u8),
+            JsonValue::Number(n) => {
+                if (*n as i128) < u8::MIN as i128 || (*n as i128) > u8::MAX as i128 {
```
joshuajbouw, line 99: Same as above, regarding `u64` over `i128`

joshuajbouw (changes requested):
Just small ask here.

src/json.rs
mrLSD: It's overflow check

joshuajbouw: I get that, but why would you check it `as i128`

mrLSD (approved):

joshuajbouw (approved):

PR #143, opened 2021-06-17
Release 1.3.0

artob:


etc/eth-contracts/contracts/AdminControlled.sol
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: CC0 1.0 Universal
```
artob, line 1: @mfornet @sept-en This is incorrect, BTW. It should be just `SPDX-License-Identifier: CC0-1.0`.

artob: Fixed in 3b6bc3753616c65bd5160e4d88bec1c818b5c275.

src/meta_parsing.rs
```diff
@@ -140,7 +140,7 @@ pub fn parse_type(field_type: &str) -> ParsingResult<ArgType> {
 pub fn near_erc712_domain(chain_id: U256) -> RawU256 {
     let mut bytes = Vec::with_capacity(70);
     bytes.extend_from_slice(
-        &keccak("EIP712Domain(string name,string version,uint256 chainId)".as_bytes()).as_bytes(),
+        keccak("EIP712Domain(string name,string version,uint256 chainId)".as_bytes()).as_bytes(),
```
artob, line 143: This looks computed at runtime. Surely we could compute it at compile time?

sept-en (approved):
LGTM!

PR #150, opened 2021-06-17
Fix warnings

sept-en:
The purpose of PR is to fix all warnings we currently have.
For some constants it was weird to add an unnecessary `engine` feature check, so I decided to manually turn off the warning by using the following approach:
```rust
#[cfg_attr(not(feature = "engine"), allow(dead_code))]
```

### DIFF
--- a/src/precompiles/bn128.rs
+++ b/src/precompiles/bn128.rs
@@ -48,8 +48,11 @@ mod consts {
 pub(super) mod addresses {
     use crate::precompiles;
 
+    #[cfg_attr(not(feature = "engine"), allow(dead_code))]
     pub const ADD: [u8; 20] = precompiles::make_address(0, 6);
+    #[cfg_attr(not(feature = "engine"), allow(dead_code))]
     pub const MUL: [u8; 20] = precompiles::make_address(0, 7);
+    #[cfg_attr(not(feature = "engine"), allow(dead_code))]
     pub const PAIR: [u8; 20] = precompiles::make_address(0, 8);
 }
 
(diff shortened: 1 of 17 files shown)


birchmd:
Still some warnings showing up in the GH action (appears in `Files changed` tab). Please also fix those in this PR

sept-en:
@birchmd thanks! I fixed those as well. This happens because we are using different toolchains for clippy-checks locally in `Makefile` and in GH workflows. Need to address this thing later as well.

src/precompiles/bn128.rs
```diff
@@ -48,8 +48,11 @@ mod consts {
 pub(super) mod addresses {
     use crate::precompiles;
 
+    #[cfg_attr(not(feature = "engine"), allow(dead_code))]
```
birchmd, line 51: Like @mfornet mentioned, we really need to clean up this `engine` feature. What value is it adding?

birchmd (approved):
Thanks for this!

src/precompiles/bn128.rs
sept-en: Yes, I think that this initially appeared for some good purposes, but not it's quite confusing, as now we have 74 usages of it in our code. And this is even after I minimized this a little bit with having one statement for several includes in the file. Let's address this in future PRs though.

artob (approved):

src/precompiles/bn128.rs
birchmd: #154 removes the engine flag

PR #153, opened 2021-06-18
Fix debug CI pipeline

sept-en:


artob (approved):

artob:
Ah, it was a race condition, I see :+1:

PR #154, opened 2021-06-18
Remove engine feature flag

birchmd:
As @mfornet pointed out previously the `engine` feature flag doe not seem to add any value, and in fact is a determent to testing because it hides so much logic.

This PR simply removes it.


sept-en (approved):
Nice, thanks for doing this!

mrLSD (approved):

joshuajbouw (approved):
Straight forward.

PR #155, opened 2021-06-18
Refactor: Move tests to `src/tests`

birchmd:
Per [matklad's blog post](https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html), it is better in our situation to have all our tests be part of the same compilation artifact (as opposed to rust's "integration tests" which are all compiled separately).

In this PR we resolve the conflicting versions of wasmtime which was preventing us from having all our tests in one place. It also moves most of the tests from `tests` to `src/tests` as suggested in the blog post. An additional benefit of moving the tests under the main source directory is being able to easily reuse code in tests (for example I could delete the `AuroraConfig` struct from the state migration test because it duplicated some things from `test_utils`).

The only tests that remain in `tests` are the ones for the connector. These in particular have a lot of code duplication from production which will be able to be eliminated once it is brought under `src/tests`. However, we are not doing that in this PR because it needs #154 to be merged first. So it will be done as a follow up PR.

One other thing I noticed is that the new `meta-call` feature meant we were not running the tests for meta parsing any more. This is also corrected in this PR.


mrLSD (approved):

tests/test_upgrade.rs
```diff
@@ -1,55 +0,0 @@
...
-    master_account
-        .call(
-            accounts(0).to_string(),
-            "deploy_upgrade",
-            &[],
-            DEFAULT_GAS,
-            0,
-        )
-        .assert_success();
-}
```
sept-en, line 55: Why we completely remove this?

sept-en (approved):
Good job! This should really speed up the CI process.

Added some small question there, check it, please.

tests/test_upgrade.rs
birchmd: This test doesn't add any value over the state migration one. They call the same contract methods but the state migration one actually checks the new contract was deployed by calling a method that's didn't exist previously.

mfornet (approved):
Love those PR when the amount of LOC removed are 8x the amount of LOC added!

src/test_utils/mod.rs
```diff
@@ -103,7 +103,8 @@ impl<'a> OneShotAuroraRunner<'a> {
         );
 
         near_vm_runner::run(
-            &self.base.code,
+            self.base.code.hash.as_ref().to_vec(),
+            &self.base.code.code.as_slice(),
```
joshuajbouw, line 107: `base.code.code` seems a bit... confusing :P

joshuajbouw (approved):
Besides nit, looks good!

src/test_utils/mod.rs
birchmd: Yes, it was nicer before for sure. This became necessary because I needed to use an older version of `near-vm-logic` so that it matches the version used by `near-sdk-sim`.

PR #158, opened 2021-06-22
Fix NEP-141 Compliance

mfornet:
Return number as json (string) in ft_on_transfer

### DIFF
--- a/src/fungible_token.rs
+++ b/src/fungible_token.rs
@@ -1,3 +1,4 @@
+use crate::json::parse_json;
 #[cfg(feature = "log")]
 use crate::prelude::format;
 use crate::types::*;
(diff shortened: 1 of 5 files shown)


sept-en (approved):
Good catch! LGTM!

Please fix tests and let's merge this ASAP.

src/fungible_token.rs
```diff
@@ -245,12 +246,10 @@ impl FungibleToken {
...
             PromiseResult::Successful(value) => {
-                if let Ok(unused_amount) = String::from_utf8(value) {
-                    let unused_amount = if let Ok(v) = unused_amount.parse::<u128>() {
-                        v
-                    } else {
-                        amount
-                    };
+                if let Some(unused_amount) = parse_json(value.as_slice())
+                    .map(|x| (&x).try_into().ok())
+                    .flatten()
```
birchmd, line 251:
```suggestion
                    .and_then(|x| (&x).try_into().ok())
```

birchmd (approved):
As @mfornet has already mentioned, explicitly putting quotes around outputs in some places is quite error prone.

Approving for the sake of quickly addressing this bug, but we should definitely find a more robust way to handle this in the future (e.g. functions that are supposed to return json return a `JsonValue` and then output can be always be serialized using the same logic).

mfornet:
Thanks for fixing tests @sept-en

PR #159, opened 2021-06-23
Release 1.4.1

artob:


sept-en (commented):
@artob please note that we need to change the EVM version in the `VERSION` file during the release. Right now `get_version()` still returns "1.0.0".

PR #160, opened 2021-06-24
Fix(engine): Fix scheduled nightly clippy

birchmd:
The [scheduled run of nightly clippy was broken](https://github.com/aurora-is-near/aurora-engine/runs/2902922641?check_suite_focus=true). If it would have run it would have also found two more needless borrows, so those are fixed here too.


sept-en (approved):
LGTM!

joshuajbouw (approved):
Love these new clippy errors. Good to go.

PR #162, opened 2021-06-25
Tests: always compile with `mainnet` feature enabled.

sept-en:
I noticed that we still compile test-build without specifying `mainnet` feature that we recently added. Right now `mainnet` feature doesn't have any more features enabled, still I think we need this for consistency, so if new dependency features are added, we never miss them and always run tests on close to the production feature set.

Later we may even add 3 separate build configurations for tests running with `mainnet`, `testnet`, `betanet` features.


birchmd (approved):

joshuajbouw (approved):
Makes sense

PR #164, opened 2021-06-28
Move eth-connector tests under `src`, remove duplicate code

birchmd:
As a follow up to #155 we move the last test from `tests/`.


mfornet (approved):

src/parameters.rs
```diff
@@ -197,6 +197,7 @@ pub struct IsUsedProofCallArgs {
 
 /// withdraw result for eth-connector
 #[derive(BorshSerialize)]
+#[cfg_attr(test, derive(BorshDeserialize))]
```
sept-en, line 200: Recently we discovered that this `test` attribute won't work as expected as the cfg `test` won't be set as `cargo test` compiles only modules inside `tests` with this flag. 

Please check how we ended up implementing the desired behavior in [rainbow-bridge#598](https://github.com/aurora-is-near/rainbow-bridge/pull/598) by specifying target arch cfg instead.

src/types.rs
```diff
@@ -118,6 +118,7 @@ fn long_signature(name: &str, params: &[ParamType]) -> Hash {
 }
 
 #[derive(Default, BorshDeserialize, BorshSerialize, Clone)]
+#[cfg_attr(test, derive(serde::Deserialize, serde::Serialize))]
```
sept-en, line 121: The same thing here.

sept-en (approved):
Good job, thanks for this!

joshuajbouw (approved):
Straight forward

src/types.rs
birchmd: I think `test` is what we want here because the `serde` (de)serialization is only used in our own tests that are gated by `#[cfg(test)]`. If another crate were to depend on this one then I think it would be bad to expose the `serde` (de)serialization since we don't want others to be using it. So the fact that the crate can never be compiled with the `test` flag by other dependencies I think is a feature here.

src/parameters.rs
birchmd: Changed in https://github.com/aurora-is-near/aurora-engine/pull/164/commits/eb2a1c0af36beb00501b0cca9dfc99ad3ddba524

Here, unlike the other place, I will make the change because it is conceivable that we might depend on this crate and want the deserialization (e.g. if we wrote a client in Rust and shared types).

PR #165, opened 2021-06-28
EIP-2718: Typed Transaction Envelope support

birchmd:
EIP-2718 is required for the Berlin HardFork.

Spec: https://eips.ethereum.org/EIPS/eip-2718

This change is pretty minimal because it doesn't specify any new transaction types, only the format that new types will use. For example, when we add EIP-2930 support then we will add a new variant to the `EthTransaction` enum introduced in this PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -221,7 +221,7 @@ mod contract {
         let input = sdk::read_input();
 
         let EthTransaction::Legacy(signed_transaction) =
-            EthTransaction::try_from(input.as_slice()).sdk_expect("ERR_INVALID_TX");
+            EthTransaction::try_from(input.as_slice()).sdk_unwrap();
 
         let state = Engine::get_state().sdk_unwrap();
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -215,12 +215,13 @@ mod contract {
     /// Must match CHAIN_ID to make sure it's signed for given chain vs replayed from another chain.
     #[no_mangle]
     pub extern "C" fn submit() {
-        use crate::transaction::EthSignedTransaction;
-        use rlp::{Decodable, Rlp};
+        use crate::prelude::TryFrom;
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -215,12 +215,13 @@ mod contract {
...
-        use rlp::{Decodable, Rlp};
+        use crate::prelude::TryFrom;
+        use crate::transaction::EthTransaction;
 
         let input = sdk::read_input();
-        let signed_transaction =
-            EthSignedTransaction::decode(&Rlp::new(&input)).sdk_expect("ERR_INVALID_TX");
+
+        let EthTransaction::Legacy(signed_transaction) =
+            EthTransaction::try_from(input.as_slice()).sdk_expect("ERR_INVALID_TX");
```
sept-en, line 224: I guess as you already added custom parsing errors for TXs, we might use `sdk_unwrap()` which will specify the exact error why the provided TX is invalid. Though this will require implementing `AsRef<[u8]>` trait.

sept-en (approved):
LGTM!

joshuajbouw:
This can be improved, I think. I'll play around with it based on the idea that I had.

joshuajbouw (changes requested):
Code looks fine, but from what I understand is that we would need a hard fork enum of some sort perhaps. If we push it in as now and it makes it into the engine, we need to change all tx if I'm not mistakened.

Perhaps a precursor PR is required that will link up other parts such as the precompiles.

joshuajbouw (approved):
Yeah this is ok. Thanks for the clarification.

src/lib.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/165/commits/0d67c67be44d6b8fddc7b81bc88cb9697564ae9c

Issue #114, opened 2021-05-29
EthConnector: remove deprecated EIP-712 implementation


sept-en:
Currently, `prover.rs` file contains EIP-712 implementation that was previously used as a design approach for exits from Aurora. It's not used anywhere anymore so it makes sense to remove it.

Or do we need it for any future implementations?

PR #168, opened 2021-07-01
Remove Prover - deprecated EIP-712 implementation

mrLSD:
Related to: #114


src/types.rs
```diff
@@ -117,6 +117,15 @@ fn long_signature(name: &str, params: &[ParamType]) -> Hash {
...
 
+/// Validate Etherium address from string and return EthAddress
+pub fn validate_eth_address(address: String) -> EthAddress {
+    let data = hex::decode(address).expect("ETH_ADDRESS_FAILED");
+    assert_eq!(data.len(), 20, "ETH_WRONG_ADDRESS_LENGTH");
+    let mut result = [0u8; 20];
+    result.copy_from_slice(&data);
+    result
+}
+
```
sept-en, line 128: As we are changing this function in the PR and we are striving to have a more testable environment, I'd suggest returning an error instead of panicking from the function.

src/connector.rs
```diff
@@ -110,7 +109,7 @@ impl EthConnectorContract {
         // Get initial contract arguments
         let contract_data = EthConnector {
             prover_account: args.prover_account,
-            eth_custodian_address: validate_eth_address(args.eth_custodian_address),
+            eth_custodian_address: crate::types::validate_eth_address(args.eth_custodian_address),
```
sept-en, line 112: let's validate address beforehand and check for the error

sept-en (approved):
LGTM!

Added a couple of minor comments that I think worth addressing in this PR. If you don't think so, just ignore and let's add this in the following ones.

birchmd (approved):
LGTM
I agree with @sept-en about returning a `Result` from `validate_eth_address`.

joshuajbouw (approved):
Whoops forgot to push this review in on Friday. @sept-en is right on the return. Besides that, good to go.

src/types.rs
```diff
@@ -248,7 +266,7 @@ pub enum ErrorKind {
     InvalidEcRecoverSignature,
 }
 
-pub type Result<T> = core::result::Result<T, ErrorKind>;
+//pub type Result<T> = core::result::Result<T, ErrorKind>;
```
sept-en, line 269: It's better to remove such things later instead of commenting.

src/types.rs
```diff
@@ -117,13 +117,31 @@ fn long_signature(name: &str, params: &[ParamType]) -> Hash {
...
+#[derive(Debug)]
+pub enum ValidationError {
+    Decode,
+    WrongEthAddress,
+}
+
+impl AsRef<[u8]> for ValidationError {
+    fn as_ref(&self) -> &[u8] {
+        match self {
+            Self::Decode => b"FAILED_DECODE",
```
sept-en, line 129: I'd prefer having `EthAddressFailedDecode` error which will be more descriptive if we call the error `ValidationError` (a general name instead of `EthAddressValidationError`.
But we can rename this error later so if more types/methods use the same error type it will be easier to recognize which part is failed exactly. Otherwise, it may be hard to track the specific line of code that emits the error.

sept-en (approved):
@mrLSD thanks for adding those errors!

I have some minor comments there but I'm fine if we merge as is to iterate faster.

src/types.rs
```diff
@@ -117,6 +117,33 @@ fn long_signature(name: &str, params: &[ParamType]) -> Hash {
...
+#[derive(Debug)]
+pub enum ValidationError {
+    EthAddressFailedDecode,
+    WrongEthAddress,
+}
+
+impl AsRef<[u8]> for ValidationError {
+    fn as_ref(&self) -> &[u8] {
+        match self {
+            Self::EthAddressFailedDecode => b"FAILED_DECODE",
```
sept-en, line 129:
```suggestion
            Self::EthAddressFailedDecode => b"FAILED_DECODE_ETH_ADDRESS",
```

src/types.rs
sept-en: Fixed in 4e32931

src/types.rs
sept-en: Fixed in 4e32931

PR #171, opened 2021-07-08
ERC-20: allow admin to change metadata

sept-en:


etc/eth-contracts/contracts/EvmErc20.sol
```diff
@@ -13,16 +13,37 @@ import "./IExit.sol";
  * `ERC20` functions.
  */
 contract EvmErc20 is ERC20, AdminControlled, IExit {
+    string private _name;
```
0x3bfc, line 16: I think we don't need this here. It is already in [ERC20](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol#L40)

etc/eth-contracts/contracts/EvmErc20.sol
```diff
@@ -13,16 +13,37 @@ import "./IExit.sol";
...
+    constructor (string memory metadata_name, string memory metadata_symbol, uint8 metadata_decimals, address admin)
+        ERC20(metadata_name, metadata_symbol)
+        AdminControlled(admin, 0)
+    {
+        _name = metadata_name;
+        _symbol = metadata_symbol;
+        _decimals = metadata_decimals;
+    }
+
+    function name() public view override returns (string memory) {
```
0x3bfc, line 29: These functions already exists in the inherited ERC20 contract.

etc/eth-contracts/contracts/EvmErc20.sol
```diff
@@ -13,16 +13,37 @@ import "./IExit.sol";
...
+        return _name;
+    }
+
+    function symbol() public view override returns (string memory) {
```
0x3bfc, line 33: the same here

0x3bfc (changes requested):
Minor changes

etc/eth-contracts/contracts/EvmErc20.sol
sept-en: This field is `private` there

0x3bfc (approved):

etc/eth-contracts/contracts/EvmErc20.sol
sept-en: We need to override the function as the field it referenced is `private` there

etc/eth-contracts/contracts/EvmErc20.sol
sept-en: We need to override the function as the field it referenced is `private` there

etc/eth-contracts/hardhat.config.js
```diff
@@ -1,4 +1,4 @@
-require('@nomiclabs/hardhat-ethers');
+require('@nomiclabs/hardhat-waffle');
```
mfornet, line 1: Out of curiosity, why using waffle here, instead of ethers?

mfornet (approved):

birchmd (approved):

etc/eth-contracts/hardhat.config.js
sept-en: We need to have `hardhat-waffle` as this is adds additional functionality to `chai` (e.g. `.reverted`).
There's no need for `require("@nomiclabs/hardhat-ethers")` as `@nomiclabs/hardhat-waffle` already does it.

artob (approved):
LGTM :+1:

As a general note, as discussed with @sept-en today, we ought to be checking in the resulting Solidity bytecode as well, since it isn't necessarily 100% reproducible otherwise. I'll add that.

PR #179, opened 2021-07-14
ERC-20: forbid using invalid NEP141 AccountID for mapping

### LINKED ISSUE #173: Deploy ERC-20 tokens without case-sensitivity

sept-en:
Fixes #173.


birchmd:
The code itself looks good, but I'd like to understand the context here.

It was mentioned on the Monday call that this is meant to fix a phishing attack where someone tries to spoof a token using different capitalization. But I thought @mfornet said this attack wouldn't be possible if the front-end interfaces were sufficiently good.

It's also unclear to me what this attack would accomplish. Could you spell out in more detail what could happen from a user's perspective?

sept-en:
@birchmd it's not only about the attack itself (which needed to be really complex and I don't expect this to happen in the nearest future), but for the code consistency and right conceptual structure.
Without this fix deploying some address: `123abc..cba321.factory.bridge.near` and `123Abc..cba321.factory.bridge.near` (one letter is uppercase) will deploy 2 ERC-20 tokens for the same NEP-141 address. So we might end up with many different ERC-20 tokens deployed for the same NEP-141 account.

birchmd (approved):
Oh, [NEAR account IDs must be lowercase](https://docs.near.org/docs/concepts/account#account-id-rules) anyways, so it makes sense that we always ensure lowercase is used.

sept-en:
@birchmd yes, exactly. I guess this was explicitly made to eliminate even the possibility of such issues. For our case, it's expected that we can receive this as an input because users may specify ERC-20 checksummed address which contains both upper- and lower- case letters. But we need all those versions to point to the same ERC-20 on Aurora.

Created #180 for the purpose of checking the validity of all used `AccountId`s

joshuajbouw:
Still in draft, do you want that review?

src/engine.rs
```diff
@@ -463,15 +464,19 @@ impl Engine {
             .map(|result| Address(result.as_slice().try_into().unwrap()))
     }
 
-    pub fn register_token(&mut self, erc20_token: &[u8], nep141_token: &[u8]) {
+    pub fn register_token(&mut self, erc20_token: &[u8], nep141_account_id: &str) {
+        // Convert nep141 AccountID to lower case
+        let nep141_account_id = nep141_account_id.to_lowercase();
+        let nep141_key = nep141_account_id.as_bytes();
```
mfornet, line 470: Some suggestion, instead of doing lower case inside the contract, we can verify if the input is a valid near account id ([the code already exists here](https://github.com/aurora-is-near/aurora-engine/blob/50bf61cf7036e20675010c7399428de08f2e92ba/src/prelude.rs#L65). And fail if it is not a valid account id.

This way it is also clear that we are disallowing invalid account id, instead of trying to convert it to one.

sept-en: I like the suggestion, I mentioned this in #180. The only reason why I don't want to add it here is that I expect that someone may try manually creating the deployment transaction and providing Ethereum checksummed address as an input.

But from the another point of view, it's already NEP-141 input and not ERC-20 and there are no checksum addresses in Near.

Also, I still think that it's ok to have `.to_lowercase` in `get_erc20_from_nep141` (`engine.rs` function) but have a check in `get_erc20_from_nep141` (`lib.rs` function), because otherwise the function in Engine will need to have something like `Result<Option<Vec<u8>>>` as a return type or `Result<Vec<u8>, GetErc20FromNep141Error>`.

@birchmd @joshuajbouw gents, what do you think?

mfornet: Tbh, I think making this lower case looks very arbitrary given that we are only bridging NEP141 accounts which should be valid NEAR accounts. 

I wouldn't do `.to_lowercase` on `get_` either, for consistency + education. Also it is easy to reason about our code. 
I like the approach of compilers, they try to be smart reporting errors, but not too smart fixing them, to avoid unintended consequences or assumptions.

I'm ok in the engine function either returning `Result<Vec<u8>, GetErc20FromNep141Error>` (👈 preferred)  or `Option<Vec<u8>>`. Also we should consider add some logging in case of failure.

src/engine.rs
```diff
@@ -463,15 +464,19 @@ impl Engine {
             .map(|result| Address(result.as_slice().try_into().unwrap()))
     }
 
-    pub fn register_token(&mut self, erc20_token: &[u8], nep141_token: &[u8]) {
+    pub fn register_token(&mut self, erc20_token: &[u8], nep141_account_id: &str) {
```
joshuajbouw, line 467: This should take `nep141_account_id` as a `String`.

src/engine.rs
joshuajbouw: What I don't like is that this is making a copy of the `nep141_account_id` `&str`. That and we are making redundant copies of `erc20_token` too if I'm not mistaken. This method should take ownership of both parameters and looking at its use, it should be ok.

joshuajbouw: If I recall... And this is just based on previous knowledge. The `to_lowercase` method is a bit "expensive".

The return type you mentioned is fine.

src/engine.rs
```diff
@@ -463,15 +464,19 @@ impl Engine {
...
-        assert!(map.lookup_left(nep141_token).is_none());
-        map.insert(nep141_token, erc20_token);
+        assert!(map.lookup_left(nep141_key).is_none());
+        map.insert(nep141_key, erc20_token);
     }
 
-    pub fn get_erc20_from_nep141(&self, nep141_token: &[u8]) -> Option<Vec<u8>> {
-        Self::nep141_erc20_map().lookup_left(nep141_token)
+    pub fn get_erc20_from_nep141(nep141_account_id: &str) -> Option<Vec<u8>> {
+        Self::nep141_erc20_map().lookup_left(nep141_account_id.to_lowercase().as_bytes())
```
joshuajbouw, line 479: Same issue here. Its copying the `&str` with the `to_lowercase`.

joshuajbouw:
In general, there needs to be better data handling that doesn't involve copying borrowed data first.

src/engine.rs
sept-en: I'd prefer having `AccountId` which is an alias to `String` right now. But in this case, I get a warning (which will result in error in CI) on clippy to use `&str` instead. Do you think I should force clippy to bypass this check?

joshuajbouw: No, for that kind of check I wouldn't want to bypass it.

sept-en:
@birchmd @joshuajbouw @mfornet 
Gents, please check the update! This contains custom error types and forbids providing invalid NEP-141 `AccountID`.

birchmd (approved):
LGTM

I wonder if it would be nicer to validate the input closer to the boundary of the system and not have errors bubbling up from core logic. I think the PR is ok as-is, but just something that I thought I'd mention to see what others think.

With the aforementioned model the code would look something like

```rust
/// Can only be constructed with `ValidAccountId::from_bytes` which does validation.
/// Therefore if a function takes `ValidAccountId` it does not need to worry about invalid accounts.
struct ValidAccountId<'a>(&'a [u8]);
impl ValidAccountId {
    fn from_bytes(bytes: &[u8]) -> Option<Self> {
        if is_valid_account(bytes)  { Some(Self(bytes)) }
        else { None }
    }
}

pub fn register_token(&mut self, erc20_token: &[u8], nep141_token: ValidAccountId) { ... }
pub fn get_erc20_from_nep141(nep141_token: ValidAccountId) -> Option<Vec<u8>> { ... }

// lib.rs
fn deploy_erc20_token() {
    ...
    let nep141_token = ValidAccountId::from_bytes(&args.nep141).sdk_expect(ERR_INVALID_NEP141_ACCOUNT_ID);
    engine.register_token(address.as_bytes(), nep141_token).sdk_unwrap();
    ....
}
```

sept-en:
@birchmd thanks, this is a good suggestion! I wanted to do something similar in the scope of PR that will fix #180.
I agree with you as the model you mentioned will simplify the code, its usage, and readability.

joshuajbouw (approved):

joshuajbouw:
@sept-en when you get back, please fix the conflicts. Thanks buddy.

PR #182, opened 2021-07-15
EIP-2930 refactor and extra changes

joshuajbouw:
This is a bit of a catch-all and will have solutions to outstanding problems that I have been aware of but hadn't realised that it will meet its limitations now.

This is intended to be added to the `develop` branch after https://github.com/aurora-is-near/aurora-engine/pull/181. It just kept growing until it virtually refactored all of the precompile library.

Shortlist of changes:
* Removes the state generic from everything
* Reduces and implies a lot of the code. Code reduction of around 325 lines.
* Easier to read
* No duplicate code

### DIFF
--- a/src/precompiles/mod.rs
+++ b/src/precompiles/mod.rs
@@ -8,12 +8,13 @@ mod native;
 mod secp256k1;
 use evm::{executor, Context, ExitError};
 
+use crate::parameters::PromiseCreateArgs;
 pub(crate) use crate::precompiles::secp256k1::ecrecover;
-use crate::prelude::{vec, Vec};
+use crate::prelude::{vec, BTreeMap, PhantomData, Vec};
 use crate::AuroraState;
 use crate::{
     precompiles::blake2::Blake2F,
-    precompiles::bn128::{BN128Add, BN128Mul, BN128Pair},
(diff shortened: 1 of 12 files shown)


src/precompiles/mod.rs
```diff
@@ -368,84 +196,290 @@ impl<'backend, 'config> executor::Precompiles<AuroraStackState<'backend, 'config
...
-            ExitToEthereum::<AuroraStackState>::ADDRESS => {
-                Some(ExitToEthereum::<AuroraStackState>::run(
-                    input, target_gas, context, state, is_static,
-                ))
+        let output = self.precompiles.get_fun(&address).and_then(|p| {
+            let mut res = (p)(input, target_gas, context, is_static);
+            if let Ok(mut o) = &mut res {
+                if let Some(p) = o.promise.take() {
+                    state.add_promise(p)
+                }
```
mfornet, line 204:
```suggestion
        let output = self.precompiles.get_fun(&address).and_then(|precompile| {
            let mut res = (precompile)(input, target_gas, context, is_static);
            if let Ok(mut output) = &mut res {
                if let Some(promise) = output.promise.take() {
                    state.add_promise(promise)
                }
```

birchmd (approved):
Nice job!

sept-en (approved):
Very nice!

Issue #176, opened 2021-07-13
Extend JSON parser for integer types


mrLSD:
Currently, for JSON parser base number is f64.
We don't use negative integer numbers, so should be separated and implemented `JsonValue` entities:
* U8(u8)
* U64(u64)
* F64(f32)
JSON not supported `u128` type.

Entity `Number(f64)` should be removed.

Implement new Traits for parser and parser functions.

Rewrite tests for: `f64, u8, u64`

mrLSD:
#141

birchmd:
Done in #183

PR #183, opened 2021-07-16
Added parser for Integer types

mrLSD:
* updated upstream parser rjson: [PR #2](https://github.com/yw662/rjson/pull/2)
* added parsers for: u64, i64
* changed parsers for: f64, u8
* changed tests

Related to #176


sept-en (approved):
LGTM!

I'd still leave this as draft until the upstream PR is merged and the current one is updated accordingly having the reference to the upstream. Though If we are ok maintaining our own `rjson` repo, it's ok for me to proceed.

Cargo.toml
```diff
@@ -62,7 +62,7 @@ logos = { version = "0.12", default-features = false, features = ["export_derive
 ethabi = { git = "https://github.com/darwinia-network/ethabi", branch = "xavier-no-std", default-features = false }
 hex = { version = "0.4", default-features = false, features = ["alloc"] }
 byte-slice-cast = { version = "1.0", default-features = false }
-rjson = { version = "0.3.1", default-features = false }
+rjson = { git = "https://github.com/aurora-is-near/rjson", branch="master", default-features = false, features = ["integer"] }
```
joshuajbouw, line 65: This should be tagged to a specific commit

src/json.rs
```diff
@@ -289,28 +297,29 @@ mod tests {
         assert_eq!(err, JsonError::NotJsonType);
     }
 
+    #[test]
+    #[should_panic]
+    fn test_json_type_u64_with_u128_value() {
+        let _ = parse_json(format!(r#"{{"foo": {} }}"#, u128::MAX).as_bytes());
```
birchmd, line 303: Shouldn't this parse as an `F64` if it doesn't fit in `U64`? Seems like panicking is too extreme.

src/json.rs
```diff
@@ -289,28 +297,29 @@ mod tests {
...
         let json = parse_json(r#"{"foo": 123}"#.as_bytes()).unwrap();
         let val = json.u64("foo").ok().unwrap();
         assert_eq!(val, 123);
 
-        let json = parse_json(format!(r#"{{"foo": {} }}"#, u64::MAX).as_bytes()).unwrap();
-        let val = json.u64("foo").ok().unwrap();
-        assert_eq!(val, u64::MAX);
+        // let json = parse_json(format!(r#"{{"foo": {} }}"#, u64::MAX).as_bytes()).unwrap();
+        // let val = json.u64("foo").ok().unwrap();
+        // assert_eq!(val, u64::MAX);
```
birchmd, line 314: Please delete outdated code.

joshuajbouw, line 314: Remove commented out code here

src/json.rs
```diff
@@ -30,9 +32,8 @@ pub enum JsonError {
 
 #[derive(Debug, Ord, PartialOrd, Eq, PartialEq)]
 pub enum JsonOutOfRangeError {
```
joshuajbouw, line 34: Think that this should be `JsonError`, or keep it the same and change the enums to just be `U8` & `U128`.

joshuajbouw (changes requested):

Cargo.toml
sept-en: agree!

src/json.rs
sept-en: AFAIR we intentionally added a separate sub-enum for the `JsonOutOfRangeError`. What's the point of removing it?

src/json.rs
```diff
@@ -30,9 +32,8 @@ pub enum JsonError {
...
-    U8,
-    U64,
-    U128,
+    OutOfRangeU8,
```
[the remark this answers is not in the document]
mrLSD: Our github actions doesn't complain for current PR state. And it's related to our current toolchain.
All lint is a success.

src/json.rs
mrLSD: Yes, it was special extended casef or error kind.

src/json.rs
mrLSD: It's `rjson` library parser logic.
Also `u128::MAX > f64::MAX`
I extended with more explicit panuc message

src/json.rs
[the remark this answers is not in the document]
birchmd: We run nightly clippy [as a scheduled job](https://github.com/aurora-is-near/aurora-engine/actions/workflows/scheduled_lints.yml), not part of normal CI. We do not run as part of normal CI to not block PRs spuriously, but when we already know something will fail the scheduled job then we should try to prevent it.

Please remove the `OutOfRange` prefix.

src/json.rs
birchmd: No, that is not correct. `f64::MAX` is much larger than `u128::MAX`. For example see code [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7ccfae9adc9068876d9a5bdf58fc4115).

I know it's coming from `rjson`. I am saying I think the `rjson` logic is wrong to panic here (in my opinion). But I don't know if there is a json spec or something which says how big numbers are allowed to be.

frankbraun: According to https://www.json.org JSON numbers are always floats, but the spec doesn't give a a maximum size. A classic case of "implementation defined"...

src/json.rs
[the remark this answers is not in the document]
mrLSD: make check-clippy without errors.
So pls provide me cargo command to check it correctly.

src/json.rs
birchmd: You are right that `f64` does not have the same precision as `u128`, but it can still represent bigger numbers.

The question is whether it is expected to panic if I pass in a number which is larger than a `u64` and has more precision than an `f64` can represent. I argue that most users would not expect a panic in this case and would instead expect a loss of precision. It is very normal in floating point arithmetic to have loss of precision (which of course is why we are trying to avoid floats in our use case).

src/json.rs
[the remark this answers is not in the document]
birchmd: `cargo +nightly clippy --no-default-features --features=mainnet -- -D warnings` (you may need to update your nightly install to the latest one first via `rustup`)

src/json.rs
mrLSD: So seems we need put some changes to our `rjson` implementation.

src/json.rs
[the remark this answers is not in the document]
mrLSD: So I'm try reproduce
1. It will use current engine nightly: cargo clippy --no-default-features --features=mainnet -- -D warnings
**SUCCESS**
2. cargo clippy --no-default-features --features=mainnet -- -D warnings
3. cargo +nightly clippy --no-default-features --features=mainnet -- -D warnings
**FAILED**:
```
    Checking evm v0.28.0 (https://github.com/aurora-is-near/sputnikvm.git?rev=09d4fe09dcb5fcabed8c1076699c8a2e70f14c23#09d4fe09)
error: internal compiler error: unexpected concrete region in borrowck: ReStatic

```

src/json.rs
mrLSD: https://github.com/aurora-is-near/rjson

src/json.rs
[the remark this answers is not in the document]
birchmd: Very strange. I do not see this error. Try doing `rustup update nightly` before running `clippy`.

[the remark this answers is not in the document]
mrLSD: I made clean. And run again.
```
$ make clean
$ rustc +nightly -V
rustc 1.56.0-nightly (a6ece5615 2021-08-03)
$ cargo +nightly clippy --no-default-features --features=mainnet -- -D warnings
```
Same result - SUCCESS.

src/json.rs
birchmd: Are you going to make those changes or should I? I suppose there is also the question of whether we should bother putting time into `rjson` at all if we will eventually move to using `serde` as @joshuajbouw works on integrating `std` into the contract.

src/json.rs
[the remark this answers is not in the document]
birchmd: Weird. I thought it would have triggered [clippy::enum_variant_names](https://rust-lang.github.io/rust-clippy/v0.0.212/index.html#enum_variant_names) because each enum variant starts with the same `OutOfRange` prefix. But if it does not trigger then I guess it's fine.

Cargo.toml
```diff
@@ -62,7 +62,7 @@ logos = { version = "0.12", default-features = false, features = ["export_derive
 ethabi = { git = "https://github.com/darwinia-network/ethabi", branch = "xavier-no-std", default-features = false }
 hex = { version = "0.4", default-features = false, features = ["alloc"] }
 byte-slice-cast = { version = "1.0", default-features = false }
-rjson = { version = "0.3.1", default-features = false }
+rjson = { git = "https://github.com/aurora-is-near/rjson", banch="master", ref="cc3da949", default-features = false, features = ["integer"] }
```
sept-en, line 65: A couple of typos:

```suggestion
rjson = { git = "https://github.com/aurora-is-near/rjson", branch = "master", rev = "cc3da949", default-features = false, features = ["integer"] }
```

src/json.rs
mrLSD: @birchmd @joshuajbouw - my suggestion to merge it **AS IS** - current implementation is ready for it.
But after some near period we will migrate to `serde`. And just do not spend additional time to `rjson` question.

mrLSD: Currently, we spent a lot of time for `rjson`. I believe enough. The current solution is working. So the next step will be improvements with `serde`.

birchmd (approved):
`rjson` panicking sometimes is suboptimal. But the changes in this PR are an improvement over what we are doing currently. And with `serde` replacing `rjson` coming soon it's fine to merge this.

PR #189, opened 2021-07-17
Speed up workflow build caching

strokovok:
- Remove default caching mechanism. Contents of `~/.cargo` are now shared across runs since all runners are now on the same machine. Caching of `target` directory using remote storage is not efficient and can be implemented faster as written below.
- Wrap Rust compiler with [sccache](https://github.com/mozilla/sccache), which partly allows to reuse compiled dependencies ([it's also mentioned here](https://doc.rust-lang.org/cargo/guide/build-cache.html)). Currently, there are 481mb of cache-hittable artifacts.
- Share `target/debug` and `target/release` across runs. This allows profits even on partial cache-hit from previous builds + saves a lot of disk space for us :)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,14 +8,12 @@ jobs:
     steps:
       - name: Clone the repository
         uses: actions/checkout@v2
-      - name: Cache Cargo artifacts
+      - name: Cache target artifacts
         uses: actions/cache@v2
         with:
           path: |
-            ~/.cargo/registry
-            ~/.cargo/git
             target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          key: ${{ runner.os }}-aurora-engine-target-${{ hashFiles('**/Cargo.lock') }}
       - name: Install the toolchain
         uses: actions-rs/toolchain@v1
         with:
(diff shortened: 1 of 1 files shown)


joshuajbouw (approved):
Mmm, yeah conceptually this looks ok. Why the move away from `cache@v2` though?

strokovok:
> Why the move away from `cache@v2` though?

Well, cache@v2 works like this:
- In case of cache-hit it downloads ~700mb archive from GitHub storage and unzips it
- In case of cache-miss it builds from scratch, zips all artifacts (2.5GB -> 700mb) and uploads archive to GitHub storage
- Both of this steps work ~40secs in our case
- GitHub storage is limited to 5GB, so this cache-packages do not live long :)
- One change of dependency completely invalidates cache

There are multiple reasons to get rid of it:
1. Since we have our own machine with a lot of memory, we don't need to spend time for zipping/unzipping/uploading/downloading. We just can copy all the files as they are to some local cache in less than 1 second.
2. In case of cache@v2 it's only possible to work with huge monolith archives - one change of some dependency changes the hash and completely invalidates the whole cache. With our own machine we are able to use partial cache invalidation which significantly increases cache-hit probability for single dependency

joshuajbouw:
Makes sense. Thanks for the explanation and the care and attention to that assessment.

PR #190, opened 2021-07-19
Feat(engine): Use new math API host functions

birchmd:
Updates our contract to use the new host functions merged in https://github.com/near/nearcore/pull/4380

Note: these host functions have not yet been released to mainnet, but they are on testnet. The host functions are in protocol version 46; you can see [here](https://explorer.testnet.near.org/stats) that 46 is the version of testnet, while [mainnet](https://explorer.near.org/stats) is still on 45.

Using the host functions (specifically `ecrecover`) makes simple ETH transfer on Aurora less than 1/5th the cost in terms of NEAR gas.

Requires #193 before merging.


src/sdk.rs
```diff
@@ -372,6 +382,47 @@ pub fn keccak(input: &[u8]) -> H256 {
...
+    }
+}
+
+/// Recover address from message hash and signature.
+pub fn ecrecover(hash: H256, signature: &[u8]) -> Result<Address, ECRecoverErr> {
+    unsafe {
+        let hash_ptr = hash.as_ptr() as u64;
+        let sig_ptr = signature.as_ptr() as u64;
+        let recover_register_id = 1;
+        let keccak_register_id = 2;
```
joshuajbouw, line 402: Should be consts.

src/sdk.rs
```diff
@@ -372,6 +382,47 @@ pub fn keccak(input: &[u8]) -> H256 {
...
+        let result = exports::ecrecover(
+            32,
+            hash_ptr,
+            64,
+            sig_ptr,
+            signature[64] as u64,
+            0,
+            recover_register_id,
+        );
```
joshuajbouw, line 411: Any value that is concrete, we should be making consts for.

joshuajbouw (changes requested):
These should use the testnet feature so that we don't accidentally use it prematurely.

src/sdk.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/190/commits/ede8a4df16419d61b2d50f301b2eae188324c1d3

src/sdk.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/190/commits/ede8a4df16419d61b2d50f301b2eae188324c1d3

birchmd:
> testnet feature

Done in https://github.com/aurora-is-near/aurora-engine/pull/190/commits/5c5544bd00051b8e53795d0f08137ee1742c2992

joshuajbouw:
Requires https://github.com/aurora-is-near/aurora-engine/pull/193 before merging.

joshuajbouw (approved):

PR #193, opened 2021-07-20
Add different networks to makefile

joshuajbouw:
Now that we have testnet specific code with an upcoming PR https://github.com/aurora-is-near/aurora-engine/pull/190, this adds the ability to compile for different networks. This prevents accidental compilation for the wrong network.

This does make some rather large changes. For one, it doesn't make symlinks. I am somewhat predicting that in the current setup it is way too easy to accidentally upload the wrong .wasm file to the wrong network. I make it more explicit. Instead of symlinks, this now copies the WASM file to the root folder with explicit names like `mainnet-release.wasm`.

Todo:
- [x] add options to make file
- [x] update CI

There does need to be a follow-up PR that will include CI testing for each network release.


.gitignore
```diff
@@ -24,5 +24,4 @@ res/
 etc/state-migration-test/target/
 
 node_modules/*
-artifacts/
```
birchmd, line 27: Why remove this?

birchmd (approved):
LGTM

We'll need to have CI run testnet tests at some point, but can be done in a future PR.

.gitignore
joshuajbouw: Its a duplicate, just a nit that I noticed.

PR #195, opened 2021-07-21
Remove rust toolchain installation in workflow

strokovok:
Makes no sense after migrating to self-hosted runner (local runtime already have this toolchain).
Furthermore, it creates unwanted data-races


joshuajbouw (approved):
As discussed, should be good. Perhaps re-run the CI a few times to ensure the issue is fixed?

PR #197, opened 2021-07-22
Optimize for performance instead of code size

birchmd:
Benchmarks show that this significantly reduces gas usage (around 47%) for non-trivial contracts (e.g. uniswap). We have a larger binary as a result (678K -> 865K on my machine), but I think it is worthwhile.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,6 +26,7 @@ jobs:
   bully-build:
     name: Bully build
     runs-on: self-hosted
+    needs: test
     steps:
       - name: Clone the repository
         uses: actions/checkout@v2
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ autobenches = false
 crate-type = ["cdylib", "rlib"]
 
 [profile.release]
-opt-level = "z"
+opt-level = 3
 debug = false
 debug-assertions = false
 overflow-checks = true


.github/workflows/tests.yml
```diff
@@ -26,6 +26,7 @@ jobs:
   bully-build:
     name: Bully build
     runs-on: self-hosted
+    needs: test
```
birchmd, line 29: This is unrelated to the `opt-level` change, but seems to make CI run better (prevents some kind of race causing a file to be deleted which is still needed by other tests).

strokovok: Hmm, that's interesting.
I suspect that, unfortunately, this will not solve the problem. Two or more concurrent "test" jobs can possibly produce the same issue.
Let me explain: each job is launched on random free runner (1 of 16), so each of them have it's own isolated workdir. Except only moment - they still share `target/release` and `target/debug` directories (you can see that I made them as symlinks), and this directories are shared for all runs and all jobs (even concurrent). I did a little research before doing it, it was stated that cargo `target` dir is OK for concurrent builds, but after seeing this - I doubt a lot. I mean, it's impossible to explain those races another way (because those are only files that are shared/concurrently used). I should probably redesign this part...
Do you know which particular file was under race? If it's under `target/release` or `target/debug` - it's definitely my fault here :)

joshuajbouw:
Thats a huge improvement, but I think we should have a discussion about which path we should go if contract bloat is better to reduce cost of gas, which I do agree, is far better. We should also have quite a large reduction when we have the native precomps too.

joshuajbouw (approved):

PR #198, opened 2021-07-23
Parallelize test suits

strokovok:
- Parallelize tests using [strategy.matrix](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix)
- Fix issue with data race (copy target parts instead of symlinking). `target/release` and `target/debug` are not actually concurrency-safe, my bad

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,6 @@ jobs:
   bully-build:
     name: Bully build
     runs-on: self-hosted
-    needs: test
     steps:
       - name: Clone the repository
         uses: actions/checkout@v2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,13 +13,15 @@ jobs:
         uses: actions/checkout@v2
       - run: echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - name: Prepare target directory
-        run: |
-          rm -rf target && mkdir target
-          ln -s ~/build-cache/aurora-engine/debug/ target/debug
-          ln -s ~/build-cache/aurora-engine/release/ target/release
+        run: rm -rf target && cp -r ~/build-cache/aurora-engine target
       - run: make ${{ matrix.net }}net-test-build
       - name: Run ${{ matrix.net }}net cargo test
         run: cargo test --locked --verbose --features ${{ matrix.net }}net-test
+      - name: Cache and clear target directory
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -3,26 +3,25 @@ on: [push, pull_request]
...
-      - run: make testnet-test-build
-      - run: make betanet-test-build
-      - name: Run mainnet cargo test
-        run: cargo test --locked --verbose --features mainnet-test
-      - name: Run testnet cargo test
-        run: cargo test --locked --verbose --features testnet-test
-      - name: Run betanet cargo test
-        run: cargo test --locked --verbose --features betanet-test
+          mv -n target/debug/* ~/build-cache/aurora-engine/debug
+          mv -n target/release/* ~/build-cache/aurora-engine/release
```
birchmd, line 23: I don't think `no-clobber` will be good enough because we may want to overwrite some files that are updated sometimes. To be honest I'm not 100% sure on this because I don't know exactly how rust compilation artifacts are represented. Regardless, I think `rsync` might be a better choice here.

I also think it would be good to update the whole `target` directory, not just `debug` and `release` because there are some other build artifacts we care about in `target` not under those directories (e.g. solidity build artifacts and wasm build artifacts).

birchmd (commented):
Thanks for looking into this! I think this is definitely a step in the right direction!

.github/workflows/tests.yml
strokovok: I've came to the exactly same conclusion.
Artifacts in `debug` in `release` are suffixed with some hash, so I assumed that compiler kinda only creates artifacts with new hash instead of overwriting existing ones. But it seems to me that it's not true, because cleaning build-cache and starting from scratch gives the same time result :(

So this new solution doesn't help at all, you are right. I'll delete it.
I'm now thinking about implementing the same thing that we had in `actions-rs/cache`, but saving cache locally, and without using compression. I'll do this in another commit. Thanks for details!

birchmd (approved):
It's not obvious to me that this will be faster because even though they are now running in parallel, they are also now building from scratch. But it probably doesn't hurt and I assume you'll bring in a new caching solution in a new PR.

Please also merge in `develop` and revert [my `needs` change](https://github.com/aurora-is-near/aurora-engine/commit/068e38eff6340c948d3f3d6e32ba25e04ab1fa75#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR29) since it will not be needed now without the caching.

PR #201, opened 2021-07-27
Add build-caching to testing workflow

strokovok:
Implement same thing that we had with [actions/cache](https://github.com/actions/cache), but use locally installed [cache-util](https://github.com/aurora-is-near/devops-stuff/tree/main/cache-util) instead.

~4m -> ~1m


joshuajbouw (approved):
Looks good.

Issue #191, opened 2021-07-20
NEP-141 compliance correctness


joshuajbouw:
This is a tracking issue for [NEP-141] correctness. There are a couple of issues, namely revolving around JSON values. Right now we are Borsh serializing / deserializing but this should be JSON. That and a couple of the returns are returning not as JSON.

There are a few blockers to fixing this which needs to be attacked asap. Firstly, we should be using `serde`'s JSON library `serde-json-core` due to the issue with integer parsing https://github.com/aurora-is-near/aurora-engine/issues/176. Note we can't use `serde_json` because it is not `no_std` compatible. I would imagine that `serde` will be smaller in size than `rjson` so that does need to be measured.

NOTE: The ones that are crossed out below are because of an issue that we can not solve right now without getting off of `no_std` due to `serde`s conflicting with `borsh`.

Tasks:
- [x] ~~Create a branch which replaces `rjson` with `serde` with `serde-json-core`.~~
- [x] ~~Replace `dev-dependency` use of `serde_json` with `serde-json-core`.~~
- [x] ~~Measure changed compilation size. Hopefully, this is less.~~
- [ ] Fix NEP-141 compliance by using JSON with `rjson`, not Borsh.
- [x] fn `storage_deposit` does not return valid JSON. There is a last comma on the which isn't valid in `json`. See: https://github.com/aurora-is-near/aurora-engine/issues/191#issuecomment-886572448
- [x] fn `ft_balance_of` returns number instead of `String` number. See: https://github.com/aurora-is-near/aurora-engine/issues/191#issuecomment-886572448
- [x] fn `ft_metadata` does not exist

Related issues:
* ~~https://github.com/aurora-is-near/aurora-engine/issues/188~~
* ~~https://github.com/aurora-is-near/aurora-engine/issues/177~~
* ~~https://github.com/aurora-is-near/aurora-engine/issues/176~~

Related PRs:
* https://github.com/aurora-is-near/aurora-engine/pull/183

[NEP-141]: https://github.com/near/NEPs/blob/master/specs/Standards/FungibleToken/Core.md

mfornet:
Regarding NEP-141 compliance here are some incompatibilities that should be addressed. Notice this is not a comprehensive list of all issues with our implementation, for that we need to go through every I/O interface defined by NEP-141 and check the implementation is correct.

[...]

PR #202, opened 2021-07-28
NEP-141 compliance correctness

mrLSD:
Related to #191 
Fixed:
* `storage_deposit`

Added:
* `ft_metadata`
* `get_metadata`


joshuajbouw (approved):

src/fungible_token.rs
```diff
@@ -27,6 +27,17 @@ pub struct FungibleToken {
     pub account_storage_usage: StorageUsage,
 }
 
+#[derive(Default, BorshDeserialize, BorshSerialize, Clone)]
```
birchmd, line 30: Instead of deriving `Default`, I think it might be nice to have a constructor that returns the values for ETH on Aurora (since that is the primary use case for this struct).

birchmd (changes requested):
Where is the fix for `ft_balance_of`?

src/connector.rs
```diff
@@ -58,6 +59,7 @@ impl EthConnectorContract {
             contract: Self::get_contract_data(&EthConnectorStorageId::Contract),
             ft: Self::get_contract_data(&EthConnectorStorageId::FungibleToken),
             paused_mask: Self::get_contract_data(&EthConnectorStorageId::PausedMask),
+            metadata: Self::get_contract_data(&EthConnectorStorageId::FungibleTokenMetadata),
```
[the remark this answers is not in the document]
joshuajbouw: Thats true. In this case, we would.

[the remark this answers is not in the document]
joshuajbouw: Possibly a new method the owner can call to add the metadata to their FT?

Makefile
```diff
@@ -106,6 +106,11 @@ clean:
 	@rm -Rf *.wasm
 	cargo clean
 
+test-pure: FEATURES=mainnet,integration-test,meta-call
+test-pure: Cargo.toml Cargo.lock $(shell find src -name "*.rs") 
+	RUSTFLAGS='-C link-arg=-s' $(CARGO) build --target wasm32-unknown-unknown --release --no-default-features --features=$(FEATURES) -Z avoid-dev-deps
+	$(CARGO) test --features mainnet-test
```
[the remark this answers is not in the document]
mrLSD: It's not run Solidity code.

src/connector.rs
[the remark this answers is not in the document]
mrLSD: Yes, it should be some migration for existing contracts.

mrLSD:
> Where is the fix for `ft_balance_of`?

It was fixed on `develop` branch

src/fungible_token.rs
mrLSD: What did you mean? How is it related to Metadata?

Makefile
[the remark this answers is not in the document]
birchmd: Some rust tests will fail without the solidity artifacts if they do not exist, so this Make target will not work in that case. And in the case the solidity artifacts already exist they are not recompiled, meaning it will behave the same as `test-mainnet` in that case. So I think we should not have this Make target.

src/connector.rs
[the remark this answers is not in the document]
mrLSD: But currently, we don't have a migration mechanism. #137 is stopped.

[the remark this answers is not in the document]
birchmd: We have [this function which is called on migration](https://github.com/aurora-is-near/aurora-engine/blob/develop/src/lib.rs#L182). It could fill in the value we want for existing deployments.

Makefile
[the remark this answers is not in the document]
mrLSD: So let's understand why we need Makefile? For development simplification. For example, to run Node-related code I should run it in a separate environment. But I should run ONLY rust code. So for rust-related source code, I shouldn't use Node dependency. So without that, it's impossible for me to run any tests.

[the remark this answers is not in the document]
mrLSD: We don't have a dedicated command for that in Makefile. So I should have one exactly for cases when I need to test only Rust code.

src/parameters.rs
```diff
@@ -272,6 +273,7 @@ pub struct FinishDepositEthCallArgs {
 pub struct InitCallArgs {
     pub prover_account: AccountId,
     pub eth_custodian_address: AccountId,
+    pub metadata: FungibleTokenMetadata,
```
birchmd, line 276: Changing this struct means we also need to update the json library:

- https://github.com/aurora-is-near/aurora.js/blob/master/src/engine.ts#L153
- https://github.com/aurora-is-near/aurora.js/blob/master/src/schema.ts#L122
- https://github.com/aurora-is-near/aurora.js/blob/master/src/schema.ts#L257

src/fungible_token.rs
birchmd: Having empty metadata is never useful.

The only time we use this struct is to describe the FT which is contained in the Aurora Engine contract, that is the Aurora ETH. So it would be useful for us to have a constant like `FungibleTokenMetadata::ETH` which is used in places where we currently have calls to `default()`. This could also be useful in the state migration.

src/connector.rs
```diff
@@ -616,6 +626,11 @@ impl EthConnectorContract {
     pub fn set_paused_flags(&mut self, args: PauseEthConnectorCallArgs) {
         self.set_paused(args.paused_mask);
     }
+
+    /// Return metdata
+    pub fn get_metadata(&self) {
+        sdk::return_output(&self.metadata.try_to_vec().unwrap()[..]);
```
birchmd, line 632: Do we need to return this as json (not Borsh)?

src/connector.rs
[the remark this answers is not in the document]
mrLSD: Do you mean hardcoded?

Makefile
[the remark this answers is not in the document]
birchmd:
```
$ git status
On branch develop
Your branch is up to date with 'origin/develop'.

$ make test-mainnet
RUSTFLAGS='-C link-arg=-s' cargo build --target wasm32-unknown-unknown --release --no-default-features --features=mainnet,integration-test,meta-call -Z avoid-dev-deps
    Finished release [optimized] target(s) in 0.92s
cp target/wasm32-unknown-unknown/release/aurora_engine.wasm mainnet-release.wasm
cargo test --features mainnet-test
   Compiling aurora-engine v1.0.0 (/datadrive/birchmd/aurora-engine)
    Finished test [optimized] target(s) in 42.02s
     Running unittests (target/debug/deps/aurora_engine-0cb5a3fcb128ded8)

running 88 tests
test benches::benches ... ignored
test json::tests::test_json_type_u128 ... ok
test json::tests::test_json_type_bool ... ok
test json::tests::test_json_type_string ... ok
test json::tests::test_json_type_u64 ... ok
test json::tests::test_json_type_u8 ... ok
test parameters::tests::test_roundtrip_view_call ... ok
test parameters::tests::test_view_call_fail ... ok
test precompiles::blake2::tests::test_blake2f ... ok
test precompiles::bn128::tests::test_alt_bn128_add ... ok
test precompiles::hash::tests::test_ripemd160 ... ok
test precompiles::hash::tests::test_sha256 ... ok
test precompiles::identity::tests::test_identity ... ok
test precompiles::modexp::tests::test_berlin_modexp_big_input ... ok
test precompiles::modexp::tests::test_berlin_modexp_bigger_input ... ok
test precompiles::modexp::tests::test_berlin_modexp_empty_input ... ok
test precompiles::bn128::tests::test_alt_bn128_mul ... ok
test meta_parsing::tests::test_parse_type ... ok
...
```

`test-mainnet` already will only run rust tests (see above), we do not need a new Make target for this use case.

src/fungible_token.rs
mrLSD: Why it is empty??? We initialized that.
Default - only for tests. We can implement something different for tests, if we need.

src/connector.rs
mrLSD: It's Borsh

src/connector.rs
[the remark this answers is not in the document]
birchmd: Yes, hardcoded. Something like

```rust
    pub extern "C" fn state_migration() {
        if connector::read_metadata().is_empty() {
            connector::write_metadata(FungibleTokenMetadata::ETH);
        }
    }
```

src/fungible_token.rs
birchmd: I understand right now that `Default` is only used in tests.

What I am saying is that having this `Default` implementation is misleading because it implies that empty metadata is a reasonable value, but it is not.

Instead we could have a constant which contains the metadata that is useful for our (Aurora) use case.

src/connector.rs
birchmd: I know we are currently returning it as Borsh. I am asking whether we should be returning it as json for compliance with what other contracts are expecting.

src/lib.rs
```diff
@@ -591,6 +591,11 @@ mod contract {
         );
     }
 
+    #[no_mangle]
+    pub extern "C" fn get_metadata() {
```
birchmd, line 595: This function name should be `ft_metadata` according to [the spec](https://nomicon.io/Standards/FungibleToken/Metadata.html#reference-level-explanation).

src/connector.rs
mrLSD: We discussed it with @joshuajbouw 
According to [Metadata](https://github.com/near/NEPs/blob/master/specs/Standards/FungibleToken/Metadata.md) there is no concrete restrictions.

src/fungible_token.rs
mrLSD: Good point. I'll remove it. And put some default impl in tests.

Makefile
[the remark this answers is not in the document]
mrLSD: You're wrong:
```
$ make test-mainnet 
cd etc/eth-contracts && yarn && yarn build
/bin/sh: 1: yarn: not found
Makefile:36: recipe for target 'etc/eth-contracts/res/EvmErc20.bin' failed
make: *** [etc/eth-contracts/res/EvmErc20.bin] Error 127
```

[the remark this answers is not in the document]
mrLSD: It's impossible to run it without Node.js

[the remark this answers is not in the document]
birchmd: That should only happen if `EvmErc20.bin` does not exist. And if that is the case then you cannot compile the rust part either because it [imports that file](https://github.com/aurora-is-near/aurora-engine/blob/develop/src/lib.rs#L351).

src/connector.rs
birchmd: This seems like an oversight in the spec.

The whole point of this PR is to be compliant with what other contracts (e.g. `ref.finance`) are expecting and I worry that they might be expecting json.

For example the wrapped NEAR FT returns json

```
$ NEAR_ENV=mainnet ./bin/near view wrap.near ft_metadata
View call: wrap.near.ft_metadata()
{
  spec: 'ft-1.0.0',
  name: 'Wrapped NEAR fungible token',
  symbol: 'wNEAR',
  icon: null,
  reference: null,
  reference_hash: null,
  decimals: 24
}
```

birchmd: Indeed `ref.finance` does assume metadata is returned via json: [`ftGetTokenMetadata`](https://github.com/ref-finance/ref-ui/blob/main/src/services/ft-contract.ts#L59) calls into [`ftViewFunction`](https://github.com/ref-finance/ref-ui/blob/main/src/services/ft-contract.ts#L24) which calls into [`viewFunction` (in `near-api-js`)](https://github.com/near/near-api-js/blob/master/src/account.ts#L499) with the default parser, [which is json](https://github.com/near/near-api-js/blob/master/src/account.ts#L503).

src/connector.rs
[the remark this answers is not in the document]
mrLSD: I think that data should be put from CLI args, not as constants.

src/connector.rs
mrLSD: We found with @joshuajbouw that reference FT implementation does not compline NEP-141.
So we should discuss it publicly.

src/connector.rs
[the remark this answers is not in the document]
birchmd: That might be a more flexible way of doing the migration. Though in my opinion it introduces some complexity without adding much value because in this case we do want to write a constant into the storage.

src/parameters.rs
mrLSD: Adde issue: https://github.com/aurora-is-near/aurora.js/issues/7

src/connector.rs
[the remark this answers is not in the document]
birchmd: Maybe even having the metadata in storage at all is too much complexity. Taking wrapped NEAR as an example, they do [hardcode the metadata](https://github.com/near/core-contracts/blob/master/w-near/src/lib.rs#L48-L58).

It would save us some trouble if we did the same because we wouldn't need to change the `InitCallArgs` either.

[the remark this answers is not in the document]
mrLSD: I added migration. But let's take a broad discussion on Friday's call.
It'll be easy to change as a hard-coded constant anymore.

mrLSD:
NB: for discussion
* Metadata `init` as constant data (currently implemented as user parameters input)
* `ft_metadata` results `json vs borsh`

src/lib.rs
```diff
@@ -180,8 +180,23 @@ mod contract {
...
+            &[EthConnectorStorageId::FungibleTokenMetadata as u8],
+        );
+
+        //=========================================================
+        // Migrate Metadata
+        if !sdk::storage_has_key(&metadata_key[..]) {
+            use crate::fungible_token::FungibleTokenMetadata;
+
+            let metadata =
+                FungibleTokenMetadata::try_from_slice(&sdk::read_input()).expect(ERR_FAILED_PARSE);
```
birchmd, line 197: The `state_migration` function is not called directly by a user, it is [called as part of the upgrade process](https://github.com/aurora-is-near/aurora-engine/blob/develop/src/sdk.rs#L456-L462), so you would need to thread this data through from the [`deploy_upgrade` function](https://github.com/aurora-is-near/aurora-engine/blob/develop/src/lib.rs#L169).

To see an example of the upgrade flow (stage -> deploy -> state_migration), we have the [state migration test](https://github.com/aurora-is-near/aurora-engine/blob/develop/src/tests/state_migration.rs#L12).

mrLSD: Looks as really bad design for migration.
You can't control migration, for that design.

mrLSD: We should put parameters to `sefl_upgrade` all call migration directly.
But for all the best migrations that I know - it should be a direct call. Otherwise, you can't control it.

birchmd: It was suggested we do it this way in [Eugene Kuzyakov's review](https://github.com/aurora-is-near/aurora-engine/pull/74/files#diff-8bcc54eddebf73a8eb83cf6739e4e992583f4fb2cee5c7a8f317680b465cfdfbR333).

His argument was that upgrade should include state migration otherwise you could end up in a broken place where the new code doesn't understand how to read the old state. That's why we have state migration automatically called during upgrade.

joshuajbouw (changes requested):
As per comments in Slack, do the following to get information for a constant:

* Please check with NEAR what that JSON file is supposed to be like. It doesn't explain exactly what it should be in the standard.
* Double check our symbol with Alex or Arto
* SVG encode the aurora icon, get from Matt
* Calculate reference hash of the JSON file
* Give JSON file to Matt to upload to our website

joshuajbouw (approved):
Looks good to me now, I believe that was everything.

src/connector.rs
```diff
@@ -616,6 +626,37 @@ impl EthConnectorContract {
...
         self.set_paused(args.paused_mask);
     }
+
+    /// Return metdata
+    pub fn get_metadata(&self) {
+        let icon = if let Some(ref icon) = self.metadata.icon {
+            format!(r#""{}""#, icon)
+        } else {
+            "null".to_string()
+        };
```
birchmd, line 636: Nit: Could avoid allocations here by using `Cow<str>`. Should also probably factor this out into a function like

```rust
fn string_or_null<T: Display>(data: Option<&T>) -> Cow<str> { ... }
```

src/lib.rs
```diff
@@ -180,8 +180,30 @@ mod contract {
...
+        //=========================================================
+        // Migrate Metadata
+        if !sdk::storage_has_key(&metadata_key[..]) {
+            use crate::fungible_token::FungibleTokenMetadata;
+
+            let metadata = FungibleTokenMetadata {
+                spec: "ft-1.0.0".to_string(),
+                symbol: "ETH".to_string(),
+                name: "Ether".to_string(),
+                icon: None,
```
birchmd, line 200: I still think we should provide an icon. Does anyone have experience with the [data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) and [optimized SVG](https://codepen.io/tigt/post/optimizing-svgs-in-data-uris) that [the spec](https://nomicon.io/Standards/FungibleToken/Metadata.html#reference-level-explanation) talks about?

src/connector.rs
```diff
@@ -616,6 +626,37 @@ impl EthConnectorContract {
...
+        let reference = if let Some(ref reference) = self.metadata.reference {
+            format!(r#""{}""#, reference)
+        } else {
+            "null".to_string()
+        };
+        let reference_hash = if let Some(ref reference_hash) = self.metadata.reference_hash {
+            format!("{:?}", reference_hash)
```
birchmd, line 643: The reference implementation [encodes as base64 string](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/fungible_token/metadata.rs#L15). We should do the same.

src/lib.rs
```diff
@@ -180,8 +180,30 @@ mod contract {
     /// code.
     #[no_mangle]
     pub extern "C" fn state_migration() {
-        // This function is purposely left empty because we do not have any state migration
-        // to do.
+        // Only owner can call migration
+        sdk::assert_private_call();
```
birchmd, line 184: Since @artob said he does not use the state migration mechanism, I think we should remove the implementation here, change `Connector::get_metadata` to return an `Option` (may or may not be present in storage), and then in `ft_metadata()` provide our default value (symbol = ETH, name = Ether, etc) in the `None` case.

birchmd (changes requested):

src/connector.rs
mrLSD: Looks like overoptimization.

src/lib.rs
birchmd: The [official Ethereum assets](https://ethereum.org/en/assets/#brand) don't come in svg, but it seems like lots of tokens use png/jpeg in their metadata (e.g. `rekt.tkn.near`, `token.skyward.near`). So I suggest we set the data URL to be

```
data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAGQAAABkCAYAAABw4pVUAAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAAJcEhZcwAADsQAAA7EAZUrDhsAAAs3SURBVHhe7Z1XqBQ9FMdFsYu999577wUfbCiiPoggFkQsCKJP9t57V7AgimLBjg8qKmLBXrD33hVUEAQ1H7+QXMb9Zndnd+/MJJf7h8Pu3c3Mzua3yTk5SeZmEZkySplADFMmEMOUCcQwZQggHz58EHfu3FF/2a0MAWTjxo2iWbNm6i+7ZT2QW7duiUWLFolixYqJQ4cOqVftlfVAZs6cKdauXSuqV68uKlWqpF61V1YDoUXMmTNHrFu3TtSoUUNCmTBhgnrXTlkL5Nu3b2Ly5MmyuwJIzZo1RaNGjUTx4sXFu3fvVCn7ZC2QVatWiQULFvwPSL169USnTp1UKftkJZCbN2+KGTNmSBiLFy/+BwhWoUIFsX//flXaLlkJZPr06WkwIoE0btxYNGzYUFSsWFGVtkvWATlw4IB05BqGGxAMBz9u3Dh1lD2yCsjXr1/THHk8IDwvVaqUeP36tTraDlkFZOXKldKRO2HEAoKD79ixozraDlkD5Pr16/848nhANBQc/N69e9VZzJc1QCIduRcgGA4eKLbICiD79u37nyN3WiwgvMZ7Y8eOVWczW8YDwZFPmTIlauvA4gHhsUSJEuLFixfqrObKeCArVqxwdeROiwUE43UcfNu2bdVZzZXRQK5duyYduRsEp8UDog1fsnPnTnV2M2U0kFiO3GlegeDgy5cvr85upowFQqg6d+5cVwCR5hUI71NuzJgx6lPMk5FAPn365Doij2ZegWCUIUX/9OlT9WlmyUggy5Yti+vInZYIEAwH37JlS/VpZsk4IJcvX5bTsl5bB5YoEMqRDd62bZv6VHNkHJBp06YlBANLFAiGgy9btqz6VHNkFJBdu3Z5duROSwYIxjEjRoxQn26GjAHy8ePHuCPyaJYsEMozgn/48KG6ivBlDJAlS5Yk5MidlgqQ+vXri+bNm6urCF9GALl48aJ05G6V7cWSBYJxDOu5Nm/erK4mXBkBJBlH7rRUgGAmOfjQgZBbSsaROy1VIBjHDxs2TF1VeAoVyPv37+WI3K2SE7H0AMKxJUuWFHfv3lVXF45CBZKKI3daegDBcPBNmzZVVxeOQgNy/vz5hEfkbsbxAGFtb6pAOL5y5cpye0NYCg1Iqo5c29KlS2WEVKdOHdGkSZOUoeDgS5cura4yeIUCZMeOHWLevHkpASEBScvAB/Xs2VMUKVJE1K1bV44pUgHDcbVq1RJDhgxRVxusAgfy5s0bMXXq1IRgOMsuX75c7gcZP368aN++vez3W7VqJfLnzy8KFCggU+tUKNncZMFwDA6eNcRBK3AgCxculOas8HiG82duffXq1WLkyJGiRYsWokGDBrI1UPHMlQOjaNGisqUUKlRIPrKclLKA0RUdWfnRDNCUD1qBAjl79qyYNWuWa6VHGq0CEGw7oHsaNGiQrCBMg9DmBKJNgylYsKAciQOFfYhUtlcwHEe3GKQCA/Lnzx/PyUMc9Zo1a+SAsV+/fvLXSgXxa3eCiAXECaZw4cISDPPpGijniweG93HwXHtQCgwIk0E4cjcAGhItAf8AuG7dukknzbgAENFgYLGAaNNgKMcibGYNdXdGxUeDgz8aOHCg+hb+KxAgr169kpUcCUKb01GzOJrKonuJB0KbFyBOAw4thgCgdu3aaWAA4AYGB8/a4iAUCBBG405Hrv2Dm6MGhFulx7JEgWjTYHisVq2a/GxapBMGgLguLAj5DuTMmTP/OHLtqPETdAW6u4h01IlYskC06e6MIICROlA0GH19vM51+y1fgfz+/TvNkWtHjR/p27ev7JboJrx2S7EsVSAYUDCgcC4CAEbtXJsGg4PnO/kpX4Fs3bpVwiB0BEz37t09O+pELD2AOE23GM5ZpkwZGeVxraRnBgwYoL6dP/INCCNyfAeOukOHDmmZVLcKTdXSG4jTNBidAaDlXLlyRX3L9JdvQPr06SObvHbU6dUa3MxPINp0d5Y3b16RJ08e9S3TX74Befz4sejcubOoWrWqdNi2AgEEj8DIkiWLdO4PHjxQ3zL95asPQQcPHpSTR/gOv6D4BUQ7+uzZs4usWbOK7du3q2/ln3wHosU+j3LlysmIxa1SUzG/gOTLl0+2ilGjRqlv4b8CA4K+fPkievXqJZt9MgPAaJbeQHT3hA9kJX6QChSI1smTJ+U4RKct3Co5EUsvIHRP2bJlEzlz5hRHjhxRVxusfANy4cIF9Sy6GLnrAZhbRXu1VIEAguiJVuHlfltbtmxRz9JfvgHhxpQMBt++fatecdfPnz/lYIvtAcmOU1IBQi4LEG3atJHXEkssEWK0fvv2bfVK+svXLosJKW4AQ3QSb07h6tWr0uEz+Eq0G0sGCAM+IieOI98WS3///hVDhw4VOXLkkAlRP+W7D9mwYYNMLtJa4n1xRBqe3bIMKL2CSQQI3VPu3Lllq+C64olsNPMnBCJdunRRr/qnQJw6IS/pdypg/vz5cff38YscPny49C9eujGvQCgDiB49eqhPii4WgJPuAQQ+Lqi1v4EAefToUVrWFzCsyWIx2q9fv1QJd92/f1+0bt1aLlaINdqPB4TuCRD80rmtbCzhR8hG66SizvKeOHFClfBXgQBBe/bskfcr0dO1pOFZU3Xs2DFVIrqY/q1SpUpa1tUrELqnXLlySRhe5jKYw2d2kHBcz4OwIjLIXVaBAUF0V5Ezh7Nnz5Z27949VSq6CBDoOphHiQYECDyyTgsQ/fv3V0dH1/Hjx2V6h7wbEAguMH4ABBlBKlAgbneE090Yd21Yv369+P79uyrtrpcvX/6TtIwEorsnlvA8efJEHeUuRuFdu3aVKR2CCCcMnpNyf/78uSodjAIFgk6fPh11txQtCGBebhlO0pLuhKSlBkISEBhMjMXTxIkTZYVzvBOEhgFQriloBQ4EEUrGWhKEryEyu3HjhjoiuggWqDxAeOnrufcW5QkUIkFoGEBiUi0MhQKEeel4q995DyjcZ/Hz58/qSHfRrcTbSUuZdu3ayTEOYawbDIz3iLDiRYB+KRQgiP/3waJrNxjagMI0MK2AKC1ZjR49Wm5/JqEZDQTGe8A4fPiwOjJ4hQYEsS3By/5CwFCOVsWAzatIAhKVed3MQznWEIepUIEg/IUzFI5lgCEgYG1XrKQlyT9CY3wFXZBb5UcaURZ+JWyFDoSs8KRJk2L6E6dRDoB0YyQtneukSGAOHjxYDu70KNut8iONckRcJvzbpNCBIAZmXrcpYBoekRpgyBQzhiE1wkDOKwiMsuSr6BJNkBFAENEU45DIyo9nwGGxNs44ERAY5QlxmQsxRcYAIcxMdKubtmS3RVOe7u3Hjx/qKsKXMUAQA0EiKbdKj2XJAiEC2717t/p0M2QUEETaw0so7LREgVCO8l4Sj0HLOCAIB+81FMYSAUIZQmGSkybKSCAs1I7MCseyRIEwaveSJwtDRgJBR48e9RwKewXC+0x0AdtUGQsEMSL3cnMaL0B4j1wWc/Qmy2ggzG/ruXg3ENq8AmHgyCSZyTIaCLp06VLce8DHA8LrrGDxMnEVtowHgjZt2hR1QguLB4R0Su/evdXZzJYVQJBe25UoELK4Nv1PQ2uAPHv2LKo/iQaEv0mNeFn4bYqsAYL4p5IsGfIChOfMb7Dp1CZZBQTRQiJDYTcgerrWNlkHhHVbkV1XJBAemXDirqe2yTog6Ny5c9LJayhOIBgrS1h1b6OsBIKocB0KO4FwtwVu7WSrrAWC9NouDYQsLstCbZbVQNjmwCwjQFjCwzTuqVOn1Lt2ymogiBk/PafOfbdsl/VAEEBs+gfEsZQhgDChxVKgjKAMASQjKROIYcoEYpgygRglIf4D6lp/+XognSwAAAAASUVORK5CYII=
```

(you can copy and paste into your browser to see it is correct)

src/connector.rs
mrLSD: I'm not sure what do you mean. We receive as input parameter Base64 string, and return for ft_metadata - Base64 .
string. We don't need to encode/decode it.

birchmd (approved):

joshuajbouw (approved):

joshuajbouw:
Just need conflicts sorted and good to go.

birchmd:
> Just need conflicts sorted and good to go.

Resolved in https://github.com/aurora-is-near/aurora-engine/pull/202/commits/2cfda549017f5a78cdeb2b63c77fc4acd530beec

Issue #170, opened 2021-07-07
Implement ETH tests to ensure full compatible implementation


joshuajbouw:
As the same with upstream SputnikVM, we should also implement the [whole suite of tests][Tests]. 

Upstream [only tests for general state changes][UpstreamTests]. We would need to test for more than just that as we have a more complete implementation with precompiles and others. While we do feel confident about our implementation and upstream does take care of the bulk of the main tests, it still would be great to ensure full compatibility as the EVM updates.

[Tests]: https://github.com/ethereum/tests
[UpstreamTests]: https://github.com/rust-blockchain/evm-tests

mrLSD:
> Upstream only tests for general state changes. We would need to test for more than just that as we have a more complete implementation with precompiles and others.

I think it should be a separated tasks: EVM tests and precompiles tests. And if I'm correct precompiles is not a part of SpuntikVM.

mrLSD:
Also, we can't test it as a fully compatible part of Engine. For example, `backend` is incorporated into the executor. And it's private. But for `VM` testing and state testing, we need custom `MemroyBackend`.

But good news, that we can test `state, vm` as some part of Engine/EVM. So I think it will have good value.

andrcmdr:
[Design proposal for testing framework for Engine compliance with Ethereum Test Suite/Bundle and the Ethereum Yellow Paper](https://github.com/aurora-is-near/aurora-engine/discussions/377)

PR #208, opened 2021-08-02
Timestamp should be in milliseconds for Ethereum compatibility

birchmd:
This was caught during the IOSG hackathon. It's something we probably would have caught ourselves if we had better compatibility test coverage, e.g. #170


sept-en (approved):
Thanks for adding this! LGTM!

joshuajbouw (approved):

PR #211, opened 2021-08-03
Update etc/state-migration-test/Cargo.lock

birchmd:
Looks like this was missed during the version upgrade.


joshuajbouw (approved):
Can confirm.

Issue #209, opened 2021-08-02
Fix `block_hash`


joshuajbouw:
For EIP-1898, it requires polling of `block_hash`es which we do not support at this time.

Initially, @birchmd  submitted https://github.com/near/nearcore/pull/4273 for `nearcore` which returned the block hash from NEAR but how they and ETH does transactions could lead to issues. There likely is a better solution for this though which we can do on our end.

Per discussion and call, it has been decided that simply this needs to be returned:
```
sha256(concat(
    block_address_type_prefix, // e.g. 0xA
    near_block_height,
    chain_id,
))
```

* `block_address_type_prefix`: is a future proofing measure
* `near_block_height`: return from `sdk`'s `block_index`
* `chain_id`: Aurora's ETH chain ID.

sept-en:
The suggestion for the block hash function:

```
sha256(concat(
    block_address_type_prefix, // e.g. 0xA
    near_block_height,
    chain_id,
))
```

PR #213, opened 2021-08-03
feat(engine): Blockhash definition

birchmd:
This PR defines an Aurora block hash for each height of the NEAR blockchain (including skipped ones). This definition is unique on different networks (via chain_id) and different accounts on the same network(*). It also includes a prefix byte which may allow us to change this scheme in the future and verify the change is implemented based on the hash output.

(*) Note: in the future, when we have sharding, this may need to change because there could be multiple Aurora accounts on the same network that are contributing to the same set of blocks.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -20,6 +20,10 @@ use crate::state::AuroraStackState;
 use crate::storage::{address_to_key, bytes_to_key, storage_to_key, KeyPrefix, KeyPrefixU8};
 use crate::types::{u256_to_arr, AccountId, Wei, ERC20_MINT_SELECTOR};
 
+/// Used as the first byte in the concatenation of data used to compute the blockhash.
+/// Could be useful in the future as a version byte, or to distinguish different types of blocks.
+const BLOCK_HASH_PREFIX: u8 = 0;
+
 #[cfg(not(feature = "contract"))]
 pub fn current_address() -> Address {
     crate::types::near_account_to_evm_address("engine".as_bytes())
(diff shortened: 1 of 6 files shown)


src/engine.rs
```diff
@@ -245,6 +249,36 @@ impl Engine {
...
+    /// ```text
+    /// block_hash = sha256(concat(
+    ///     BLOCK_HASH_PREFIX,
+    ///     block_height as u64,
+    ///     chain_id,
+    ///     engine_account_id,
+    /// ))
+    /// ```
+    pub fn compute_block_hash(chain_id: [u8; 32], block_height: u64, account_id: &[u8]) -> H256 {
+        let mut data = Vec::with_capacity(1 + 8 + 32 + account_id.len());
```
0x3bfc, line 264: Avoiding magic numbers. I suggest to replace them with predefined constants: 

-`BLOCK_HASH_PREFIX_SIZE` -> `1`
-`BLOCK_HEIGHT_SIZE` -> `8`
-`CHAIN_ID_SIZE` -> `32`

birchmd: Good call. Done in https://github.com/aurora-is-near/aurora-engine/pull/213/commits/3334f18a944cf97048cde9f3b38a67db7706fc63

sept-en (approved):
LGTM!

NB: This PR fixes #209

sept-en:
NB: this RP fixes #209

joshuajbouw (approved):

Issue #216, opened 2021-08-04
State changes are not committed when a transaction execution returns an error


birchmd:
Currently in the engine contract we [panic when there is an error ](https://github.com/aurora-is-near/aurora-engine/blob/7ae5657cb6f41eb1389c4d5c3b99293dce47876e/src/types.rs#L421)from a transaction execution. This causes the state changes made during the contract execution to revert on NEAR. This is a problem because we want some state changes to stick, for example incrementing the nonce (i.e. as of today replay attacks could be possible in some cases). Another example is [charging gas](https://github.com/aurora-is-near/aurora-engine/pull/215) (presently gas will never be charged for failed transactions; including errors like OOG).

To see this bug in action take a look at [this test](https://github.com/birchmd/aurora-engine/blob/state-revert-bug/src/tests/state_migration.rs#L34). Running `make check` on that branch you see it fails because the nonce is not incremented:

```
thread 'tests::state_migration::test_state_revert' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', src/tests/state_migration.rs:65:5
```

Note: we do have tests which check for state changes on error, however they were run using the NEAR runtime directly which does not have the "revert-on-panic" logic baked in the way it is in a production node, therefore they do not show this bug. The example above uses `near-sdk-sim` to reproduce it.

To fix this we must not panic in the contract after any point where we want to keep state changes. This means we should encode failures like OOG or NSF (not sufficient funds) into the `SubmitResult` and always return it. It would probably be nice if this was backward compatible with the current `SubmitResult`, however that might be too hacky and we should just bite the bullet and make a breaking change while usage is still fairly low.

joshuajbouw:
I tried to play around with this and it was trivial to do the proposed `SumbitResult` change. The issue though is that I am pretty sure that the test itself is wrong. It is checking the status which isn't an incrementing value that can be used to indicate that nonce incremented?

birchmd:
> I am pretty sure that the test itself is wrong

Indeed, there was an issue with the original PoC test (but not the conclusion that this bug exists and needs fixing). [A corrected test](https://github.com/aurora-is-near/aurora-engine/pull/218/commits/81e580c769e6a8f03c8f6488c25f14faba04cfa8) is now present in #218

joshuajbouw:
This is closed as of https://github.com/aurora-is-near/aurora-engine/pull/218

PR #218, opened 2021-08-05
Return actual status of a transaction

joshuajbouw:
In response to the issue https://github.com/aurora-is-near/aurora-engine/issues/216 that @birchmd had found, it is true that certain errors should still be considered "successful" even though the execution had in fact failed. This fixes that.

Unfortunately, this is a breaking pub API change, but functionally similar.

DO note that this is missing an [important test](https://github.com/birchmd/aurora-engine/blob/state-revert-bug/src/tests/state_migration.rs#L34) which @birchmd had created as I am unsure if this test is actually correct, but I am currently unsure how such a test could be created.

Closes https://github.com/aurora-is-near/aurora-engine/issues/216.

### DIFF
--- a/src/parameters.rs
+++ b/src/parameters.rs
@@ -61,28 +61,31 @@ impl From<Log> for ResultLog {
     }
 }
 
+/// The status of a transaction.
 #[derive(Debug, BorshSerialize, BorshDeserialize, PartialEq, Eq)]
-pub enum EvmStatus {
-    Succeed,
-    Revert,
+pub enum TransactionStatus {
+    Succeed(Vec<u8>),
+    Revert(Vec<u8>),
     OutOfGas,
     OutOfFund,
     OutOfOffset,
+    CallTooDeep,
 }
 
-impl EvmStatus {
+impl TransactionStatus {
     pub fn is_ok(&self) -> bool {
-        *self == EvmStatus::Succeed
+        matches!(*self, TransactionStatus::Succeed(_))
     }
 
     pub fn is_revert(&self) -> bool {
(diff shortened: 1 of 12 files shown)


birchmd (approved):

src/tests/state_migration.rs
```diff
@@ -49,19 +50,32 @@ fn deploy_evm() -> AuroraAccount {
             0,
         )
         .assert_success();
+    let init_args = InitCallArgs {
+        prover_account,
+        eth_custodian_address: "d045f7e19B2488924B97F9c145b5E51D0D895A65".to_string(),
```
sept-en, line 55: How we got this `eth_custodian_address`?

src/parameters.rs
```diff
@@ -61,13 +61,53 @@ impl From<Log> for ResultLog {
...
+            || *self == TransactionStatus::CallTooDeep
+    }
+}
+
+impl AsRef<[u8]> for TransactionStatus {
+    fn as_ref(&self) -> &[u8] {
+        match self {
+            Self::Succeed(_) => b"SUCCESS",
+            Self::Revert(_) => b"ERR_REVERT",
+            Self::OutOfFund => b"ERR_OUT_OF_FUND",
```
sept-en, line 97: As I mentioned in the other PR, `OutOfFund` is not correct, but if you want to keep it for consistency with SputnikVM, we should at least use the correct description in the error itself:

```suggestion
            Self::OutOfFund => b"ERR_OUT_OF_FUNDS",
```

sept-en (approved):
Nice, this is really needed! LGTM!

PR #220, opened 2021-08-05
ETH-tests: vm engine

mrLSD:
Related to issue: #170 

Basic source tests for SpuntikVM: [evm-tests](https://github.com/rust-blockchain/evm-tests)

Cover only: VM tests

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "src/tests/eth/ethtests"]
+	path = src/tests/eth/ethtests
+	url = https://github.com/ethereum/tests/
--- a/src/tests/eth/utils.rs
+++ b/src/tests/eth/utils.rs
@@ -1,0 +1,150 @@
+use evm::backend::MemoryAccount;
+use primitive_types::{H160, H256, U256};
+use sha3::{Digest, Keccak256};
+use std::collections::BTreeMap;
+
+pub fn u256_to_h256(u: U256) -> H256 {
+    let mut h = H256::default();
+    u.to_big_endian(&mut h[..]);
+    h
+}
+
+pub fn unwrap_to_account(s: &ethjson::spec::Account) -> MemoryAccount {
+    MemoryAccount {
+        balance: s.balance.clone().unwrap().into(),
+        nonce: s.nonce.clone().unwrap().into(),
+        code: s.code.clone().unwrap().into(),
+        storage: s
+            .storage
+            .as_ref()
+            .unwrap()
+            .iter()
+            .map(|(k, v)| {
+                (
+                    u256_to_h256(k.clone().into()),
+                    u256_to_h256(v.clone().into()),
+                )
+            })
+            .collect(),
+    }
+}
+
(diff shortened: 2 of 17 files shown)


src/tests/eth/utils.rs
```diff
@@ -0,0 +1,150 @@
+use evm::backend::MemoryAccount;
+use primitive_types::{H160, H256, U256};
+use sha3::{Digest, Keccak256};
+use std::collections::BTreeMap;
+
+pub fn u256_to_h256(u: U256) -> H256 {
```
birchmd, line 6: I don't like that we are verbatim copying so [many lines of code](https://github.com/rust-blockchain/evm-tests/blob/master/jsontests/src/utils.rs#L6-L134) from [`evm-jsontests`](https://github.com/rust-blockchain/evm-tests/tree/master/jsontests). Could we add that crate as a dependency instead maybe?

mrLSD: We cant' These crate has bugs. For example, when I tried to run tests from that git repo I received errors.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/tests/eth/ethtests"]
+	path = src/tests/eth/ethtests
+	url = https://github.com/ethereum/tests/
```
mfornet, line 3: should we lock a particular commit?

joshuajbouw: Yes absolutely.

src/tests/eth/utils.rs
joshuajbouw: Compatibility errors or?

joshuajbouw, line 6: Just a note, I swear that there is an `Into` trait created for `U256.into()` for `H256`..?

mrLSD: Code just can't compile. I mean upstream. So as cate dependency I have the same problem.

mrLSD: utils file just copy-pusted from upstream...

.gitmodules
mrLSD: Yes, locked particular commit, as in reference implementation. For some reason, VMTests was deleted from https://github.com/ethereum/tests/

src/tests/eth/utils.rs
joshuajbouw: Can you check? I don't want to add to tech debt by adding in unnecessary methods if we do not need them.

mrLSD (commented):
U256 don't contain `into` implementation for H160

src/tests/eth/utils.rs
mrLSD: U256 doesn't contain into implementation for H160

joshuajbouw (approved):
This is good to go for me then. Just resolve the conflict.

joshuajbouw:
These unfortunately can not be merged into this library due to conflicting license from sourced code. Must target a new library entirely. `aurora-engine-tests` will be fine.

PR #222, opened 2021-08-09
Update etc/state-migration-test/Cargo.lock

birchmd:
Looks like this got missed in https://github.com/aurora-is-near/aurora-engine/pull/183

### DIFF
--- a/etc/state-migration-test/Cargo.lock
+++ b/etc/state-migration-test/Cargo.lock
@@ -708,7 +708,7 @@ dependencies = [
 [[package]]
 name = "rjson"
 version = "0.3.2"
-source = "git+https://github.com/aurora-is-near/rjson#cc3da9495e7e520900d66d2b517539f74fff93cf"
+source = "git+https://github.com/aurora-is-near/rjson?rev=cc3da949#cc3da9495e7e520900d66d2b517539f74fff93cf"
 
 [[package]]
 name = "rlp"


etc/state-migration-test/Cargo.lock
```diff
@@ -708,7 +708,7 @@ dependencies = [
 [[package]]
 name = "rjson"
 version = "0.3.2"
-source = "git+https://github.com/aurora-is-near/rjson#cc3da9495e7e520900d66d2b517539f74fff93cf"
+source = "git+https://github.com/aurora-is-near/rjson?rev=cc3da949#cc3da9495e7e520900d66d2b517539f74fff93cf"
```
sept-en, line 711: Shouldn't it be something like

```suggestion
source = "git+https://github.com/aurora-is-near/rjson?rev=cc3da9495e7e520900d66d2b517539f74fff93cf"
```

birchmd: The lock file is automatically generated so I can only assume it's right as is

sept-en: Ok, it didn't generate this change for me, but I'm ok with it :)

sept-en (approved):

joshuajbouw:
I don't think that this is quite required anymore as the `Cargo.lock` will just be naturally updated.

PR #223, opened 2021-08-09
Fix(engine): do not panic when user has insufficient balance to cover gas

birchmd:
Now that https://github.com/aurora-is-near/aurora-engine/pull/218 is merged we can properly handle the case where the signer of a transaction does not have enough balance to cover the gas cost. Note in the original logic that the `sdk_unwrap()` caused a panic which meant the nonce increase was reverted (same problem as reported in https://github.com/aurora-is-near/aurora-engine/issues/216 ). This PR includes a `near-sdk-sim` test which shows the behaviour is now correct (this new test would have failed previously).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,13 +77,13 @@ mod contract {
     use borsh::{BorshDeserialize, BorshSerialize};
 
     use crate::connector::EthConnectorContract;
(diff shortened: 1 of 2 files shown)


sept-en (approved):
Nice, thanks for adding this!

joshuajbouw (approved):

PR #225, opened 2021-08-10
Fix(testnet): Method to fix balance of aurora account

birchmd:
**Background:**

Previously we minted and burned bridged ETH in the NEP-141 contract when it became usable in the EVM. However, in #133 we changed that design to instead hold all the ETH currently available in the EVM in the Aurora account on the NEP-141 contract. This change was made because it makes the `ExitToNear` logic much simpler (it is simply a transfer from the Aurora account to the destination account in the NEP-141 contract).

However we did not mint tokens to cover the ones that were burned when they entered the EVM prior to #133. This includes the over 12k ETH that is held by the faucet.

**Issue:**

As a result, the `ExitToNear` functionality is now broken on testnet because the Aurora account in the NEP-141 contract has run out of tokens. Note: this only impacts testnet because #133 was merged before the eth-connector was deployed to mainnet, so mainnet has always had the updated design.

**Solution (this PR):**

This PR introduces a method for testnet only which mints tokens to the Aurora account in the NEP-141 contract, restoring the invariant which exists with the `ExitToNear` design change introduced in #133. Namely, this invariant is that the total available amount of ETH (ETH sent to the exit precompile addresses is no longer available) in the EVM is equal to the balance of the Aurora account in the NEP-141 contract. A test is included to show the method works properly.

This new method only needs to be called one time on testnet, so after this has happened we can revert this PR to avoid confusion in the future.

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -26,7 +26,7 @@ pub(crate) const PAUSE_WITHDRAW: PausedMask = 1 << 1;
 #[derive(BorshSerialize, BorshDeserialize)]
 pub struct EthConnectorContract {
     contract: EthConnector,
-    ft: FungibleToken,
+    pub ft: FungibleToken,
     paused_mask: PausedMask,
 }
 
@@ -578,7 +578,7 @@ impl EthConnectorContract {
     }
 
     /// Save eth-connector contract data
-    fn save_ft_contract(&mut self) {
+    pub(crate) fn save_ft_contract(&mut self) {
         sdk::save_contract(
             &Self::get_contract_key(&EthConnectorStorageId::FungibleToken),
             &self.ft,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -637,6 +637,43 @@ mod contract {
         );
     }
 
+    /// Due to the design change to stop minting and burning bridged ETH tokens
+    /// (see https://github.com/aurora-is-near/aurora-engine/pull/133 ),
+    /// there is currently an incorrect number of tokens in the Aurora NEP-141 token
+    /// account. This issue only impacts testnet because at the time the design was changed
+    /// the eth-connector had not been deployed to mainnet. The purpose of this function is
+    /// to mint the correct number of tokens in order to make up for the ones that were burned
+    /// before the design change in #133.
+    #[cfg(feature = "testnet")]
+    #[no_mangle]
+    pub extern "C" fn balance_evm_and_nep_141() {
+        use crate::precompiles::native::{ExitToEthereum, ExitToNear};
+        use crate::prelude::String;
+
+        let mut connector = EthConnectorContract::get_instance();
+        let aurora_account = unsafe { String::from_utf8_unchecked(sdk::current_account_id()) };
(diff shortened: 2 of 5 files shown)


src/connector.rs
```diff
@@ -26,7 +26,7 @@ pub(crate) const PAUSE_WITHDRAW: PausedMask = 1 << 1;
 #[derive(BorshSerialize, BorshDeserialize)]
 pub struct EthConnectorContract {
     contract: EthConnector,
-    ft: FungibleToken,
+    pub ft: FungibleToken,
```
joshuajbouw, line 29: Should this be encapsulated?

src/lib.rs
```diff
@@ -637,6 +637,43 @@ mod contract {
...
+    /// to mint the correct number of tokens in order to make up for the ones that were burned
+    /// before the design change in #133.
+    #[cfg(feature = "testnet")]
+    #[no_mangle]
+    pub extern "C" fn balance_evm_and_nep_141() {
+        use crate::precompiles::native::{ExitToEthereum, ExitToNear};
+        use crate::prelude::String;
+
+        let mut connector = EthConnectorContract::get_instance();
+        let aurora_account = unsafe { String::from_utf8_unchecked(sdk::current_account_id()) };
```
joshuajbouw, line 654: Even though that we will know that this is fine, it doesn't help that I don't like unsafe code randomly in a method like this. I would do this to the `current_account_id` method itself.

src/precompiles/native.rs
```diff
@@ -46,7 +46,7 @@ impl ExitToNear {
     ///
     /// Address: `0xe9217bc70b7ed1f598ddd3199e80b093fa71124f`
     /// This address is computed as: `&keccak("exitToNear")[12..]`
-    pub(super) const ADDRESS: Address =
+    pub(crate) const ADDRESS: Address =
         super::make_address(0xe9217bc7, 0x0b7ed1f598ddd3199e80b093fa71124f);
```
joshuajbouw, line 50: Second thought, this whole method should live elsewhere. Not part of precompiles mod.

src/precompiles/native.rs
```diff
@@ -188,7 +188,7 @@ impl ExitToEthereum {
     ///
     /// Address: `0xb0bd02f6a392af548bdf1cfaee5dfa0eefcc8eab`
     /// This address is computed as: `&keccak("exitToEthereum")[12..]`
-    pub(super) const ADDRESS: Address =
+    pub(crate) const ADDRESS: Address =
         super::make_address(0xb0bd02f6, 0xa392af548bdf1cfaee5dfa0eefcc8eab);
```
joshuajbouw, line 192: As with this.

joshuajbouw (changes requested):

birchmd:
@joshuajbouw per my description in the PR:

> This new method only needs to be called one time on testnet, so after this has happened we can revert this PR to avoid confusion in the future.

So I don't think we should worry about small details like what should be `pub` or where code lives. The point of this was to enable a minimal fix for the on-going testnet issue; none of these changes should be considered permanent.

sept-en (commented):
Thanks @birchmd !

PR #227, opened 2021-08-11
Fix promise, broken in upstream

joshuajbouw:
As per https://github.com/rust-blockchain/evm/commit/a81c4dfabc85f96e0f6823334bd32c34f070d099, much-needed logic was removed as there were complications for other teams with multiple memory states. This is an attempt to remedy the situation to instead return the data that needs to be done through the promise as a `LOG` return.


src/engine.rs
```diff
@@ -613,12 +620,19 @@ impl Engine {
             }
         };
 
-        let (values, logs, promises) = executor.into_state().deconstruct();
+        let (values, logs) = executor.into_state().deconstruct();
+        for log in logs {
```
birchmd, line 624: So this is the same logging mechanism that any smart contract on the EVM can use? This is not secure since it means any contract could create a log with the right topics to create a promise without having sent ETH to the precompile address.

birchmd (changes requested):

src/engine.rs
birchmd: As discussed with @mfornet , we can securely determine which logs are created by our precompiles by using the `address` field instead of `topics`.

src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![feature(array_methods)]
+#![feature(map_into_keys_values)]
```
birchmd, line 2: Not sure what is going on here. Locally when I run `make check` it says I do not need this because the feature is stable. But CI seems to think it is needed, and even when I add it here [it still fails](https://github.com/aurora-is-near/aurora-engine/pull/227/checks?check_run_id=3314455527).

src/precompiles/native.rs
```diff
@@ -65,34 +59,29 @@ impl Precompile for ExitToNear {
...
-        if Self::required_gas(input)? > target_gas {
-            return Err(ExitError::OutOfGas);
+    fn run(input: &[u8], target_gas: Option<u64>, context: &Context) -> EvmPrecompileResult {
+        if let Some(target_gas) = target_gas {
+            if Self::required_gas(input)? > target_gas {
+                return Err(ExitError::OutOfGas);
+            }
         }
 
+        // TODO: How to fix this?
```
birchmd, line 80: Still need to do something about this. Maybe there is a way from the `context` to determine if this is a static call or not?

src/precompiles/native.rs
```diff
@@ -198,34 +195,29 @@ impl Precompile for ExitToEthereum {
...
-        if Self::required_gas(input)? > target_gas {
-            return Err(ExitError::OutOfGas);
+    fn run(input: &[u8], target_gas: Option<u64>, context: &Context) -> EvmPrecompileResult {
+        if let Some(target_gas) = target_gas {
+            if Self::required_gas(input)? > target_gas {
+                return Err(ExitError::OutOfGas);
+            }
         }
 
+        // TODO: how to fix this?
```
birchmd, line 216: Same here

src/precompiles/native.rs
mfornet: What if we move the is_static field upstream into context directly. My guess is that it won't break any code for others, and will do the trick for us.

src/lib.rs
joshuajbouw: `rustup update` it was recently made stable. If CI is failing then it is building with an older rust v.

src/precompiles/native.rs
joshuajbouw: Thats the idea at least, to compromise and keep `is_static` upstream. It will break API, but won't really hurt anyone.

src/lib.rs
joshuajbouw: Ok, @strokovok and I worked together to resolve this.

joshuajbouw:
This should be good to go now.

joshuajbouw:
Issues brought on by develop branch now fixed.

src/precompiles/mod.rs
```diff
@@ -78,7 +70,12 @@ pub trait Precompile {
     fn required_gas(input: &[u8]) -> Result<u64, ExitError>;
 
     /// Runs the precompile function.
-    fn run(input: &[u8], target_gas: u64, context: &Context, is_static: bool) -> PrecompileResult;
+    fn run(
+        input: &[u8],
+        target_gas: Option<u64>,
+        context: &Context,
+        is_static: bool,
+    ) -> EvmPrecompileResult;
```
sept-en, line 78: Why do we make `target_gas` optional? I see that this is present in the upstream as well, but what's the purpose?

sept-en (approved):
LGTM!

src/precompiles/mod.rs
joshuajbouw: I've never figured that out to be fair. That could change in the future. For now, its just always something.

src/precompiles/mod.rs
```diff
@@ -128,8 +122,12 @@ impl Precompiles {
             ExitToNear::run,
             ExitToEthereum::run,
         ];
+        let mut map = BTreeMap::new();
+        for (address, fun) in addresses.into_iter().zip(fun) {
+            map.insert(address, fun);
+        }
```
birchmd, line 128: I suspect the following would work (and be a couple lines shorter):

```rust
let map = addresses.into_iter().zip(fun).collect();
```

src/precompiles/native.rs
```diff
@@ -46,7 +40,7 @@ impl ExitToNear {
     ///
     /// Address: `0xe9217bc70b7ed1f598ddd3199e80b093fa71124f`
     /// This address is computed as: `&keccak("exitToNear")[12..]`
-    pub(super) const ADDRESS: Address =
+    pub(crate) const ADDRESS: Address =
```
birchmd, line 43: Is this change needed?

birchmd (approved):
Nothing blocking, just a couple minor comments

PR #232, opened 2021-08-13
ETH-tests: state engine

mrLSD:
Related to issue: #170 

Basic source tests for SpuntikVM: [evm-tests](https://github.com/rust-blockchain/evm-tests)

Cover only: state tests


src/tests/eth/istanbul_builtins.json
```diff
@@ -0,0 +1,52 @@
...
+    "name": "alt_bn128_pairing",
+    "pricing": {
+      "0": {
+        "price": { "alt_bn128_pairing": { "base": 100000, "pair": 80000 }}
+      },
+      "0": {
+        "info": "EIP 1108 transition",
+        "price": { "alt_bn128_pairing": { "base": 45000, "pair": 34000 }}
+      }
+    }
```
joshuajbouw, line 41: I think these should be `"0":` and then `"1:"`. This was taken from EVM tests, right? It is a current issue there that hasn't been resolved yet.

joshuajbouw (changes requested):

src/tests/eth/istanbul_builtins.json
joshuajbouw: See: https://github.com/rust-blockchain/evm-tests/issues/3

mrLSD: I don't use that `json file`.
I used aurora precompiles "backend".

mrLSD: @jakestutzman also I can't run tests on upstream. I told it before.
So I can't compare. ((

joshuajbouw: Ok then just delete the file.

joshuajbouw (approved):

joshuajbouw:
These unfortunately can not be merged into this library due to conflicting license from sourced code. Must target a new library entirely. `aurora-engine-tests` will be fine.

PR #233, opened 2021-08-13
Fix(blockhash): Move block height to the end of hashing data

birchmd:
The reason for this change is to enable future micro-optimizations. We could cache the sha256 state which has all the static data already included in the hasher, then each hash computation only require pushing the block height and finalizing (rather than always needing to update the sha256 state with all the static data each time).


artob (approved):

joshuajbouw (approved):
Interesting idea, and makes sense to me.

joshuajbouw:
Though, on second thought I'm unsure if this is quite as much of an optimisation as we would like. i.e, if we made a constant of the first sections of data, it is still copying it to a new allocation.

sept-en (commented):
Definitely makes sense at least for the consistency (for possible future upgrades/improvements)!

PR #234, opened 2021-08-13
Fix(CI): Ensure solidity artifacts are always recompiled

birchmd:
#233 Changed a solidity contract, but the test failed in CI because it did not get recompiled (the compilation step does not happen when the artifact already exists). This ensures solidity artifacts are always recompiled.

An optimization to this would be to check if a `.sol` file was actually changed in the diff and only recompile if there is one.


joshuajbouw (approved):

joshuajbouw:
Yeah, this is a better idea.

PR #241, opened 2021-08-18
Fix return format of `ft_balance_of`

joshuajbouw:
There was a bit of confusion with the return required for `ft_balance_of`. It was initially interpreted that it needed simply to return the value as a string, but in fact, it needed to be returned surrounded by quotations.


mfornet:
Related to #191

mfornet (approved):

mrLSD (approved):

sept-en (commented):
LGTM!

PR #244, opened 2021-08-19
Add initial EIP-1559 support

joshuajbouw:
Adds initial [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md) support for the London hardfork.

There has been numerous changes done in this PR. Notably, I made a geth-like `NormalizedEthTransaction` which given the current fork will use that structure. This is important as for the first time instead of adding fields, 1 was taken away, and 2 additional were added so it led to a bit of a large redesign.

### DIFF
--- a/engine-types/src/types.rs
+++ b/engine-types/src/types.rs
@@ -45,6 +45,7 @@ pub fn validate_eth_address(address: String) -> Result<EthAddress, ValidationErr
 /// Newtype to distinguish balances (denominated in Wei) from other U256 types.
 #[derive(Debug, Eq, PartialEq, PartialOrd, Ord, Copy, Clone, Default)]
 pub struct Wei(U256);
+
 impl Wei {
     const ETH_TO_WEI: U256 = U256([1_000_000_000_000_000_000, 0, 0, 0]);
 
@@ -89,6 +90,7 @@ impl Wei {
         self.0.checked_add(other.0).map(Self)
     }
 }
+
 impl Sub for Wei {
     type Output = Self;
 
@@ -96,6 +98,7 @@ impl Sub for Wei {
         Self(self.0 - other.0)
     }
 }
+
 impl Add for Wei {
     type Output = Self;
 
--- a/engine/src/transaction/eip_1559.rs
+++ b/engine/src/transaction/eip_1559.rs
@@ -1,34 +1,23 @@
 use crate::prelude::precompiles::secp256k1::ecrecover;
-use crate::prelude::{sdk, Address, Vec, Wei, H256, U256};
+use crate::prelude::{Vec, U256};
+use crate::transaction::eip_2930::AccessTuple;
+use aurora_engine_types::types::Wei;
+use ethabi::Address;
 use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
 
-/// Type indicator (per EIP-2718) for access list transactions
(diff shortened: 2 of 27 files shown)


joshuajbouw:
Unblocked as per https://github.com/aurora-is-near/aurora-engine/pull/280

engine-types/src/types.rs
```diff
@@ -89,14 +90,20 @@ impl Wei {
     pub fn checked_add(self, other: Self) -> Option<Self> {
         self.0.checked_add(other.0).map(Self)
     }
+
+    pub fn checked_mul(self, other: Self) -> Option<Self> {
```
birchmd, line 94: Multiplying units of tokens (Wei in this case) does not make sense. The result would have units of `Money^2` which is meaningless (there is not a reasonable notion of "money area" as far as I know).

birchmd: Fixed in https://github.com/aurora-is-near/aurora-engine/pull/244/commits/08be2514183780684699a6fc27d8cbb25f58598e

engine/src/transaction/eip_1559.rs
```diff
@@ -0,0 +1,129 @@
+use crate::prelude::{Vec, U256};
+use crate::transaction::eip_2930::AccessTuple;
+use aurora_engine_types::types::Wei;
+use ethabi::Address;
+use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
+
+/// Type indicator (per EIP-1559)
+pub const TYPE_BYTE: u8 = 0x01;
```
birchmd, line 8: Should be `0x02`?

engine/src/transaction/mod.rs
```diff
@@ -1,107 +1,133 @@
...
-    }
+impl TryFrom<&[u8]> for EthTransactionKind {
+    type Error = ParseTransactionError;
 
-    pub fn gas_limit(&self) -> U256 {
-        match self {
-            Self::Legacy(tx) => tx.transaction.gas,
-            Self::AccessList(tx) => tx.transaction_data.gas_limit,
+    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+        if bytes[0] == eip_2930::TYPE_BYTE {
```
birchmd, line 23: Will need to add a case for `eip_1559::TYPE_BYTE`

engine/src/transaction/mod.rs
```diff
@@ -1,107 +1,133 @@
...
-                    tx.access_list,
-                )
-            }
+impl From<EthTransactionKind> for NormalizedEthTransaction {
+    fn from(kind: EthTransactionKind) -> Self {
+        use EthTransactionKind::*;
+        match kind {
+            Legacy(tx) => Self {
+                address: tx.sender(),
+                chain_id: None, // This should be fine as it'll fail regardless.
```
birchmd, line 58: This looks wrong. Why not properly derive the `chain_id` from the `v` field?

birchmd (changes requested):

engine/src/transaction/eip_1559.rs
joshuajbouw: Would need to double check, I do recall pulling that off of source.

engine/src/transaction/mod.rs
joshuajbouw: Yeah that should be updated.

PR #246, opened 2021-08-20
Return view of transaction result as borsh

joshuajbouw:
This fixes an issue where the public `view` call was returning the result by name, but not the actual Borsh serialization. This fixes that.

**BREAKING**

Unfortunately, this breaks public API but for the better. The status return must be serialised to the equivalent of:
```rust
pub enum TransactionStatus {
    Succeed(Vec<u8>),
    Revert(Vec<u8>),
    OutOfGas,
    OutOfFund,
    OutOfOffset,
    CallTooDeep,
}
```

In a follow-up PR, a tests ensuring that we are not breaking API must be made.


mrLSD (approved):

sept-en (commented):
LGTM!

PR #249, opened 2021-08-20
Release 1.6.1

joshuajbouw:
This includes both a cherry-pick of #241 and the fix of #246 for the 1.6.1 hotfix.

Just to fix the CI, #240 was also cherry-picked but has nothing to do with the functions of the codebase.


sept-en (commented):
@joshuajbouw let's also introduce a fix to #237 (the issue raised in the comments)

joshuajbouw:
Good call I'll cherry pick that when I get back.

sept-en:
@joshuajbouw Yeah, I noticed that the fix was introduced directly in #235 (c815108dd commit). So we're good to go.

src/tests/eth_connector.rs
```diff
@@ -173,10 +172,9 @@ fn get_eth_on_near_balance(master_account: &UserAccount, acc: &str, contract: &s
         "ft_balance_of",
         json!({ "account_id": acc }).to_string().as_bytes(),
     );
-    String::from_utf8(balance.unwrap())
-        .unwrap()
-        .parse()
-        .unwrap()
+    let val_str = String::from_utf8(balance.unwrap()).unwrap();
+    let val = &val_str[1..val_str.len() - 1];
+    val.parse().unwrap()
```
sept-en, line 177: Btw, this is, of course, the scope of the #241 PR, but I think that things like that need to be wrapped up into another function. Otherwise, it'll be hard both for us and external contributors to get what's the reason why we're doing such weird slices

sept-en (approved):
LGTM!

src/tests/eth_connector.rs
joshuajbouw: Thats because of a weird quirk of Rust. It wouldn't let me actually remove the quotations through usual methods. I have no answer for why, but that was the only method through 5 others that I tried that actually worked in the end.

artob:
@joshuajbouw And now please merge `master` back into `develop`, fixing the merge conflicts.

PR #256, opened 2021-09-02
Split tests to new crate

mrLSD:
We should move tests to a new crate and change LICENSE for it to `GPL-v3`
`engins` crates shouldn't depend on that crate, because the license: `CCO-1.0`


mrLSD:
Thanks, @birchmd for the advice!

engine-tests/src/tests/precompiles/mod.rs
```diff
@@ -0,0 +1 @@
+
```
birchmd, line 1: Empty file?

engine/src/precompiles/blake2.rs
```diff
@@ -95,8 +95,8 @@ impl Precompile for Blake2F {
 
 #[cfg(test)]
 mod tests {
+    use super::super::utils::new_context;
```
birchmd, line 98: `super::super` seems hard to follow; probably better to go down from the root instead of up from here (same comment applies to all other places where `super::super` is used)

engine-tests/src/tests/eth_connector.rs
```diff
@@ -29,6 +28,18 @@ const EVM_CUSTODIAN_ADDRESS: &'static str = "096DE9C2B8A5B8c22cEe3289B101f6960d6
 const DEPOSITED_EVM_AMOUNT: u128 = 10200;
 const DEPOSITED_EVM_FEE: u128 = 200;
 
+#[derive(
+    Default, BorshDeserialize, BorshSerialize, Clone, serde::Deserialize, serde::Serialize,
+)]
+pub struct Proof {
```
birchmd, line 34: Previously `Proof` was imported. Is this duplicating the struct now? I think it is better to avoid duplication.

engine-tests/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+mod prelude {
+    pub use aurora_engine_types::types::*;
+    pub use aurora_engine_types::*;
+}
```
birchmd, line 4: I still don't like this `prelude` importing all the things pattern.

Previously we had a prelude to handle std vs no-std imports more cleanly. That made sense to me. However I do not see the value in this sort of prelude, and indeed see it as a potential source of confusion when there are multiple ways to import the same thing.

birchmd (commented):
Generally looks good to me. Again, only style nits. I'll let @joshuajbouw make the final call on all these things.

engine/src/precompiles/blake2.rs
mrLSD: It was fixed at #260

engine-tests/src/tests/eth_connector.rs
mrLSD: It's another interesting issue:
```
#[derive(Default, BorshDeserialize, BorshSerialize, Clone)]
#[cfg_attr(test, derive(serde::Deserialize, serde::Serialize))]
pub struct Proof {
```
So after moving to a new crate:
`#[cfg_attr(test, derive(serde::Deserialize, serde::Serialize))]`
doesn't work. Thay why  code duplicated.

birchmd: Ah, I see. So the `serde` serialization is the reason. We do not use this in production, but we do use this in the tests.

As far as I remember the reason for this is simply convenience, so we could refactor the tests to not need `serde`. Perhaps that is out of scope for this PR though.

joshuajbouw:
Lets merge this one: https://github.com/aurora-is-near/aurora-engine/pull/260 into here, fix conflict, and good to go.

PR #259, opened 2021-09-03
Chore: Upgrade nearcore dependencies to latest commit

birchmd:
There is an ongoing effort in the nearcore team to improve the runtime performance and this will ultimately lead to lower gas parameters; allowing us to fit more Ethereum gas per NEAR gas. In particular, preliminary results based on changes in https://github.com/near/nearcore/pull/4688 indicate that PR could increase the maximum Ethereum gas we can fit in one NEAR transaction by 7x.

In anticipation of that PR being merged in nearcore, we are upgrading our dependencies on nearcore to the latest commit. This will allow us to easily pull in new runtime changes to our benchmarks and thus measure their improvement.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -120,7 +120,7 @@ dependencies = [
  "aurora-bn",
  "base64 0.13.0",
  "blake2 0.9.1 (git+https://github.com/near/near-blake2.git)",
- "borsh",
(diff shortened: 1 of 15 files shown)


joshuajbouw (approved):
Looks good. Nice code quality improvements upstream on Nearcore!

PR #274, opened 2021-09-15
Test for 1inch liquidity protocol

birchmd:
This PR introduces a test which performs the following steps:

- Deploy 1inch liquidity protocol contracts
- Create a new liquidity pool
- Deposit liquidity
- Swap tokens
- Withdraw liquidity

The key result is that all of this worked within the NEAR gas limits (including one transaction which used over 5 million EVM gas).


joshuajbouw (approved):
Birch uses test!

It's super effective!

musecrypto:
Does this case still pass? <1M gas txs are reverting on mainnet because of the gas limit.

birchmd:
> Does this case still pass? <1M gas txs are reverting on mainnet because of the gas limit.

Yes, these tests still pass. They are always running in our CI.

That said, when I was recently investigating gas profiles of transactions executing on mainnet I did find there was a large part of the cost which is related to the details of the NEAR node trie, and currently not reflected in our tests. This may explain the reduced performance of mainnet relative to these tests (though without looking at your specific example I cannot say for sure).

That said, there is another wasm cost reduction coming to mainnet (I think in the next NEAR release). Hopefully this will alleviate the issues you are having. We are also working actively with NEAR on the trie costs, though the situation is complicated and it may be a while before we see improvement in that area.

PR #277, opened 2021-09-16
Fix(deploy benchmark): Realistic deploy transaction

birchmd:
The original benchmark did not attempt to make the code to deploy a valid EVM program. For some reason this still returned successfully and burned lots of gas, so I didn't think anything of it. However, my recent work with the 1inch contract showed that deploying a real contract was possible even if it consumed a lot of EVM gas. In this PR I make the deploy transactions used in the benchmark more realistic (actually writing some bytes into the code field of the created account).

As a result of this change we can establish a more accurate number for the EVM gas limit we can support within the 200 Tgas limit on NEAR. Based on fitting a line to the data point from this new benchmark (see plot below), I conclude we can support deploy transactions which consume more than 20 million EVM gas.

Note: this does not tell us anything about the gas limit for compute-heavy transactions. We still need performance improvements from NEAR to execute some UniswapV3 transactions for example.

deploy_benchamrk

### DIFF
--- a/src/benches/eth_deploy_code.rs
+++ b/src/benches/eth_deploy_code.rs
@@ -1,12 +1,13 @@
 use criterion::{BatchSize, BenchmarkId, Criterion, Throughput};
 use secp256k1::SecretKey;
 
-use crate::test_utils::{address_from_secret_key, create_eth_transaction, deploy_evm, SUBMIT};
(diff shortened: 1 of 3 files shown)


joshuajbouw (approved):
This is great. Thanks for fixing this up based on what we had observed.

PR #279, opened 2021-09-17
Fix json formatting in JsonValue serialization

birchmd:
Fixes #278 

See https://github.com/aurora-is-near/aurora-engine/issues/278#issuecomment-921809411 for details.

PR #285, opened 2021-09-24
Release 1.6.4

birchmd:
- `ft_metadata` bug fix (#279) 
- Spit code into separate crates for license compliance (#243) 
- Fix timestamp bug (#287)


mrLSD:
@birchmd  From a source code perspective, it's can be version 1.6.4. It should be 1.7.0 - because of breaking changes in source code.

birchmd:
As I understand it, our product is the (wasm) contract we deploy to Near and this is what the version refers to. Source code changes (even large ones like spitting into multiple crates) are not observable from the contract perspective. The only important change is the `ft_metadata` bug fix, which is why I chose 1.6.4 as the version.

joshuajbouw (approved):
This looks good to me. There are some small nits that I do notice such as the libs imo shouldn't be prefixed with `engine-` but, thats ok for now. Everything else and running it on my end looks good.

To be fair, we should be bumping every major version everytime we break public API. We are still technically beta, although the version says otherwise. Bumping minor in this case should be fine then.

mrLSD (approved):

sept-en (approved):

artob:
> [O]ur product is the (wasm) contract we deploy to Near and this is what the version refers to [...]
> We should be bumping every major version every time we break public API.

Correct. We follow [semantic versioning](https://semver.org), and the versioning scheme is quite systematic and predictable.

PR #287, opened 2021-09-27
Fix block.timestamp

birchmd:
The timestamp is supposed to be in seconds per the [Solidity docs](https://docs.soliditylang.org/en/latest/units-and-global-variables.html?highlight=block#block-and-transaction-properties).

Thanks @0x3bfc for reporting this!

### DIFF
--- a/engine-tests/src/tests/one_inch.rs
+++ b/engine-tests/src/tests/one_inch.rs
@@ -50,6 +50,9 @@ fn test_1inch_liquidity_protocol() {
     helper.approve_erc20_tokens(&token_a, pool.address());
     helper.approve_erc20_tokens(&token_b, pool.address());
 
+    // I don't understand why this is needed but for some reason the 1inch
+    // contract divides by zero unless I mess with the time.
+    helper.runner.context.block_timestamp += 10_000_000 * 1_000_000_000;
     let (result, profile) = helper.pool_deposit(
         &pool,
         liquidity_protocol::DepositArgs {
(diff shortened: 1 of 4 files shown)


engine-tests/src/tests/one_inch.rs
```diff
@@ -98,8 +103,8 @@ fn test_1_inch_limit_order_deploy() {
         SubmitResult::try_from_slice(&outcome.return_data.as_value().unwrap()).unwrap();
     assert!(result.status.is_ok());
 
-    // more than 4 million Ethereum gas used
-    assert!(result.gas_used > 4_000_000);
+    // more than 3.5 million Ethereum gas used
+    assert!(result.gas_used > 3_500_000);
```
birchmd, line 107: I don't understand why this change was needed; CI was failing on this assert, but I could not reproduce it locally. Maybe a difference in the solidity compiler versions? But I don't know why this just started failing now.

artob (approved):

engine-tests/src/tests/one_inch.rs
joshuajbouw: I'll verify.

joshuajbouw (approved):

joshuajbouw:
While the CI works online, I'm having trouble getting this to work locally as well. Likely cache related. Digging into this.

engine-tests/src/tests/one_inch.rs
birchmd: I can reproduce the issue locally if I totally clear the `target` folder, forcing a re-download of the 1inch source code. Looks like they [merged a PR yesterday](https://github.com/1inch/limit-order-protocol/pull/47) to lower their EVM gas usage, so it makes sense that we need to lower this now.

I think we should merge this for now, but we can think about pinning a specific commit of the 1inch source code for this test in a future PR.

PR #288, opened 2021-09-28
Feat(exit precompile): Events for exit precompile calls.

### LINKED ISSUE #245: feat: Emit Aurora event in ETH exit precompile.
In order to improve the bridge UI experience, we need to query for all the transfers initiated by an account on Aurora.
When withdrawing a bridged ERC-20, `withdrawToEthereum` or `withdrawToNear` are called and emit a Transfer event when `_burn` is executed: https://github.com/aurora-is-near/aurora-engine/blob/bb21459ac5560d89f56024a6db2d29d955ca973e/etc/eth-contracts/contracts/EvmErc20.sol#L64 We can then query and filter all those events to find the transfers.

However when withdrawing ETH, no event is emitted. Would it be possible to also emit an exit event for ETH from the precompile ?

birchmd:
Closes #245 (see that ticket for details)

@paouvrard mentioned that an explicit event for ERC-20 exit, was not required, but it seemed easy to me to include it now, so I did.

This PR includes new tests showing the events are created.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -188,6 +188,7 @@ name = "aurora-engine-tests"
 version = "1.0.0"
 dependencies = [
  "aurora-engine",
+ "aurora-engine-precompiles",
  "aurora-engine-sdk",
  "aurora-engine-types",
  "borsh 0.8.2",
--- a/engine-precompiles/src/lib.rs
+++ b/engine-precompiles/src/lib.rs
@@ -206,7 +206,7 @@ impl Precompiles {
 /// const fn for making an address by concatenating the bytes from two given numbers,
 /// Note that 32 + 128 = 160 = 20 bytes (the length of an address). This function is used
 /// as a convenience for specifying the addresses of the various precompiles.
-const fn make_address(x: u32, y: u128) -> prelude::Address {
+pub const fn make_address(x: u32, y: u128) -> prelude::Address {
     let x_bytes = x.to_be_bytes();
     let y_bytes = y.to_be_bytes();
     prelude::Address([
--- a/engine-precompiles/src/lib.rs
+++ b/engine-precompiles/src/lib.rs
@@ -233,6 +233,45 @@ const fn make_address(x: u32, y: u128) -> prelude::Address {
     ])
 }
 
+const fn make_h256(x: u128, y: u128) -> prelude::H256 {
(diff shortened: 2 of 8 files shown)


sept-en (approved):
Great job! Thanks @birchmd !

paouvrard (approved):

mrLSD (approved):

paouvrard:
@birchmd This should work for the frontend but I realized it would also be useful to filter by sender and token. Is it possible to include the `sender` and the `erc20 address` in the event ? 
That would also make it consistent with the Locked event in rainbow bridge: https://github.com/aurora-is-near/rainbow-token-connector/blob/134b3b6ad900d90c6f993c52f4e3edf985aec61a/erc20-connector/contracts/ERC20Locker.sol#L15

mfornet (changes requested):
LGTM

However I agree with @paouvrard about using the extended version of the event. 
Marking `Request changes` to avoid overlooking that!

birchmd:
@mfornet @paouvrard

I don't think we can get the sender to include in the event. The issue is that the sender is two steps back in the call stack, not the previous step as with the lockup. This is because the flow is `sender -> erc20.withdraw -> exit_precompile`. So at the level of the exit precompile, `msg.sender` returns the address of the ERC-20 token on Aurora, not the user/contract which made the withdraw call.

I can include the ERC-20 address though. I will replace the `bool indexed is_erc20` in the current event with `Address indexed erc20_address`, and in the ETH withdraw case I will set this equal `0x00`. Does that work for you?

mfornet (approved):
> I don't think we can get the sender to include in the event. The issue is that the sender is two steps back in the call stack, not the previous step as with the lockup.

I see

paouvrard:
Understood, thanks @birchmd ! It shouldn't be a problem for erc20 because we can always filter with the burn event to get the sender.

But for exits of `ETH`, the `sender` calls the `exit_precompile` directly so it should be possible include it in this case right ?

joshuajbouw (approved):
Looks great. No concerns here. Code quality and effort is excellent. Thanks.

joshuajbouw:
Just FYI, these should only be staged to testnet first, followed by mainnet in next release.

birchmd:
@paouvrard Yes, good point. Followed up in #292

Issue #180, opened 2021-07-14
Ensure all usages of `AccountId` are valid IDs


sept-en:
We use `AccountId`s in many places of the project. To eliminate possible mistakes in the `AccountId` itself, we need to perform the check on each such usage by calling `is_valid_account_id()`.

PR #289, opened 2021-09-30
Validate account id everywhere

mrLSD:
Check and validate Account ID everywhere - `engine` and other crates.
The main reason to be sure that Account ID is everywhere in the source code correct.

Related issue: #180


birchmd:
Please rebase on `develop` and remove extraneous commits.

birchmd:
> Merge branch 'develop' into feat/validate-account-id

Merge is not good enough here, needs to be a [rebase](https://git-scm.com/docs/git-rebase).

birchmd (changes requested):
I think a cleaner way to do this would be to define an `AccountId` type with a constructor that validates the input. Then anywhere we use something that is supposed to be an `AccountId` we declare in the type signature that is must be this validated type (as opposed to `String` or `&str` or `&[u8]`). This is better because then we do not have `assert` statements littered throughout the code and in the future we do not need to remember to include more such assert statements when we add new functions that require account IDs (the validation will be automatic in the type).

This approach was taken recently by `nearcore` as well (https://github.com/near/nearcore/pull/4621).

mrLSD:
> I think a cleaner way to do this would be to define an `AccountId` type with a constructor that validates the input. Then anywhere we use something that is supposed to be an `AccountId` we declare in the type signature that is must be this validated type (as opposed to `String` or `&str` or `&[u8]`). This is better because then we do not have `assert` statements littered throughout the code and in the future we do not need to remember to include more such assert statements when we add new functions that require account IDs (the validation will be automatic in the type).

Initial issue #180  has a different proposal. Current PR solves it.
I know about a well-known approach in NEAR as `ValidAccountId` data structure. My opinion `AccountId` should be `String`,  but `ValidAccountId` as `ValidAccountId(String)` that indicate about valid date. As the next step, we should introduce it. And that should be implemented as new PR, because it solves another problem and much more bigger for changes.

birchmd:
> Initial issue #180 has a different proposal.

The issue says "we need to perform the check on each such usage" does not necessarily mean literally putting an `assert` at each point in the code. Calling `is_valid_account_id()` in a validated type constructor and using that type everywhere `AccountId` is currently used also performs the check on each usage.

As a side point, I do not think that 1 sentence issues should be considered design proposals for an implementation. Short issues like that only convey the problem, possibly with some hint at what the solution could involve. It is up to the person implementing the solution to make sure that what they implement is high quality and maintainable code. And in some cases, it might even make sense to write a more detailed design proposal for the implementation in the issue before starting to write code. This can make the review process easier because we can agree on the high level approach first.

> My opinion `AccountId` should be `String`, but `ValidAccountId` as `ValidAccountId(String)`

This implies there is some use-case for possibly invalid account IDs. I cannot think of one though. Do you have a suggestion as to where `AccountId` would be used as opposed to `ValidAccountId`? I think any business logic which uses account IDs should require the validated type, and any place we obtain external input we should always parse it into the validated type if it is meant to be interpreted as an account ID later.

mrLSD:
@birchmd for me, it's different things - new design of AccountId type, and cover valid account_id's in the code.

mrLSD:
AccountId as constructor self-validated is the best approach.
And I completely agree with that.
So let's decide - do we just close PR and starting implementation a new design for `AccountId` type.
Or merge it (because it covers uncovered places). And then starting implementation of new design (it will take almost all business logic, as it's basic type for `Aurora`).

birchmd:
I am personally of the opinion that we close this PR. The validated account ID type PR will end up remove all these asserts, so we double up on the work in some sense by merging this now. I also do not think this is a high risk issue that requires a quick fix be merged immediately because the issue has been open for months now.

mrLSD:
@joshuajbouw @birchmd I analyzed 
https://github.com/near/nearcore/tree/master/core/account-id
Strictly saying that solution is good enough for us.

PR #311, opened 2021-10-18
Refund user when ExitToNear promise fails

### LINKED ISSUE #269: Handle errors when sending tokens within Aurora to NEAR

birchmd:
Closes #269 

The solution we agreed on was the one suggested by @mfornet where we change the `ExitToNear` precompile API to accept an address to refund the assets to in case the exit promise fails. This is a breaking change, however, to allow this PR to be merged without breaking anything the API change is hidden behind a compile-time feature flag. @mfornet @sept-en and I will need to coordinate on migrating the existing bridged ERC-20 assets and enabling this feature. After this feature is enabled by default then we can clean up the code to remove the old logic.

This PR is relatively large, so guide the reviewer I draw your attention to the following pieces:

- The main logic change is of course in `engine-precompiles/src/native.rs`; the refund itself happens in the new function in `engine/src/lib.rs`
- I've added a few new tests to `engine-tests/src/tests/erc20_connector.rs`, however because these tests rely on resolving promises they could not use the `AuroraRunner` testing utility which existing tests used. Therefore, there is a great deal of boilerplate code added as well to wrap up `near-sdk-sim` calls. I recommend simply focusing on the tests themselves (in particular the assert statements show how the balances behave with and without this refund mechanism).
- The whole test suite passes with and without the new feature enabled. You can check this locally but running `make check` and `make check error-refund=yes`. As a sanity test you could also check locally that various tests fail if you do not consistently enable the feature (i.e. if you compile the contract with the new feature, but then run the tests without it or vice versa).
- There are two "new" solidity contracts, but really they are the old contracts with small modifications for the API change. Unfortunately I do not know if conditional compilation is a thing in solidity, so I had to duplicate the whole contract. Below is the diff

```
$ diff etc/eth-contracts/contracts/test/Tester.sol etc/eth-contracts/contracts/test/TesterV2.sol
6c6
< contract Tester {
---
> contract TesterV2 {
44,45c44,46
<         bytes memory input = abi.encodePacked("\x00", recipient);
<         uint input_size = 1 + recipient.length;
---
>         address sender = msg.sender;
>         bytes memory input = abi.encodePacked("\x00", sender, recipient);
>         uint input_size = 1 + 20 + recipient.length;


$ diff etc/eth-contracts/contracts/EvmErc20.sol etc/eth-contracts/contracts/EvmErc20V2.sol
15c15
< contract EvmErc20 is ERC20, AdminControlled, IExit {
---
> contract EvmErc20V2 is ERC20, AdminControlled, IExit {
52c52,53
<         _burn(_msgSender(), amount);
---
>         address sender = _msgSender();
>         _burn(sender, amount);
55,56c56,57
<         bytes memory input = abi.encodePacked("\x01", amount_b, recipient);
<         uint input_size = 1 + 32 + recipient.length;
---
>         bytes memory input = abi.encodePacked("\x01", sender, amount_b, recipient);
>         uint input_size = 1 + 20 + 32 + recipient.length;
```


engine-precompiles/src/native.rs
```diff
@@ -29,6 +34,9 @@ mod costs {
     // TODO(#51): Determine the correct amount of gas
     pub(super) const FT_TRANSFER_GAS: Gas = 100_000_000_000_000;
 
+    #[cfg(feature = "error_refund")]
+    pub(super) const REFUND_ON_ERROR_GAS: Gas = 60_000_000_000_000;
```
joshuajbouw, line 38: Is this an estimation?

engine-precompiles/src/native.rs
```diff
@@ -228,6 +236,16 @@ impl Precompile for ExitToNear {
         context: &Context,
         is_static: bool,
     ) -> EvmPrecompileResult {
+        #[cfg(feature = "error_refund")]
+        fn parse_input(input: &[u8]) -> (Address, &[u8]) {
+            let refund_address = Address::from_slice(&input[1..21]);
+            (refund_address, &input[21..])
```
joshuajbouw, line 242: Just FYI, its a good idea that we start setting the end byte len instead of implying them. May result in unwanted behavior.

joshuajbouw (commented):
Adding comments for now, I'm going to need a fresh start in the morning to go through this.

engine-precompiles/src/native.rs
joshuajbouw: Can you add a TODO message to accurately measure the amount of gas? Then add in an issue so that we can track it.

engine-precompiles/src/native.rs
```diff
@@ -242,9 +260,14 @@ impl Precompile for ExitToNear {
...
         //      0x1 -> Erc20 transfer
-        let mut input = input;
         let flag = input[0];
-        input = &input[1..];
+        #[cfg(feature = "error_refund")]
+        let (refund_address, mut input) = parse_input(input);
+        #[cfg(not(feature = "error_refund"))]
+        let mut input = parse_input(input);
+        let current_account_id = String::from_utf8(sdk::current_account_id()).unwrap();
+        #[cfg(feature = "error_refund")]
```
joshuajbouw, line 269: The `cfg!` macro will work much better here and generally should be done if its inside a method or function. Doesn't cause confusion for my IDE too.

engine-precompiles/src/native.rs
```diff
@@ -325,19 +348,46 @@ impl Precompile for ExitToNear {
...
+        } else {
+            Some(exit_event.erc20_address.0)
+        };
+        #[cfg(feature = "error_refund")]
+        let refund_args = RefundCallArgs {
+            recipient_address: refund_address.0,
+            erc20_address,
+            amount: types::u256_to_arr(&exit_event.amount),
+        };
+        #[cfg(feature = "error_refund")]
```
joshuajbouw, line 363: Same as above, `cfg!` macro is better here I believe.

engine/src/engine.rs
```diff
@@ -890,7 +896,27 @@ impl Engine {
...
             promise.attached_balance,
             promise.attached_gas,
+        )
+    }
+
+    fn schedule_promise_callback(base_id: u64, promise: PromiseCreateArgs) -> u64 {
+        #[cfg(feature = "log")]
+        sdk::log_utf8(
+            crate::prelude::format!(
+                "[CALLBACK] Call contract: {}.{}",
```
joshuajbouw, line 906: We should use a standard format for logs, much like we did in #323

engine/src/lib.rs
```diff
@@ -372,6 +375,9 @@ mod contract {
         let mut engine = Engine::new(predecessor_address()).sdk_unwrap();
 
         let erc20_admin_address = current_address();
+        #[cfg(feature = "error_refund")]
+        let erc20_contract = include_bytes!("../../etc/eth-contracts/res/EvmErc20V2.bin");
+        #[cfg(not(feature = "error_refund"))]
```
joshuajbouw, line 380: Again, `cfg!` here

engine/src/lib.rs
```diff
@@ -401,6 +407,71 @@ mod contract {
         // TODO: charge for storage
     }
 
+    /// Callback invoked by exit to NEAR precompile to handle potential
+    /// errors in the exit call.
+    #[no_mangle]
+    pub extern "C" fn refund_on_error() {
```
joshuajbouw, line 413: A little bit unfortunate here. We also need to be sure that this can't be abused in anyway. Is it possible to ensure that this can only be invoked by our systems?

engine/src/lib.rs
```diff
@@ -668,6 +739,35 @@ mod contract {
...
+            relayer_id: aurora_account_str,
+            fee: 0,
+            msg: None,
+        };
+        let verify_id = sdk::promise_create(
+            &aurora_account,
+            b"verify_log_entry",
+            &[],
+            0,
+            20_000_000_000_000,
```
joshuajbouw, line 761: Should be a const

engine/src/lib.rs
```diff
@@ -668,6 +739,35 @@ mod contract {
...
+        );
+        sdk::promise_then(
+            verify_id,
+            &aurora_account,
+            b"finish_deposit",
+            &args.try_to_vec().unwrap(),
+            0,
+            50_000_000_000_000,
```
joshuajbouw, line 769: Should be a const

joshuajbouw (changes requested):

engine-precompiles/src/native.rs
mfornet: In CLion and VSCode you can tell which features should be considered active and which should not, so you can show relevant parts you are reviewing or working with.

joshuajbouw: I know that, but thats not exactly the issue I'm having with CLion. The issue is that it thinks its always the former.

engine/src/lib.rs
birchmd: Literally the next line takes care of this.

```
sdk::assert_private_call();
```

engine-precompiles/src/native.rs
birchmd: I was under the impression that `cfg!` would be less efficient because it evaluates to a `bool` and both branches would exist in the code, whereas `#[cfg]` literally changes what code is shown to the compiler and therefore has zero runtime overhead.

I agree something using `cfg!` might be easier to read, but hopefully we won't have this `error_refund` feature around for very long. It can be made active by default once @mfornet and @sept-en do the necessary state migration, and then can be removed entirely after that.

joshuajbouw: Fair. Then lets leave it in for now, resolve the other point outs.

engine/src/lib.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/311/commits/567522d92292b4f4effbe6a40260fdbb7dc0d67f

engine/src/lib.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/311/commits/567522d92292b4f4effbe6a40260fdbb7dc0d67f

engine/src/engine.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/311/commits/2635bff43b7a28444565c6c79df0f01fcf78aed4

engine-precompiles/src/native.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/311/commits/5678aba5f81b8a8940af14bed1a8dc3382354bf0

engine/src/engine.rs
```diff
@@ -886,23 +892,35 @@ impl Engine {
...
-            crate::prelude::format!(
-                "Call contract: {}.{}",
-                promise.target_account_id,
-                promise.method
-            )
-            .as_bytes(),
-        );
+    fn schedule_promise(promise: PromiseCreateArgs) -> u64 {
+        sdk::log!(&crate::prelude::format!(
+            "Call contract: {}.{}",
```
birchmd, line 897: @mfornet there are tests that depend on the specific format of this log. Do we use it in production as well? If not I will change to harmonize with other logs (ie underscores in the label and only space (no colon) as the separator with the value).

mfornet: Let's change them to fit the guideline

mfornet (approved):
Thanks for PR, and sorry for late review. Everything is good here.

What is the plan ahead regarding upgrades to the ERC20 contracts 🤔 .  The two paths ahead I'm seeing are:

- Deploy the new contract on the old address. 
- Deploy the new binaries on different address; move all balances from one to the other; update the erc20-nep141 mapping.

The first one is easier, but I'm concerned about this breaking some assumptions about how this particular tokens were deployed. What do you think?

In any case we should plan ahead, and have this contracts to be upgradable. So in case we need to do an upgrade in the future it is easier. We can set the owner to be `aurora` account. This is not blocking this PR.

joshuajbouw (approved):

birchmd:
> Thanks for PR, and sorry for late review. Everything is good here.
> 
> What is the plan ahead regarding upgrades to the ERC20 contracts 🤔 . The two paths ahead I'm seeing are:
> 
> * Deploy the new contract on the old address.
> * Deploy the new binaries on different address; move all balances from one to the other; update the erc20-nep141 mapping.
> 
> The first one is easier, but I'm concerned about this breaking some assumptions about how this particular tokens were deployed. What do you think?
> 
> In any case we should plan ahead, and have this contracts to be upgradable. So in case we need to do an upgrade in the future it is easier. We can set the owner to be `aurora` account. This is not blocking this PR.

I'm not sure which is the better option. Perhaps @sept-en would like to weigh in.

PR #318, opened 2021-10-27
Introduce new Account ID data structure

mrLSD:
Introduce a new Account ID data structure.

## Motivation
A self-validated data structure that guarantees Account ID correctness everywhere.

## Changes
* Full code refactoring related to Account ID
* Test refactoring related to Account ID


joshuajbouw:
Please check tests

engine/src/connector.rs
```diff
@@ -134,16 +134,12 @@ impl EthConnectorContract {
...
-            assert!(
-                is_valid_account_id(account_id.as_bytes()),
-                "ERR_INVALID_ACCOUNT_ID"
-            );
-            TokenMessageData::Near(account_id.into())
+            TokenMessageData::Near(account_id)
         } else {
             TokenMessageData::Eth {
-                address: data[0].into(),
+                address: account_id,
```
birchmd, line 142: Is this right? This is supposed to be an eth address (encoded in hex) isn't it, not a valid NEAR account ID?

engine/src/parameters.rs
```diff
@@ -219,10 +221,6 @@ impl TryFrom<NEP141FtOnTransferArgs> for String {
     type Error = ParseError;
 
     fn try_from(value: NEP141FtOnTransferArgs) -> Result<Self, Self::Error> {
-        if !is_valid_account_id(value.sender_id.as_bytes()) {
```
birchmd, line 222: With this error condition removed(*), this conversion looks infallible, so should be `From` instead of `TryFrom`.

(*) And I agree it should be removed, the validation makes much more sense in the `TryFrom<JsonValue>` implementation, nice job there.

birchmd (changes requested):
Requesting changes because of my question on `TokenMessageData::Eth`. I think the address field there should stay `String`, not `AccountId`, please confirm if this is the case.

engine/src/connector.rs
mrLSD: It is used only for:
```
            // Deposit to Eth accounts
            // fee is being minted in the `ft_on_transfer` callback method
            TokenMessageData::Eth { address, message } => {
                // Transfer to self and then transfer ETH in `ft_on_transfer`
                // address - is NEAR account
                let transfer_data = TransferCallCallArgs {
                    receiver_id: address,
                    amount: event.amount.as_u128(),
                    memo: None,
                    msg: self.set_message_for_on_transfer(event.fee, message),
                }
```

And after that it used for example here:
```
        self.ft
            .ft_transfer_call(&args.receiver_id, args.amount, &args.memo, args.msg);
```
That field should be `AccountId` type. Not `String`

engine/src/parameters.rs
mrLSD: I agree. It was changed.

engine/src/connector.rs
birchmd: Seems like the field should be renamed if it is not actually an address.

mrLSD: Yes, it is receiver_id. I'll rename it.

birchmd (approved):
Nice work @mrLSD

joshuajbouw (approved):

PR #321, opened 2021-11-01
Fix: log full address of sender in submit call

birchmd:


joshuajbouw (approved):
Fixes issues we have currently with current display. Wrong use of display, over debug.

PR #322, opened 2021-11-01
Feat(engine): log storage writes from applying transactions

birchmd:


engine/src/engine.rs
```diff
@@ -1049,9 +1051,17 @@ impl ApplyBackend for Engine {
...
+                    writes_counter += 2; // 1 for nonce, 1 for balance
 
                     if let Some(code) = code {
-                        Engine::set_code(&address, &code)
+                        Engine::set_code(&address, &code);
+                        code_bytes_written = code.len();
+                        sdk::log!(crate::prelude::format!(
+                            "Address {:?} set code with {} bytes",
+                            address,
+                            code_bytes_written,
```
joshuajbouw: Should be written as two logs such as:
`written_bytes <address> <code_bytes_writte` that was its easier for a tool to parse.

engine/src/engine.rs
```diff
@@ -1079,14 +1090,28 @@ impl ApplyBackend for Engine {
...
+        #[allow(unused_variables)]
+        let total_bytes = 32 * writes_counter + code_bytes_written;
+        #[allow(unused_assignments)]
+        if code_bytes_written > 0 {
+            writes_counter += 1;
+        }
+        sdk::log!(
+            crate::prelude::format!("Total number of storage writes: {}", writes_counter).as_str()
+        );
+        sdk::log!(crate::prelude::format!("Total bytes written: {}", total_bytes).as_str());
```
joshuajbouw: Same as above

joshuajbouw (approved):

PR #326, opened 2021-11-03
Modify message verification inside ft_transfer_call for external receiver_id

mrLSD:
Currently, we have `NEP-141` compatibility restriction for `nETH`. The `ft_transfer_call is` not compatible with all kinds of message data because our implementation expects the following format of the message: `<near_account_id>:<32 bytes fee><20 bytes for ETH address receiving the fee>` . Having any other message provided or even an empty one (which is a popular option to use for exchanges) will result in a _panic_ and _transaction revert_.

There's a reason why we were doing that: we are making that verification early in order to prevent future panics in `ft_on_transfer` (where it's hard to recover if there's an error). But as the message itself will be used only inside the `ft_on_transfer` , meaning that the message is intended to be processed by Aurora.

As @sept-en suggested we can do this early verification of the message format and overflow checks only if the `receiver_id == aurora`. Because otherwise, it makes no sense to force all other contracts to have the same message format as we do for Aurora.


engine/src/connector.rs
```diff
@@ -492,21 +492,23 @@ impl EthConnectorContract {
...
-        assert!(amount_for_check.checked_add(args.amount).is_some());
+            // Additional check overflow before process `ft_on_transfer`
+            // But don't check overflow for relayer
+            // Note: It can't overflow because the total supply doesn't change during transfer.
+            let amount_for_check = self
+                .ft
+                .internal_unwrap_balance_of_eth_on_aurora(message_data.recipient);
+            assert!(amount_for_check.checked_add(args.amount).is_some());
+        }
         assert!(self
```
sept-en, line 512: I believe this statement should be also inside those parentheses. Because why do we need to check the balance of total ETH supply on Aurora if we are not going to increase it (if the receiver is other than `aurora`)?

sept-en (changes requested):

joshuajbouw (changes requested):
Please add tests to ensure that it is working as intended.

engine/src/connector.rs
mrLSD: Fixed

mrLSD:
When I added new test, I have got some strange case (it related to `ft_resolve_transfer`). I'm not sure is it correct:
```
    let balance = get_eth_on_near_balance(&master_account, "some-test-acc", CONTRACT_ACC);
    assert_eq!(balance, 2 * transfer_amount);

    let balance = get_eth_on_near_balance(&master_account, CONTRACT_ACC, CONTRACT_ACC);
    assert_eq!(balance, DEPOSITED_FEE - 2 * transfer_amount);
```
So looks like double-spent. I investigated it. It's from canonical implementation (from NEP-141) `ft_resolve_transfer`.

joshuajbouw:
Well, if there are any double spends, this shouldn't be intentional at all, even if it is from a reference implementation. This warrants more investigation to ensure whats going on here.

engine-tests/src/tests/eth_connector.rs
```diff
@@ -457,6 +457,100 @@ fn test_ft_transfer_call_eth() {
...
+        _ => panic!("should panice"),
+    }
+
+    // Sending to excternal receiver with empty message should be success
+    let transfer_amount = 22;
+    let res = contract.call(
+        CONTRACT_ACC.parse().unwrap(),
+        "ft_transfer_call",
+        json!({
+            "receiver_id": "some-test-acc",
```
sept-en, line 510: It's ok to test the call to `some-test-acc`, but as it obviously doesn't have a `ft_on_transfer` method, it makes sense to also test against some test contract that indeed has `ft_on_transfer` method (could be empty, or just an accounting one).

engine-tests/src/tests/eth_connector.rs
```diff
@@ -457,6 +457,100 @@ fn test_ft_transfer_call_eth() {
...
+        .to_string()
+        .as_bytes(),
+        DEFAULT_GAS,
+        1,
+    );
+    match res.outcome().status {
+        ExecutionStatus::Failure(_) => {}
+        _ => panic!("should panice"),
+    }
+
```
sept-en, line 503: Also please add a check that the balance hasn't been changed after the execution and transaction failure.

engine-tests/src/tests/eth_connector.rs
birchmd: From the [NEP-141 standard](https://nomicon.io/Standards/FungibleToken/Core.html):

```
// * The receiving contract must implement `ft_on_transfer` according to the
//   standard. If it does not, FT contract's `ft_resolve_transfer` MUST deal
//   with the resulting failed cross-contract call and roll back the transfer.
```

So if `some-test-acc` does not implement `ft_on_transfer` then the transfer should be undone, no happen twice! This looks like a bug in our FT implementation.

sept-en: Yes, that's what I'm talking about and why it's important to have both cases covered in tests. Perhaps on https://github.com/aurora-is-near/aurora-engine/blob/c5454c15c12f83034f82acd7849f91cc858f58f5/engine/src/fungible_token.rs#L274 we also need to check for `promise_results_count`?

engine-tests/src/tests/eth_connector.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/326/commits/28953c6f4b4bf43be5f5df56ba7bc778ec6877d4

engine-tests/src/tests/eth_connector.rs
birchmd: > makes sense to also test against some test contract that indeed has `ft_on_transfer` method

New test added in https://github.com/aurora-is-near/aurora-engine/pull/326/commits/28953c6f4b4bf43be5f5df56ba7bc778ec6877d4

> we also need to check for `promise_results_count`?

This [check is done `EthConnectorContract::ft_resolve_transfer`](https://github.com/aurora-is-near/aurora-engine/blob/c5454c15c12f83034f82acd7849f91cc858f58f5/engine/src/connector.rs#L471) so it does not need to repeated. The code could be refactored to make it clear that check is already done, but in the spirit of doing a minimal change for the issues at hand I think we should do it in a future PR.

birchmd (approved):

sept-en (approved):

joshuajbouw (approved):

PR #333, opened 2021-11-05
Test asserting we do not change the engine state binary format

birchmd:
The purpose of this test is to make sure that if we accidentally change the binary format of the `EngineState` then we will know about it. This is important because changing the state format will break the contract unless we do a state migration.

### DIFF
--- a/engine-tests/src/tests/sanity.rs
+++ b/engine-tests/src/tests/sanity.rs
@@ -16,6 +16,29 @@ const TRANSFER_AMOUNT: Wei = Wei::new_u64(123);
 const GAS_PRICE: u64 = 10;
 
 #[test]
+fn test_state_format() {
+    // The purpose of this test is to make sure that if we accidentally
+    // change the binary format of the `EngineState` then we will know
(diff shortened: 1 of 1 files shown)


engine-tests/src/tests/sanity.rs
```diff
@@ -15,6 +15,29 @@ const INITIAL_NONCE: u64 = 0;
...
+#[test]
+fn test_state_format() {
+    // The purpose of this test is to make sure that if we accidentally
+    // change the binary format of the `EngineState` then we will know
+    // about it. This is important because changing the state format will
+    // break the contract unless we do a state migration.
+    let args = aurora_engine::parameters::NewCallArgs {
+        chain_id: aurora_engine_types::types::u256_to_arr(&666.into()),
+        owner_id: "boss".into(),
+        bridge_prover_id: "prover_mcprovy_face".into(),
```
joshuajbouw, line 27: LOL

joshuajbouw (approved):

sept-en (approved):
Elegant and really useful!

PR #334, opened 2021-11-06
Add CI for checking committer(s) email(s)

strokovok:
Reuse [this](https://github.com/aurora-is-near/.github/blob/master/.github/workflows/check_emails.yml) workflow for checking that committers' emails are suffixed with "@aurora.dev"

### DIFF
--- a/.github/workflows/emails.yml
+++ b/.github/workflows/emails.yml
@@ -2,4 +2,4 @@ name: Emails
 on: ["push", "pull_request"]
 jobs:
   test:
-    uses: aurora-is-near/.github/.github/workflows/check_emails.yml@test-emails-workflow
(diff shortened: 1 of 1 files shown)


birchmd (approved):
Looks good in terms of doing checking what it is supposed. But I don't think this will do anything about the committer emails be from the aurora domain. We all push commits using our aurora emails. The issue is the commits GitHub creates when you merge a PR via the web interface as signed using personal emails. The way to eliminate personal emails is to always merge PRs via the command line, never via the web interface.

joshuajbouw (approved):

joshuajbouw:
Should we revive this @strokovok ?

0x3bfc:
@joshuajbouw @strokovok do we still need to enforce this check ?

mrLSD (approved):

PR #337, opened 2021-11-09
Fix 0x prefix on address, if there

### LINKED ISSUE #335: `0x` isn't being stripped when being passed into the methods
***URGENT***: This needs to be fixed before Nov 12th.

We had an incident where a user had manually entered their address with the `0x` prefix which should be an accepted format as it is fairly standard.

It was debated internally if we should allow this or not and if this was actually an oversight. Given that `0x` is standard across ETH as inputs for addresses, we deemed it a good idea to make the case for accepting addresses with the `0x` prefix even though the current implementation is fairly sound enough. This will reduce confusion.

A word of warning however, if someone does incorrectly enter an address, even with missing characters, there is nothing that we can do to recover the funds since this is completely decentralised. There should be a strict warning about this.

### Fix

* Allow for the `0x` prefix on addresses, as its standard. While also continuing to accept it without.
* Check on length of the tx and to ensure that the format certainly does have the prefix of `0x`. Any other will not be accepted.
* Tests must be created to ensure that it does in fact work.

### Links

[Relevant tx](https://etherscan.io/tx/0xf19872ca79edfcf6cd930984eb2d382eda9132313e44f048af2f67307ed09092): Note the input for the address on NEAR.

joshuajbouw:
There was an incident detailed in #335. This fixes that incident.

In the future, it would be a good idea to make the test framework more general which would allow for different sizes of inputs to the Engine.

This is labelled as critical as we have to rescue two users funds in a timely manner.

This closes #335.

### DIFF
--- a/engine-tests/src/tests/eth_connector.rs
+++ b/engine-tests/src/tests/eth_connector.rs
@@ -591,6 +591,76 @@ fn test_ft_transfer_call_without_message() {
 }
 
 #[test]
+fn test_deposit_with_0x_prefix() {
+    let (master_account, contract) = init(CUSTODIAN_ADDRESS);
+
+    let eth_custodian_address: [u8; 20] = {
+        let mut buf = [0u8; 20];
+        let bytes = hex::decode(CUSTODIAN_ADDRESS).unwrap();
+        buf.copy_from_slice(&bytes);
+        buf
+    };
+    let recipient_address = [10u8; 20];
+    let deposit_amount = U256::from(17);
+    let deposit_event = aurora_engine::deposit_event::DepositedEvent {
+        eth_custodian_address,
+        sender: [0u8; 20],
+        // Note the 0x prefix before the deposit address.
+        recipient: [
+            CONTRACT_ACC,
+            ":",
+            "0x",
+            hex::encode(&recipient_address).as_str(),
+        ]
+        .concat(),
+        amount: deposit_amount,
+        fee: U256::zero(),
+    };
+
+    let event_schema = ethabi::Event {
+        name: aurora_engine::deposit_event::DEPOSITED_EVENT.into(),
+        inputs: aurora_engine::deposit_event::DepositedEvent::event_params(),
+        anonymous: false,
+    };
+    let log_entry = aurora_engine::log_entry::LogEntry {
+        address: eth_custodian_address.into(),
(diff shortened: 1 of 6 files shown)


sept-en (approved):
LGTM!

Some thoughts: as we are already doing these additional checks and improvements, I think maybe it also makes sense to add additional checks in `parse_event_message` for the validity of NEAR account in `TokenMessageData::Eth` branch.
WDYT @joshuajbouw @birchmd ?

birchmd (approved):
Looks like it should fix the immediate issue, thanks @joshuajbouw !

The way data flows through this deposit process via specifically formatted strings makes the logic very hard to follow. I understand why it must happen this way on NEAR itself, but I think some refactoring of this code could be done to make the logic easier to follow (and to make it all type-checked).

We should merge the present PR right away since it is urgent, but it might make sense to do some refactoring as a follow up PR as it could expediate fixes this like this in the future.

birchmd:
> it also makes sense to add additional checks in parse_event_message for the validity of NEAR account in TokenMessageData::Eth branch.

@sept-en this is done by @mrLSD in his `AccountId` PR (see [here](https://github.com/aurora-is-near/aurora-engine/pull/318/files#diff-a105b0599aee5469a32de458841009257720647a34938e3605fe67f4daad548bR140)). I do not think it needs to be done here and we should merge that other PR instead.

sept-en:
@birchmd I agree! Yeah, I didn't check that mentioned PR before, but it's nice that it's there.

mrLSD (approved):

joshuajbouw:
Thanks for the temp lower level tests @birchmd

joshuajbouw:
> Looks like it should fix the immediate issue, thanks @joshuajbouw !
> 
> The way data flows through this deposit process via specifically formatted strings makes the logic very hard to follow. I understand why it must happen this way on NEAR itself, but I think some refactoring of this code could be done to make the logic easier to follow (and to make it all type-checked).
> 
> We should merge the present PR right away since it is urgent, but it might make sense to do some refactoring as a follow up PR as it could expediate fixes this like this in the future.

Thats exactly as I had been feeling when going through it. It was quite hard to follow, especially when I saw "message", and realised it actually did contain the address, etc. This area does need more work, for sure.

PR #338, opened 2021-11-09
Value field for call args

### LINKED ISSUE #309: Engine method `call` has no way to attach ETH
The [method `call`](https://github.com/aurora-is-near/aurora-engine/blob/f1ebb8264383e7047661f830f75849f872df47e6/engine/src/lib.rs#L194) is supposed to be equivalent to the [method `submit`](https://github.com/aurora-is-near/aurora-engine/blob/f1ebb8264383e7047661f830f75849f872df47e6/engine/src/lib.rs#L206), except allows interacting with the EVM from a NEAR account as opposed to an Aurora account. However `call` does not have any way to attach an ETH balance to the transaction which greatly limits its usefulness.

I propose we add a `value` field to [`FunctionCallArgs` struct](https://github.com/aurora-is-near/aurora-engine/blob/f1ebb8264383e7047661f830f75849f872df47e6/engine/src/parameters.rs#L113) to address this issue.

Note: to preserve backwards compatibility we may wish to accept both a version of `FunctionCallArgs` with and without the new field; in the case it is not given it defaults to 0 (the current behaviour). This may or may not be possible since borsh is not a self-describing format. If there are currently no mainnet usages of `call` then maybe the breaking change is fine.

andrcmdr:
**Implementation to address [issue of attaching ETH balance to the transation](https://github.com/aurora-is-near/aurora-engine/issues/309)**

PR fixes #309

- Introduction of `CallArgsType` for generalization of `FunctionCallArgs` to support `value` field in `call()` function and to support backward compatibility with previous data structure.
- Introduction of `Wei` compatible Borsh-encoded type for value field, to support passing `value` field in `call()` function and attach an ETH balance to the transaction, and to provide backward type compatibility.

**Added:**

- `Wei` compatible Borsh-encoded value field for the new `FunctionCallArgs` structure, to attach an ETH balance to the transaction.
- Preserve legacy `FunctionCallArgsLegacy` structure, to provide backward type compatibility.
- Introduce `CallArgsType` enum for choosing type (type casting) between current or legacy Borsh-encoded parameters for passing to the `call` function, and to provide backward type compatibility.

**Introduce trait based generic helpers:**
- Wrapper for handling in a generic way current and legacy Borsh-encoded parameters returned by the `sdk::read_input_borsh()` function call, to provide backward type compatibility.
- Type casting from `CallArgsType` into current and legacy Borsh-encoded parameters for passing to the `call` function, and to provide backward type compatibility.

**Added:**
- `Wei` compatible Borsh-encoded raw value to attach an ETH balance to the transaction.
- Type casting from `Wei` compatible Borsh-encoded raw value into the `Wei` value, to attach an ETH balance to the transaction.
_(This is the only correct way to overcome Rust's orphaning rules, by implementing local type, compatible with `Wei` and `BorshSerialize`/`BorshDeserialize` macro derived methods, 'cause both of them placed in external crates)._

Change correspondingly the `call()` method on the EVM contract and `Engine::call_with_args()` function.

Some fix in test and test utils.

Updating local workspace (fix on exact revision in toml for `near-blake2` and `evm` crates) and toolchain to make compilation and building the whole project in one wasm32 executable possible on current nightly toolchain (which tested and works fine in local environment).
**That possibly couldn't work properly with current CI pipeline settings and used toolchain, involving not passing tests and builds successfully**. For addressing this and for testing workspace and toolchain updates locally I've made a push of these changes (related only to workspace and toolchain update) in a [separate feature branch](https://github.com/aurora-is-near/aurora-engine/tree/toolchain-and-workspace-update), [here is the commit](https://github.com/aurora-is-near/aurora-engine/commit/c2b66076ff9a40333d82328a94087003719bfff8).


rust-toolchain
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2021-08-01"
+channel = "nightly-2021-11-05"
```
birchmd, line 2: Please keep the scope of a PR confined to the feature/fix it is implementing. If we want to change the nightly toolchain and update `Cargo.lock` then it should be done as a separate PR.

engine/src/engine.rs
```diff
@@ -607,11 +607,16 @@ impl Engine {
         Ok(SubmitResult::new(status, used_gas, logs))
     }
 
-    pub fn call_with_args(&mut self, args: FunctionCallArgs) -> EngineResult<SubmitResult> {
-        let origin = self.origin();
-        let contract = Address(args.contract);
-        let value = Wei::zero();
-        self.call(origin, contract, value, args.input, u64::MAX, Vec::new())
+    /// Call the EVM contract with arguments (attached ETH balance, input, gas limit, etc.)
+    pub fn call_with_args(&mut self, args: impl Into<CallArgsType>) -> EngineResult<SubmitResult> {
```
birchmd, line 611: I think we should simply change the signature here to only accept `CallArgsType`. The conversion from the legacy type can be done at the call-site if needed.

engine/src/parameters.rs
```diff
@@ -133,9 +133,44 @@ impl SubmitResult {
...
+pub enum CallArgsType
+where
+    Self: Sized,
+{
+    New(FunctionCallArgs),
+    Legacy(FunctionCallArgsLegacy),
+}
+
+/// Wrapper for handling in a generic way current and legacy Borsh-encoded parameters returned by the sdk::read_input_borsh() function call, to provide backward type compatibility
+pub fn call_args_type_wrap<T: Into<CallArgsType>>(args: T) -> impl Into<CallArgsType> { args }
```
birchmd, line 158: I would prefer we explicitly have some logic which tries to deserialize the new type and falls back on the old type in case of failure, for the sake of clarity.

It's not obvious to me that this will even work as intended; how does it decide at runtime which type it will use to deserialize?

birchmd (changes requested):

engine/src/parameters.rs
andrcmdr: Yeah, sure, will change it!
Will make PR for toolchain update from feature branch made previously here: [toolchain-and-workspace-update](https://github.com/aurora-is-near/aurora-engine/tree/toolchain-and-workspace-update)

rust-toolchain
andrcmdr: Yeah, sure, will change it.

Will make PR for toolchain update from feature branch made previously here: [toolchain-and-workspace-update](https://github.com/aurora-is-near/aurora-engine/tree/toolchain-and-workspace-update)

andrcmdr: **Update:**
Here's the [PR from feature branch which contain just toolchain update](https://github.com/aurora-is-near/aurora-engine/pull/339)

engine/src/parameters.rs
andrcmdr: Yeah, I'm also not sure about how it'll be parsed, i.e. the linear bytes taken from memory through SDK function call - this trick with impl Traits works pretty prefect with high-level types, but not with low level types, like byte sequences, I guess. So, yeah, should do this explicitly in a low level style.

mrLSD: Please notice:
```
pub fn call_args_type_wrap<T: Into<CallArgsType>>(args: T) -> impl Into<CallArgsType> { args } 
```
I understand that you like `impl Trait` notation. But if you use trait bounds, you simply can use it:
```
pub fn call_args_type_wrap<T: Into<CallArgsType>>(args: T) -> T { args } 
```

mrLSD: Also, it's not obvious what that function does. I think more explicit logic will be helpful.

andrcmdr: It was just a workaround for:
`let x: impl Foo = bar();`
'Cause Rust still doesn't support `impl Trait` in variable type declarations.
Thus, the workaround would look like:
`fn wrapper<T: Foo>(x: T) -> impl Foo { x }
let x = wrapper(bar());`
I've start use some impl Trait acrobatics, which suits more for type casting, but in this case I should use just explicit deserialization from byte sequence and try from slice (Michael gave a hint above already)...
So, I already rewrote it. Will make new PR with cherry picking of changes to new feature branch based on updated/rebased yesterday develop branch.

andrcmdr:
Will make new PR with cherry picking of changes to new feature branch based on updated/rebased yesterday develop branch.

PR #341, opened 2021-11-09
Test verifying the EVM log returns the correct address

### LINKED ISSUE #336: Invalid low level EVM function `call` behaviour
The `call` (low level EVM function) should emit the callee contract address in the transaction logs, however it emits the caller contract address. The following example reproduces the issue:

```javascript
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract Caller {
    function greet(address to) public {
        to.call(abi.encodeWithSelector(Greeter(to).greet.selector));
    }
}


contract Greeter { // callee contract
    event Logger(address sender);

    function greet() public {
        emit Logger(msg.sender);
    }
}
```

***On Aurora testnet:*** 

By calling `Caller.greet()`, it returns the caller contract address in the transaction logs.

image

image (1)

***On Goerli testnet:*** 

image
image

`delegateCall` is working fine as expected (emitting the caller contract address in the logs). For more info please refer to [solidity docs](https://docs.soliditylang.org/en/v0.8.9/types.html?highlight=address.#members-of-addresses)

birchmd:
Adds a test showing the correct address is returned by the engine.

This means the issue reported in #336 must not come from the engine, and instead must arise from elsewhere (maybe the relayer?).

Closes #336

### DIFF
--- a/engine-tests/src/test_utils/solidity.rs
+++ b/engine-tests/src/test_utils/solidity.rs
@@ -74,6 +74,17 @@ impl ContractConstructor {
         }
     }
 
+    pub fn deploy_without_constructor(&self, nonce: U256) -> TransactionLegacy {
+        TransactionLegacy {
+            nonce,
+            gas_price: Default::default(),
+            gas_limit: u64::MAX.into(),
+            to: None,
+            value: Default::default(),
+            data: self.code.clone(),
+        }
+    }
+
     pub fn deploy_without_args(&self, nonce: U256) -> TransactionLegacy {
         self.deploy_with_args(nonce, &[])
     }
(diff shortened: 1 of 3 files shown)


0x3bfc (approved):
Thanks @birchmd. This proves that it is not an engine related issue which is great. Will continue troubleshooting other components.

0x3bfc:
@birchmd after troubleshooting, it seems to be a relayer issue. The handling of the `address` value in the event always relies on the `transaction.to` value which is not the right way to handle such important information.

That's actually a relief because it can be resolved by reindexing. 

More details will be followed here https://github.com/aurora-is-near/aurora-relayer/issues/118

PR #342, opened 2021-11-09
Remove Default trait bound from engine IO

birchmd:
The primary goal of this PR is to remove the `Default` trait bound that was required on the `IO` type after #314 . The primary reason for this trait bound was the usage of `Default` in `LookupMap` and `BijectionMap`. As a result, this PR removes `LookupMap` and refactors `BijectionMap` to no longer have a `Default` implementation, as well as to be more type-safe.

This is change does not impact the runtime behaviour of the engine in any way. It is continuing the path towards being able to run the engine without the NEAR runtime (ie as a standalone binary).

### DIFF
--- a/engine/src/connector.rs
+++ b/engine/src/connector.rs
@@ -1,7 +1,7 @@
 use crate::admin_controlled::{AdminControlled, PausedMask};
 use crate::deposit_event::DepositedEvent;
 use crate::engine::Engine;
-use crate::fungible_token::{FungibleToken, FungibleTokenMetadata};
+use crate::fungible_token::{FungibleToken, FungibleTokenMetadata, FungibleTokenOps};
 use crate::json::parse_json;
 use crate::parameters::{
     BalanceOfCallArgs, BalanceOfEthCallArgs, FinishDepositCallArgs, InitCallArgs,
@@ -29,10 +29,9 @@ pub const UNPAUSE_ALL: PausedMask = 0;
 pub const PAUSE_DEPOSIT: PausedMask = 1 << 0;
 pub const PAUSE_WITHDRAW: PausedMask = 1 << 1;
 
-#[derive(BorshSerialize, BorshDeserialize)]
-pub struct EthConnectorContract<I: IO + Default> {
+pub struct EthConnectorContract<I: IO> {
     contract: EthConnector,
-    ft: FungibleToken<I>,
+    ft: FungibleTokenOps<I>,
     paused_mask: PausedMask,
     io: I,
 }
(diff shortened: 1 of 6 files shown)


engine/src/connector.rs
```diff
@@ -62,11 +61,15 @@ pub struct OnTransferMessageData {
     pub fee: U256,
 }
 
-impl<I: IO + Default + Copy> EthConnectorContract<I> {
+impl<I: IO + Copy> EthConnectorContract<I> {
```
birchmd, line 64: `Copy` (or at the very least `Clone`) is still required, unfortunately. The reason is because the engine and connector sub-components benefit from having a handle to the `IO` they can access in `self` methods (otherwise we end up passing in `io` all over the place and it's pretty messy). I think it might be ok if this ends up remaining the case in the long term, but perhaps in a future PR I'll see if I can sort out all the lifetimes to be able to pass a reference down instead of the actual object.

engine/src/engine.rs
```diff
@@ -201,6 +200,30 @@ impl From<BalanceOverflow> for GasPaymentError {
...
+impl AsRef<[u8]> for ERC20Address {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+}
+impl From<Vec<u8>> for ERC20Address {
+    fn from(bytes: Vec<u8>) -> Self {
+        Self(Address::from_slice(&bytes))
+    }
+}
```
joshuajbouw, line 213: Shouldn't this be a TryFrom since the bytes could be more or less? We should try to catch these.

engine/src/engine.rs
```diff
@@ -201,6 +200,30 @@ impl From<BalanceOverflow> for GasPaymentError {
...
+impl AsRef<[u8]> for NEP141Account {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+}
+impl From<Vec<u8>> for NEP141Account {
+    fn from(bytes: Vec<u8>) -> Self {
+        Self(AccountId::try_from(bytes).unwrap())
+    }
+}
```
joshuajbouw, line 225: I do think we should keep these as a try_from and then catch them in the logic. We could in the future use this logic elsewhere where it may fail.

joshuajbouw (changes requested):

engine/src/engine.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/342/commits/339dc43def5853776090c5ff17d65f8d788bbd3a

engine/src/engine.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/342/commits/339dc43def5853776090c5ff17d65f8d788bbd3a

PR #344, opened 2021-11-10
Chore: Newtypes for gas

joshuajbouw:
While working on the `engine-standalone`, I have come to realise that the gas usages, since they are costs, should be using newtypes and not a normal type. This prevents dev error in the case of trying to use the wrong values, in the wrong areas which can lead to unintentional bugs.

Newtypes are 0 cost, therefore this is a wise decision to add this bit of work in. There will be more work around these areas if the need arises as I continue work through out the library and notice improvements that should be done. This is one of them that I had noticed.

This is a semantic change and intended for dev experience and will help the code be at a higher level of quality. From this point forward, the types will be made aware to the developer when working on the library and forces them to be consciously aware of which types they are dealing with, and not just general values.

Notably, we now have:

```rust
pub struct EthGas(u64);

pub struct NearGas(u64);
```

... as well as boiler plate implementations which make working with the newtypes generally easier.

### DIFF
--- a/engine-precompiles/src/lib.rs
+++ b/engine-precompiles/src/lib.rs
@@ -21,25 +21,27 @@ use crate::hash::{RIPEMD160, SHA256};
 use crate::identity::Identity;
 use crate::modexp::ModExp;
 use crate::native::{ExitToEthereum, ExitToNear};
+use crate::prelude::types::EthGas;
+use crate::prelude::Vec;
 use crate::secp256k1::ECRecover;
-use aurora_engine_types::{account_id::AccountId, Address, BTreeMap, Box};
+use aurora_engine_types::{account_id::AccountId, vec, Address, BTreeMap, Box};
 use evm::backend::Log;
 use evm::executor;
 use evm::{Context, ExitError, ExitSucceed};
 
 #[derive(Debug, Default)]
 pub struct PrecompileOutput {
-    pub cost: u64,
-    pub output: prelude::Vec<u8>,
-    pub logs: prelude::Vec<Log>,
+    pub cost: EthGas,
+    pub output: Vec<u8>,
+    pub logs: Vec<Log>,
 }
 
 impl PrecompileOutput {
-    pub fn without_logs(cost: u64, output: prelude::Vec<u8>) -> Self {
+    pub fn without_logs(cost: EthGas, output: Vec<u8>) -> Self {
         Self {
             cost,
             output,
-            logs: prelude::Vec::new(),
+            logs: Vec::new(),
         }
     }
 }
@@ -48,7 +50,8 @@ impl From<PrecompileOutput> for evm::executor::PrecompileOutput {
     fn from(output: PrecompileOutput) -> Self {
         evm::executor::PrecompileOutput {
             exit_status: ExitSucceed::Returned,
-            cost: output.cost,
+            // This shouldn't fail, and is expected.
+            cost: output.cost.into_u64(),
             output: output.output,
             logs: output.logs,
         }
@@ -60,7 +63,7 @@ type EvmPrecompileResult = Result<evm::executor::PrecompileOutput, ExitError>;
 /// A precompiled function for use in the EVM.
 pub trait Precompile {
     /// The required gas in order to run the precompile function.
-    fn required_gas(input: &[u8]) -> Result<u64, ExitError>
+    fn required_gas(input: &[u8]) -> Result<EthGas, ExitError>
     where
         Self: Sized;
 
@@ -68,7 +71,7 @@ pub trait Precompile {
     fn run(
         &self,
         input: &[u8],
-        target_gas: Option<u64>,
+        target_gas: Option<EthGas>,
         context: &Context,
         is_static: bool,
     ) -> EvmPrecompileResult;
@@ -109,7 +112,7 @@ impl executor::PrecompileSet for Precompiles {
         is_static: bool,
     ) -> Option<Result<executor::PrecompileOutput, executor::PrecompileFailure>> {
         self.0.get(&address).map(|p| {
-            p.run(input, gas_limit, context, is_static)
+            p.run(input, gas_limit.map(EthGas::new), context, is_static)
                 .map_err(|exit_status| executor::PrecompileFailure::Error { exit_status })
         })
     }
@@ -122,14 +125,14 @@ impl executor::PrecompileSet for Precompiles {
 impl Precompiles {
     #[allow(dead_code)]
     pub fn new_homestead(current_account_id: AccountId) -> Self {
-        let addresses = prelude::vec![
+        let addresses = vec![
             ECRecover::ADDRESS,
             SHA256::ADDRESS,
             RIPEMD160::ADDRESS,
             ExitToNear::ADDRESS,
             ExitToEthereum::ADDRESS,
         ];
-        let fun: prelude::Vec<Box<dyn Precompile>> = prelude::vec![
+        let fun: prelude::Vec<Box<dyn Precompile>> = vec![
             Box::new(ECRecover),
             Box::new(SHA256),
             Box::new(RIPEMD160),
@@ -143,7 +146,7 @@ impl Precompiles {
 
     #[allow(dead_code)]
     pub fn new_byzantium(current_account_id: AccountId) -> Self {
-        let addresses = prelude::vec![
+        let addresses = vec![
             ECRecover::ADDRESS,
             SHA256::ADDRESS,
             RIPEMD160::ADDRESS,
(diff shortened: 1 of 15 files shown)


engine-types/src/types.rs
```diff
@@ -1,14 +1,129 @@
...
+/// Near gas type which wraps an underlying u64.
+pub struct NearGas(u64);
+
+impl Display for NearGas {
+    fn fmt(&self, f: &mut Formatter<'_>) -> crate::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl Deref for NearGas {
```
birchmd, line 24: I thought we agreed we do not want to implement `Deref` for newtypes.

engine-types/src/types.rs
```diff
@@ -1,14 +1,129 @@
...
+    /// Constructs a new `NearGas` with a given u64 value.
+    pub const fn new(gas: u64) -> NearGas {
+        Self(gas)
+    }
+
+    pub fn checked_add(self, rhs: Self) -> Option<Self> {
+        self.0.checked_add(rhs.0).map(Self)
+    }
+
+    pub fn checked_mul(self, rhs: Self) -> Option<Self> {
```
birchmd, line 66: I do not think it semantically makes sense to be able to multiply gas values together. The whole point of creating a new type is to add semantics to `u64`; its not a pure number any more. When you multiple two things with units together you do not get the same kind of thing back again (for example multiplying two lengths gives an area, not a length).

I think specifically **not** including a multiplication impl is a good thing because there should never be any logic which requires that operation, so if you think you need it then either the logic is wrong or one quantity has been given the incorrect type.

engine-types/src/types.rs
```diff
@@ -1,14 +1,129 @@
...
+/// Ethereum gas type which wraps an underlying u64.
+pub struct EthGas(u64);
+
+impl Display for EthGas {
+    fn fmt(&self, f: &mut Formatter<'_>) -> crate::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl Deref for EthGas {
```
birchmd, line 80: See same comment as `NearGas`

engine-types/src/types.rs
```diff
@@ -1,14 +1,129 @@
...
+    /// Constructs a new `EthGas` with a given u64 value.
+    pub const fn new(gas: u64) -> EthGas {
+        Self(gas)
+    }
+
+    pub fn checked_add(self, rhs: Self) -> Option<Self> {
+        self.0.checked_add(rhs.0).map(Self)
+    }
+
+    pub fn checked_mul(self, rhs: Self) -> Option<Self> {
```
birchmd, line 122: See same comment as `NearGas`.

engine-types/src/types.rs
```diff
@@ -1,14 +1,129 @@
...
+    }
+}
+
+impl AsMut<u64> for NearGas {
+    fn as_mut(&mut self) -> &mut u64 {
+        &mut self.0
+    }
+}
+
+impl From<u64> for NearGas {
```
birchmd, line 50: I don't think we should implement `From` either. It allows developers to fix some compilation errors by simply writing `.into()` without necessarily thinking about it very much. They should ask "why do I need a `NearGas` here and don't have one?" Sometimes it might be the case that they legitimately need to convert a `u64` into a `NearGas`, but then they should explicitly call `new()`.

Newtypes are about providing type safety and every "convenience" method that allows converting to/from the raw type undermines the safety it is providing.

engine-types/src/types.rs
```diff
@@ -82,28 +197,50 @@ impl Wei {
...
+    }
+}
+
+impl Display for Wei {
+    fn fmt(&self, f: &mut Formatter<'_>) -> crate::fmt::Result {
+        self.0.fmt(f)
     }
+}
+
+impl Deref for Wei {
```
birchmd, line 215: See same comment as `NearGas`.

engine-types/src/types.rs
```diff
@@ -82,28 +197,50 @@ impl Wei {
...
-    type Output = Self;
+impl AsMut<U256> for Wei {
+    fn as_mut(&mut self) -> &mut U256 {
+        &mut self.0
+    }
+}
 
-    fn add(self, other: Self) -> Self::Output {
-        Self(self.0 + other.0)
+impl From<U256> for Wei {
```
birchmd, line 241: See same comment as `NearGas`.

engine-types/src/types.rs
joshuajbouw: Thats precisely why I only added it for the underlaying type. I didn't want it to work for any other type.

engine-types/src/types.rs
joshuajbouw: The automagic `Deref` implementation from the `derive_more` crate added some auto magic that wasn't intended. Manually, it doesn't have that negative issues previously experienced.

birchmd: I still don't think it's needed. Just gives another way to subvert the type system. If a function requires the raw type, but we have the gas type then we should ask ourselves why that is the case. If there is an easy escape hatch then we may be lazy and not ask this question when we should.

Of course it should possible to obtain the raw type because sometimes that will be needed (namely at the boundary of our application). But in those cases we should be clear about this (for example `fn raw(self) -> u64`).

engine-types/src/types.rs
birchmd: I think even having it for the underlying type is too much. We should always be forced to ask the question about why the compiler is given an error; if there is any easy escape hatch (like `.into()`) then we might be lazy and not ask it.

Again, it should always be possible to lift a raw value into the gas value because we will need to do it sometimes (namely at the boundary of our application), but this should always be made explicit in the syntax (eg `NearGas::new(number)`) because it shouldn't happen very often.

engine-types/src/types.rs
joshuajbouw: Fair point, double checked other newtype patterns, and correct, we shouldn't be adding `Deref` in this case, especially with what you had stated. Gas throughout precompiles does need to have some math done on it with other types.

engine-types/Cargo.toml
```diff
@@ -18,6 +18,8 @@ ethabi = { git = "https://github.com/darwinia-network/ethabi", branch = "xavier-
 hex = { version = "0.4", default-features = false, features = ["alloc"] }
 primitive-types = { version = "0.10.0", default-features = false, features = ["rlp"] }
 sha3 = { version = "0.9.1", default-features = false }
+evm = { git = "https://github.com/aurora-is-near/sputnikvm.git", default-features = false }
+evm-core = { git = "https://github.com/aurora-is-near/sputnikvm.git", default-features = false }
```
birchmd, line 22: See comment about `pub use evm_core::Opcode;`

engine-precompiles/src/lib.rs
```diff
@@ -47,7 +49,8 @@ impl From<PrecompileOutput> for evm::executor::PrecompileOutput {
     fn from(output: PrecompileOutput) -> Self {
         evm::executor::PrecompileOutput {
             exit_status: ExitSucceed::Returned,
-            cost: output.cost,
+            // This shouldn't fail, and is expected.
```
birchmd, line 52: What's this comment about? `into_u64` is infallible right?

birchmd (approved):
Looks good! Just a couple minor comments.

engine-types/src/types.rs
```diff
@@ -1,14 +1,108 @@
...
+}
+
+impl NearGas {
+    /// Constructs a new `NearGas` with a given u64 value.
+    pub const fn new(gas: u64) -> NearGas {
+        Self(gas)
+    }
+
+    /// Consumes `NearGas` and returns the underlying type.
+    pub fn into_u64(self) -> u64 {
```
mrLSD, line 31: Explicit type notation looks good. But why not implement just `Into` trait (seems it will be a more idiomatic way)?

joshuajbouw: You mean the `From` trait which will automatically implement `Into` as implementing `Into` directly isn't recommended. Originally was say, as per discussions above but Michael believes and I do agree with our application, we should be explicit and use explicit conversions where possible. That way you can't just do `.into()` lazily.

engine-precompiles/src/lib.rs
joshuajbouw: Yeah, that comment should go. Just an artifact from when I was using a different structure.

engine-types/src/types.rs
```diff
@@ -1,14 +1,108 @@
-use super::{str, vec, Add, Address, String, Sub, Vec, U256};
+use crate::{str, vec, Add, Address, Display, Div, Mul, String, Sub, Vec, U256};
 use borsh::{BorshDeserialize, BorshSerialize};
 
+use crate::fmt::Formatter;
+pub use evm_core::Opcode;
```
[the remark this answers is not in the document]
joshuajbouw: Good catch.

PR #345, opened 2021-11-10
Feat(standalone): Standalone

joshuajbouw:
The purpose of this change is to add geth-like [`debug_traceTransaction`] logic to [Sputnik EVM] (rust-blockchain/evm). On top of that the ability to run a standalone engine that is able to keep in sync with the network EVM.

Why this is an important change is that it is absolutely necessary for this logic for both Block Scout and Etherscan's upcoming block explorer for Aurora.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1186,6 +1186,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "engine-standalone"
+version = "0.1.0"
+dependencies = [
+ "aurora-engine",
+ "aurora-engine-sdk",
+ "aurora-engine-types",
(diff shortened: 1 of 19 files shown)


birchmd (approved):
@joshuajbouw I rebased this PR to be on top of the rest of my sdk refactor work. And I pushed some commits to make the new standalone crate compile, plus added a test showing it can run the EVM successfully.

I think this PR is ready to merge after #344 #355 #356 are all merged.

Obviously the standalone does not do anything yet, but I think it would be good to merge this PR laying out the structure of the crate and then we can iterate on it with smaller PRs going forward.

PR #346, opened 2021-11-11
Remove sdk::current_account_id usage from engine-precompiles

birchmd:
This refactors the precompiles to take a `self` parameter as part of `run`. This means the objects implementing the precompiles can hold some state. In particular, the exit precompiles can hold the `current_account_id` instead of needing to look it up via `sdk::current_account_id()`. This is useful because it means the precompiles will not need to receive any `Env` object once we factor out blockchain context related lookups into a trait (as part of the standalone engine project).


joshuajbouw (approved):
At first I was confused, but the reasoning behind it giving it optional state makes sense to me.

Issue #307, opened 2021-10-14
Precompiles blake2/EcRecovery/modexp/bn


rakita:
In general I wanted to take precompiles, don't bother with reimplementing them and just integrate them in REVM, but it seems that this lead me into a big rabbit hole.

Here is the list of changes or just replacements that I did so that eth/tests would pass:
* blake2: one of the tests (or maybe more, I stopped on first) just crashed with `unwrap` I dont have name of exact test that crashed. reimplementing algo was very straightforward 
* EcRecovery: I replaced lib with parity alternative. The problem is probably related to this: https://github.com/openethereum/openethereum/blob/d8305c52ea805e62d7532c3ac76386873984d326/crates/vm/builtin/src/lib.rs#L834-L838
* modexp still continue failing, probably do the padding. I modified this code https://github.com/paritytech/frontier/blob/master/frame/evm/precompile/modexp/src/lib.rs so that it is more robust. I have only Berlin version, didnt test byzantine. The code is made working, but I would need to invest more time to check it and be sure that it is okay.
* bn lib: failed on this tests related to mul: https://github.com/bluealloy/revm/blob/aa12599d07a04d1e5ef6bfc35dc0d795a4b00d8d/bin/revm-ethereum-tests/src/runner.rs#L79-L104 I replaced it with parity alternative.

All new libs are apache2/MIT or similar, all changes can be found here: https://github.com/bluealloy/revm/tree/main/src/precompiles
If needed, eth test runner can be found here: https://github.com/bluealloy/revm/tree/main/bin/revm-ethereum-tests
I am running only Berlin tests and I check state root for verification.

rakita:
@birchmd, It is what it is. Hope it helps.

birchmd:
Cc @frankbraun RE security email and procedures for accepting external vulnerability reports

joshuajbouw:
I have @mrLSD assigned to this for now as he is working on implementing the precompile tests from ETH now.

rakita:
@joshuajbouw @mrLSD maybe this can help: https://crates.io/crates/revm_precompiles

mrLSD:
@rakita could you explain please how the issue related to Aurora?
The description and tests runner about REVM.
REVM - really awesome library! But I need clarification on what you propose in the Aurora context.

I investigated REVM, and am I'm not clear that you propose running for Auorara (SputnikVM) same tests logic for Ethereum like REVM?

I see, that REVM tests include only `GeneralStateTests`:
[...]

PR #348, opened 2021-11-11
 Precompiles tests

mrLSD:
Tests for Ethereum state precompile.
Related to #307

Tests base: https://github.com/ethereum/tests/

## NOTICE
There's an issue to import crates from https://github.com/rust-blockchain/evm-tests

As a temporal solution, it was copied `evm-tests` implementations to `engine-tests/` for:
* `ethjson`
* `keccak_hasher`
* `state`
* `triehash_ethereum`

**It should be fixed in future PR**.


joshuajbouw:
We never intend to copy and paste code from other projects. Need to aim to higher aspirations. We can use ETH's json tests, but we can't just lift the code from Open Ethereum and rust-blockchain/evm.

PR #349, opened 2021-11-12
Factor out blockchain environment variable access as a trait

birchmd:
This is a continuation of the effort to reactor the engine to not depend on NEAR host functions (which is necessary for the standalone engine). It depends on the previous PRs #342 #346 #347 being merged first.

### DIFF
--- a/engine-sdk/src/env.rs
+++ b/engine-sdk/src/env.rs
@@ -1,0 +1,58 @@
+use crate::error::{OneYoctoAttachError, PrivateCallError};
+use aurora_engine_types::account_id::AccountId;
+
+/// Timestamp represented by the number of nanoseconds since the Unix Epoch.
+pub struct Timestamp(u64);
+
+impl Timestamp {
+    pub fn new(ns: u64) -> Self {
+        Self(ns)
+    }
+
+    pub fn nanos(&self) -> u64 {
+        self.0
+    }
+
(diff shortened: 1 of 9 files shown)


engine-sdk/src/env.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    pub fn millis(&self) -> u64 {
+        self.0 / 1_000_000
+    }
+
+    pub fn secs(&self) -> u64 {
+        self.0 / 1_000_000_000
+    }
+}
```
joshuajbouw, line 23: Out of curiosity, why not `Duration`?

joshuajbouw (approved):
I believe that the `IO` functions could actually be made as methods as part of a struct but given time is of the essence and we need to unblock other work, I'm happy with this.

engine-sdk/src/env.rs
birchmd: I looked into using `Duration`, but it did not feel appropriate. `Duration` has no fixed starting point whereas this does (namely the unix epoch). I also looked at using `SystemTime`, but that didn't feel right either because the internals are opaque and system-dependent whereas here we know the underlying raw data is always a `u64` number representing nanoseconds since the unix epoch.

joshuajbouw: Thanks for the reasoning, makes sense to me.

PR #350, opened 2021-11-14
Chore: Rust 2021

joshuajbouw:
Updates to the Rust 2021 edition.

This was done using the `cargo fix --edition` which does the conversion necessary. 

Looks like there really hadn't been any changes that ultimately affect us. However, our downline dependencies do run into problems which need to be tracked.


birchmd:
Closing as this was done in https://github.com/aurora-is-near/aurora-engine/pull/428

PR #351, opened 2021-11-15
Borsh deserialized value field for call args

### LINKED ISSUE #309: Engine method `call` has no way to attach ETH

andrcmdr:
**Implementation to address [issue of attaching ETH balance to the transation](https://github.com/aurora-is-near/aurora-engine/issues/309)**

PR fixes #309

**Implementation for deserialized values from bytes to current or legacy Borsh-encoded parameters for passing to the engine `call` function, and to provide backward type compatibility:**
- Handling in a generic way current and legacy Borsh-encoded parameters returned by the `io.read_input_borsh()` function call, with fallback to previous `FunctionCallArgsLegacy` data structure, to provide backward type compatibility.
- Type matching from `CallArgsType` into current and legacy Borsh-encoded parameters, for passing to the `call` function, and to provide backward type compatibility.

- Preserve legacy `FunctionCallArgsLegacy` structure, to provide backward type compatibility.
- Introducing of `CallArgsType` enum for matching type between current or legacy Borsh-encoded parameters, and to provide backward type compatibility with previous data structure.
- Introducing of `Wei` compatible portable implementation of Borsh-encoded type for value field for the new `FunctionCallArgs` structure, to attach an ETH balance to the transaction and to support passing `value` field to the engine `call` function.
- Type casting from `Wei` compatible Borsh-encoded raw value into the `Wei` value, to attach an ETH balance to the transaction.
_(This is the only correct way to overcome Rust's orphaning rules, by implementing local type, compatible with `Wei` and `BorshSerialize`/`BorshDeserialize` macro derived methods, 'cause both of them placed in external crates)._
- Change correspondingly the engine `call()` method on the EVM contract and `Engine::call_with_args()` function.
- Fix tests and test utils.

### DIFF
--- a/engine-types/src/types.rs
+++ b/engine-types/src/types.rs
@@ -112,7 +112,7 @@ impl Add for Wei {
 /// Type casting from Wei compatible Borsh-encoded raw value into the Wei value, to attach an ETH balance to the transaction
 impl From<WeiU256> for Wei {
     fn from(value: WeiU256) -> Self {
-        Wei(value.into())
+        Wei(U256::from_big_endian(&value))
     }
 }
 
--- a/engine-types/src/types.rs
+++ b/engine-types/src/types.rs
@@ -8,6 +8,8 @@ pub type RawH256 = [u8; 32]; // Unformatted binary data of fixed length.
 pub type EthAddress = [u8; 20];
 pub type Gas = u64;
 pub type StorageUsage = u64;
+/// Wei compatible Borsh-encoded raw value to attach an ETH balance to the transaction
+pub type WeiU256 = [u8; 32];
 
 /// Selector to call mint function in ERC 20 contract
 ///
@@ -104,6 +106,13 @@ impl Add for Wei {
 
     fn add(self, other: Self) -> Self::Output {
         Self(self.0 + other.0)
+    }
+}
+
+/// Type casting from Wei compatible Borsh-encoded raw value into the Wei value, to attach an ETH balance to the transaction
+impl From<WeiU256> for Wei {
+    fn from(value: WeiU256) -> Self {
+        Wei(value.into())
     }
 }
 
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -72,11 +72,11 @@ mod contract {
     #[cfg(feature = "evm_bully")]
     use crate::parameters::{BeginBlockArgs, BeginChainArgs};
     use crate::parameters::{
-        CallArgs, DeployErc20TokenArgs, FunctionCallArgs, FunctionCallArgsLegacy,
-        GetErc20FromNep141CallArgs, GetStorageAtArgs, InitCallArgs, IsUsedProofCallArgs,
-        NEP141FtOnTransferArgs, NewCallArgs, PauseEthConnectorCallArgs, ResolveTransferCallArgs,
-        SetContractDataCallArgs, StorageDepositCallArgs, StorageWithdrawCallArgs, SubmitResult,
-        TransactionStatus, TransferCallCallArgs, ViewCallArgs,
+        CallArgs, DeployErc20TokenArgs, GetErc20FromNep141CallArgs, GetStorageAtArgs, InitCallArgs,
+        IsUsedProofCallArgs, NEP141FtOnTransferArgs, NewCallArgs, PauseEthConnectorCallArgs,
+        ResolveTransferCallArgs, SetContractDataCallArgs, StorageDepositCallArgs,
+        StorageWithdrawCallArgs, SubmitResult, TransactionStatus, TransferCallCallArgs,
+        ViewCallArgs,
     };
     use aurora_engine_sdk::env::Env;
     use aurora_engine_sdk::io::{StorageIntermediate, IO};
@@ -220,31 +220,19 @@ mod contract {
     #[no_mangle]
     pub extern "C" fn call() {
         let io = Runtime;
-        let bytes = io.read_input();
-        let args: Option<CallArgs> = {
-            if let Ok(value) = bytes.to_value::<FunctionCallArgs>() {
-                Some(CallArgs::New(value))
-            } else if let Ok(value) = bytes.to_value::<FunctionCallArgsLegacy>() {
-                Some(CallArgs::Legacy(value))
-            } else {
-                None
-            }
-        };
+        let bytes = io.read_input().to_vec();
+        let args = CallArgs::deserialize(&bytes).sdk_expect("ERR_BORSH_DESERIALIZE");
         let current_account_id = io.current_account_id();
-        if let Some(args) = args {
-            let mut engine = Engine::new(
-                predecessor_address(&io.predecessor_account_id()),
-                current_account_id,
-                io,
-                &io,
-            )
-            .sdk_unwrap();
-            Engine::call_with_args(&mut engine, args)
-                .map(|res| res.try_to_vec().sdk_expect("ERR_SERIALIZE"))
-                .sdk_process();
-        } else {
-            sdk::panic_utf8("ERR_BORSH_DESERIALIZE".as_bytes());
-        }
+        let mut engine = Engine::new(
+            predecessor_address(&io.predecessor_account_id()),
+            current_account_id,
+            io,
+            &io,
+        )
+        .sdk_unwrap();
+        Engine::call_with_args(&mut engine, args)
+            .map(|res| res.try_to_vec().sdk_expect("ERR_SERIALIZE"))
+            .sdk_process();
         // TODO: charge for storage
     }
 
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -72,7 +72,7 @@ mod contract {
     #[cfg(feature = "evm_bully")]
     use crate::parameters::{BeginBlockArgs, BeginChainArgs};
     use crate::parameters::{
-        CallArgsType, DeployErc20TokenArgs, FunctionCallArgs, FunctionCallArgsLegacy,
+        CallArgs, DeployErc20TokenArgs, FunctionCallArgs, FunctionCallArgsLegacy,
         GetErc20FromNep141CallArgs, GetStorageAtArgs, InitCallArgs, IsUsedProofCallArgs,
         NEP141FtOnTransferArgs, NewCallArgs, PauseEthConnectorCallArgs, SetContractDataCallArgs,
         SubmitResult, TransactionStatus, TransferCallCallArgs, ViewCallArgs,
@@ -209,11 +209,12 @@ mod contract {
     #[no_mangle]
     pub extern "C" fn call() {
         let io = Runtime;
-        let args: Option<CallArgsType> = {
-            if let Ok(value) = io.read_input_borsh::<FunctionCallArgs>() {
-                Some(CallArgsType::New(value))
-            } else if let Ok(value) = io.read_input_borsh::<FunctionCallArgsLegacy>() {
-                Some(CallArgsType::Legacy(value))
+        let bytes = io.read_input();
+        let args: Option<CallArgs> = {
+            if let Ok(value) = bytes.to_value::<FunctionCallArgs>() {
+                Some(CallArgs::New(value))
+            } else if let Ok(value) = bytes.to_value::<FunctionCallArgsLegacy>() {
+                Some(CallArgs::Legacy(value))
             } else {
                 None
             }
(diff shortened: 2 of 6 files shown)


andrcmdr:
I was radically change and simplify logic (following KISS & DRY), using more straightforward and explicit way with relying on generic borsh deserialization (`io.read_input_borsh()` and current `to_value()` method). (Thanks to @birchmd for working on loose coupling decomposition of SDK methods and host functions to `IO` trait for standalone engine).

Seems like now it works more reliable, rather than when using `impl Trait` based type matching implementation, which was less obvious, more implicit, less readable.

andrcmdr:
Some lint checks wasn't successful, because of changing in imports for new types, but this can be solved with `cargo fmt` in some next PR merge.

engine/src/lib.rs
```diff
@@ -209,11 +209,23 @@ mod contract {
...
-            .sdk_process();
+        let args: Option<CallArgsType> = {
+            if let Ok(value) = io.read_input_borsh::<FunctionCallArgs>() {
+                Some(CallArgsType::New(value))
+            } else if let Ok(value) = io.read_input_borsh::<FunctionCallArgsLegacy>() {
+                Some(CallArgsType::Legacy(value))
+            } else {
+                None
+            }
+        };
```
birchmd, line 220: A few comments here:

- I think the initial type to attempt to parse from the input should be `CallArgsType`. The reason is because an enum has built-in backwards compatibility for future changes we might make (the byte indicating the enum variant effectively functions as a version byte).
- It is a little more efficient to read the input into a vector (`io.read_input().to_vec()`) instead of invoking `io.read_input_borsh()` multiple times because `read_input_borsh` will read/write from the host register every time and allocate a new vector every time.
- I think this fallback logic should be attached to the `CallArgsType` itself as a function like `pub fn deserialize(input: &[u8]) -> Self`. This way we can re-use the logic in another place easily (if need).
- Nit: I don't really like the `CallArgsType` name. Having the `Type` suffix seems redundant (obviously its a type). Personally I would have it take on the `FunctionCallArgs` name, and the new struct could be called `FunctionCallArgsV1`, or something like that.

engine-types/src/types.rs
```diff
@@ -107,6 +109,13 @@ impl Add for Wei {
     }
 }
 
+/// Type casting from Wei compatible Borsh-encoded raw value into the Wei value, to attach an ETH balance to the transaction
+impl From<WeiU256> for Wei {
+    fn from(value: WeiU256) -> Self {
+        Wei(value.into())
```
birchmd, line 115: I prefer calling `U256::from_big_endian` directly, just to be clear about what the endianness of the encoding.

birchmd (changes requested):
Nice work, this is definitely a step in the right direction! A few more things to fix up and then this should be good to go.

andrcmdr:
Update feature branch by commit https://github.com/aurora-is-near/aurora-engine/commit/977b96b8027014c84890f5e765e54295252d5dea

**Implementation for deserialized values from bytes to current or legacy Borsh-encoded parameters for passing to the engine call function, and to provide backward type compatibility:**
- Improved code formatting by `cargo fmt`
- Addressing notes about bytes order (endianness) in array type conversion to `U256` and `Wei` types
- Caching bytes returned by `io.read_input()` to prevent double calls for reading register from engine's `call()` function during serialization attempts

engine-types/src/types.rs
andrcmdr: Addressed notes and remarks here, made a commit with corrections and improvements:
https://github.com/aurora-is-near/aurora-engine/pull/351#issuecomment-969075880

engine/src/lib.rs
andrcmdr: Addressed notes and remarks here, made a commit with corrections and improvements:
https://github.com/aurora-is-near/aurora-engine/pull/351#issuecomment-969075880

engine/src/parameters.rs
```diff
@@ -130,13 +130,33 @@ impl SubmitResult {
...
+    pub contract: RawAddress,
+    pub input: Vec<u8>,
+}
+
+/// Deserialized values from bytes to current or legacy Borsh-encoded parameters
+/// for passing to the engine `call` function, and to provide backward type compatibility
+#[derive(BorshSerialize, BorshDeserialize)]
+pub enum CallArgsType
+where
+    Self: Sized,
```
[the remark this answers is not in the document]
andrcmdr: Addressed notes and remarks here, made a commit with corrections and improvements:
https://github.com/aurora-is-near/aurora-engine/pull/351#issuecomment-969075880

engine/src/lib.rs
birchmd: Thanks for the changes. Please also make changes for the first bullet (deserialize enum directly), and third bullet (encapsulate fallback logic in the `impl CallArgs`).

andrcmdr: Here I cannot agree, 'cause took ownership on this task.
Would explain my opinion:
I was wrong about enum - it's flexible for usage with `Engine::call_with_args()` for matching of already successfully parsed type, but in case with borsh we would have a type incompatibility issue, I mean we cannot do as in example you've provide [here](https://github.com/birchmd/aurora-engine/commit/92313f73dcaa0c1f5a53d8d16ebbacf89ad66849), the better way is to returned parsed bytes would be mapped on a struct, and enum would be used in type matches in further, thus I just decide and prefer to make things more simple, clean, straightforward and obvious, and readable (what is more important!), `if let` has more good fit for this usage scenario, rather than `match`, we just trying to parse to newer type, then, if borsh returned error, we fallback to try to parse and map bytes to older type, and even if in this case we would have an error (if something unusual would be sent through register), then we fallback to exception panic with message.
CallArgs enum would be useful in `Engine::call_with_args()` to choose between types, when we send parsed bytes mapped already to new OR legacy data structure already contained in the enum as collection of CallArg types.
Let's just make things more simple and clean, straightforward and obvious, and readable as well, and then we would figure out how we would reuse that pattern for future operations with borsh, sdk and registers, if we would have such case in the future, and if we would have ones, we can make better decision on refactor and generalization of that pattern for reuse.

birchmd: Sure, you can keep using `if let` if you want instead of `match`, I don't care about that much either way.

What I do care about is 

1. The first type we attempt to deserialize from the bytes is `CallArgs`. This is better than trying to deserialize a particular variant because it means no logic changes need to happen in the case we add new variants to `CallArgs` in the future. And this is something we might do, for example if we want to allow access lists, or some other new Ethereum transaction type that does not yet exist.
2. The logic for deserializing the bytes is in an impl for `CallArgs`. By encapsulating in this way it means the logic can easily be re-used in other parts of the code, and it makes the `call` function in `lib.rs` cleaner because the deserialization logic is hidden behind a function call.

andrcmdr: Ok, this make sense, I agree, I will reimplement this as you proposed, to address your remarks, and to feel how borsh would map bytes to structs encapsulated in enums.

andrcmdr:
Update feature branch by commit https://github.com/aurora-is-near/aurora-engine/pull/351/commits/c7a5b544eb6c8f9290a2f8201925abb1b13c4802

**Implementation for deserialized values from bytes to current or legacy Borsh-encoded parameters for passing to the engine call function, and to provide backward type compatibility:**
- Implemented `deserialize()` method for `CallArgs` to follow SOI (single responsibility, open-closed, and interface segregation) principles and made code even more clean on engine top level.

engine/src/parameters.rs
```diff
@@ -130,13 +130,42 @@ impl SubmitResult {
...
+#[derive(BorshSerialize, BorshDeserialize)]
+pub enum CallArgs {
+    New(FunctionCallArgs),
+    Legacy(FunctionCallArgsLegacy),
+}
+
+impl CallArgs {
+    pub fn deserialize(bytes: &[u8]) -> Option<Self> {
+        if let Ok(value) = FunctionCallArgs::try_from_slice(bytes) {
+            Some(Self::New(value))
```
birchmd, line 160:
```suggestion
        if let Ok(value) = Self::try_from_slice(bytes) {
            Some(value)
```

andrcmdr:
Update feature branch by commit https://github.com/aurora-is-near/aurora-engine/pull/351/commits/a97fbbb8070325176df795a4f359986784dea944

**Implementation for deserialized values from bytes to current or legacy Borsh-encoded parameters for passing to the engine call function, and to provide backward type compatibility:**
- Improved implementation of `deserialize()` method for `CallArgs` to suppot Borsh bytes parsing for new structures in call args enum (for future extensibility reason).
- Made new tests for `CallArgs::deserialize()` and for old data structure with legacy arguments.
- Add comments, with explanation of decisions in implementation of ``CallArgs::deserialize()` and tests for this method.
- Fix existed tests.

engine/src/parameters.rs
andrcmdr: Addressed notes and remarks here - made a commit with improvements and corrections:
https://github.com/aurora-is-near/aurora-engine/pull/351#issuecomment-970794864

birchmd (approved):
Nice work @andrcmdr !

engine/src/parameters.rs
```diff
@@ -130,13 +130,48 @@ impl SubmitResult {
...
 pub struct FunctionCallArgs {
+    pub contract: RawAddress,
+    /// Wei compatible Borsh-encoded value field to attach an ETH balance to the transaction
+    pub value: WeiU256,
+    pub input: Vec<u8>,
+}
+
+/// Legacy Borsh-encoded parameters for the engine `call` function, to provide backward type compatibility
+#[derive(BorshSerialize, BorshDeserialize, Debug, PartialEq, Eq, Clone)]
+pub struct FunctionCallArgsLegacy {
```
joshuajbouw, line 144: Rename to `FunctionCallArgsV1`

engine/src/parameters.rs
```diff
@@ -130,13 +130,48 @@ impl SubmitResult {
     }
 }
 
-/// Borsh-encoded parameters for the `call` function.
-#[derive(BorshSerialize, BorshDeserialize)]
+/// Borsh-encoded parameters for the engine `call` function.
+#[derive(BorshSerialize, BorshDeserialize, Debug, PartialEq, Eq, Clone)]
 pub struct FunctionCallArgs {
```
joshuajbouw, line 135: Rename to `FunctionCallArgsV2`

engine/src/parameters.rs
```diff
@@ -130,13 +130,48 @@ impl SubmitResult {
...
+pub struct FunctionCallArgsLegacy {
     pub contract: RawAddress,
     pub input: Vec<u8>,
 }
 
+/// Deserialized values from bytes to current or legacy Borsh-encoded parameters
+/// for passing to the engine `call` function, and to provide backward type compatibility
+#[derive(BorshSerialize, BorshDeserialize, Debug, PartialEq, Eq, Clone)]
+pub enum CallArgs {
+    New(FunctionCallArgs),
```
joshuajbouw, line 153: `New` -> `V2`
`Legacy` -> `V1`

joshuajbouw (changes requested):

engine/src/parameters.rs
andrcmdr: OK, fixed here https://github.com/aurora-is-near/aurora-engine/pull/351/commits/8e1306750eed1424c19f05d5ff40bfa61a6d0380

engine/src/parameters.rs
andrcmdr: OK, fixed here https://github.com/aurora-is-near/aurora-engine/pull/351/commits/8e1306750eed1424c19f05d5ff40bfa61a6d0380

joshuajbouw (approved):

PR #352, opened 2021-11-15
Workspace (near-blake2) and toolchain update

andrcmdr:

- Fix usage of call to `near_blake2` in `enigne-precompiles` code;
- Fixed workspace dependencies only (fix on exact revision in toml for `near-blake2` crate);
- Updated local workspace (minimal, near-blake2 crate, without update for external dependencies);
- Use of current toolchain (nightly from November 12)
These issues was prevent successful project workspace update and was prevent successful project codebase build as well, 
thus was fixed to make compilation and building the whole project in one `wasm32` executable possible on current nightly toolchain (which tested and works fine in local environment, and should work well in CI environment too).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -122,7 +122,6 @@ dependencies = [
  "aurora-engine-sdk",
  "aurora-engine-types",
  "base64 0.13.0",
- "blake2 0.9.1 (git+https://github.com/near/near-blake2.git)",
  "borsh 0.8.2",
  "byte-slice-cast",
  "ethabi",
@@ -131,6 +130,7 @@ dependencies = [
  "hex",
  "libsecp256k1",
  "logos",
+ "near-blake2",
  "num",
  "primitive-types 0.10.1",
  "rand 0.7.3",
@@ -152,7 +152,6 @@ dependencies = [
  "aurora-engine-sdk",
  "aurora-engine-types",
  "base64 0.13.0",
- "blake2 0.9.1 (git+https://github.com/near/near-blake2.git)",
  "borsh 0.8.2",
  "byte-slice-cast",
  "ethabi",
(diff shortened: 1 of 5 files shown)


andrcmdr:
@joshuajbouw @birchmd @mrLSD @strokovok

Guys, please, take a look onto this PR.

The changes as you can see was really minimal, I should say this is atomic fix, only `near-blake2` was fixed (without this updating of local crates in workspace through `cargo update -v -w` won't work) and toolchain version was bumped up.

And this is really strange!

CI shows issues that was present previously (a long before PR) as well, even before updating the toolchain (I guess these issues was hiding for a long time, due to makefile and environment settings with conditional compilation):

Clippy check breaks 'cause of frequent use of `to_string()` - clippy think these are unnecessary:
https://github.com/aurora-is-near/aurora-engine/runs/4207448406?check_suite_focus=true

Tests breaks 'cause of unwind attribute feature used in `wasmer-runtime-core-near` crate :
https://github.com/aurora-is-near/aurora-engine/runs/4207448409?check_suite_focus=true

Bully build won't work 'cause crates should be recompiled with newer toolchain:
https://github.com/aurora-is-near/aurora-engine/runs/4207448359?check_suite_focus=true

We need to tune CI before toolchain update and make some necessary changes to code, so code would pass all clippy and test checks.

joshuajbouw:
Clippy is updated quite a bit with a lot of new defaults coming in. Its just those coming into place. I got the same issue with the unwind. I thought that was the strangest of all issues while trying to update myself.

Obviously, we need to help our dependencies update.

andrcmdr:
https://github.com/wasmerio/wasmer/issues/2588 - here's the same issue, and solution is in updating `wasmer`.
But updating `wasmer-runtime-core-near` to new `wasmer` seems to be a tough migration to the new large external dependency, and thus a bit risky, and 'cause if this I think we sould decide this as a team, not by my own only.

birchmd:
Closing as this was done in https://github.com/aurora-is-near/aurora-engine/pull/422 and https://github.com/aurora-is-near/aurora-engine/pull/426

PR #357, opened 2021-11-18
Fix:  Meta parsing imports

joshuajbouw:
When running `check --all-targets --all-features`, a compilation error occurs:

```
error[E0432]: unresolved imports `crate::meta_parsing`, `crate::prelude::keccak`
 --> engine-tests/src/tests/meta_parsing.rs:2:12
  |
2 |     crate::meta_parsing::{near_erc712_domain, parse_meta_call, prepare_meta_call_args},
  |            ^^^^^^^^^^^^
  |            |
  |            unresolved import
  |            help: a similar path exists: `aurora_engine::meta_parsing`
3 |     crate::prelude::keccak,
  |     ^^^^^^^^^^^^^^^^^^^^^^ no `keccak` in `prelude`
  ```
  This simply fixes that issue so that the correct imports are being used.

PR #367, opened 2021-11-19
Prelude imports versioning

andrcmdr:
Minor fix to include versioning of imports in prelude for remaining crates (for making it more consistent with prelude in other crates and more idiomatic).
Closely related to fixes in this commit https://github.com/aurora-is-near/aurora-engine/pull/357/commits/e2c11b1e2a6cc61ed3b81c2375ebf13af017133b and PR https://github.com/aurora-is-near/aurora-engine/pull/357

### DIFF
--- a/engine-precompiles/src/prelude.rs
+++ b/engine-precompiles/src/prelude.rs
@@ -1,8 +1,10 @@
-pub use aurora_engine_sdk as sdk;
-pub use aurora_engine_types::account_id::*;
-pub use aurora_engine_types::parameters;
-pub use aurora_engine_types::storage;
-pub use aurora_engine_types::types;
-pub use aurora_engine_types::*;
-
-pub use borsh::{BorshDeserialize, BorshSerialize};
+mod v0 {
+    pub use aurora_engine_sdk as sdk;
+    pub use aurora_engine_types::account_id::*;
+    pub use aurora_engine_types::parameters;
+    pub use aurora_engine_types::storage;
+    pub use aurora_engine_types::types;
+    pub use aurora_engine_types::*;
+    pub use borsh::{BorshDeserialize, BorshSerialize};
+}
+pub use v0::*;
--- a/engine-sdk/src/prelude.rs
+++ b/engine-sdk/src/prelude.rs
@@ -1,3 +1,6 @@
-pub use aurora_engine_types::types::{NearGas, PromiseResult, STORAGE_PRICE_PER_BYTE};
-pub use aurora_engine_types::{vec, Address, Vec, H256};
-pub use borsh::{BorshDeserialize, BorshSerialize};
+mod v0 {
+    pub use aurora_engine_types::types::{NearGas, PromiseResult, STORAGE_PRICE_PER_BYTE};
+    pub use aurora_engine_types::{vec, Address, Vec, H256};
+    pub use borsh::{BorshDeserialize, BorshSerialize};
+}
+pub use v0::*;


joshuajbouw:
I don't think we should do this in prelude, but should be done in something like an `imports.rs`.

andrcmdr:
> I don't think we should do this in prelude, but should be done in something like an `imports.rs`. It never was intentional for us to use prelude as an import. That should be what we are exporting of internal public types and functions.
> 
> What should be done is to change the title and scope to reflect this on all our crates.

@joshuajbouw 

I thought this is an internal team agreement to use versioning in module's imports to fix version of imports for usage in a crate - this is a good practice, 'cause it structured imports and improve version management for internal abstractions, types, functions, etc. in imports.

The same approach actually used in most of our crates:

https://github.com/aurora-is-near/aurora-engine/blob/develop/engine/src/prelude.rs#L1

https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-types/src/lib.rs#L11

https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-tests/src/prelude.rs#L1

But not here, which leads to inconsistency in a project - if this approach used for most of our crates let's then follow the agreements and make this for every crate:

https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-sdk/src/prelude.rs

https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-precompiles/src/prelude.rs

And this is the intention of this PRs minor fix.
@birchmd advised to make this PR, for fix versioning of imports in prelude, separated from https://github.com/aurora-is-near/aurora-engine/pull/357, which actually was accepted and merged today (with the same approach to versioning of imports in prelude)!

And it's not clear what's would be improved if instead of `prelude.rs` we would use `imports.rs`? Or I'm not understand this right.

joshuajbouw:
Typically in rust what you are exporting of commonly used types is in a prelude. But we are importing. Therefore, it's never made sense for us to have imports in a prelude.

andrcmdr:
I just thought initially about keeping consistency: either we would use versioning in prelude imports in all crates or not, but not both approaches at the same time in our crates, i.e. like this:
```
pub use aurora_engine_sdk as sdk;
pub use aurora_engine_types::account_id::*;
pub use aurora_engine_types::parameters;
pub use aurora_engine_types::storage;
pub use aurora_engine_types::types;
pub use aurora_engine_types::*;
pub use borsh::{BorshDeserialize, BorshSerialize};
```
or like that:
```
mod v0 {
    pub use aurora_engine_sdk as sdk;
    pub use aurora_engine_types::account_id::*;
    pub use aurora_engine_types::parameters;
    pub use aurora_engine_types::storage;
    pub use aurora_engine_types::types;
    pub use aurora_engine_types::*;
    pub use borsh::{BorshDeserialize, BorshSerialize};
}
pub use v0::*;
```
but not both approaches in different crates, 'cause these leads to duality in UX with types in imports.

andrcmdr:
Right, now I perceived an idea! Versioning are for exports, and exported types and abstractions can be taken out to prelude for more wide use, as an API for library!

**And an important thing here is that versioning is a good practice, for exports (which is maybe taken out to prelude), as for a library APIs, provided to others, which would be used frequently from different part of the projects, this case it is essential. But imports versioning and imports in prelude obviously not - it's a bad practice, which doesn't make any sense, but even can be destructive by potentially lead to type collisions in imports with the same path parts, which would be hidden by prelude, and same type names!**

And I just forgot how I used preludes in a Zebra:
https://github.com/andrcmdr/zebra/blob/zcash-zebra/ibcrun/src/prelude.rs
https://github.com/andrcmdr/zebra/blob/zcash-zebra/ibclib/src/prelude.rs
This was exactly for exporting useful types and methods!

I was using preludes a year ago, in Zebra, exactly for exporting useful types and methods - this pattern was used for integration of Zebra node as a library (through wrapper - IBC middleware library) to other node and ZCash oracle.

Here:
https://github.com/andrcmdr/zebra/blob/zcash-zebra/ibcrun/src/prelude.rs
https://github.com/andrcmdr/zebra/blob/zcash-zebra/ibclib/src/prelude.rs
Then we can just use:
https://github.com/andrcmdr/zebra/blob/zcash-zebra-aslib/src/zcash_zebra_aslib.rs#L29 (made this as an example)
Just like this:
```
use ibcrun::prelude::*;
use ibclib::prelude::*;
```

I was relying on these docs:
https://doc.rust-lang.org/std/prelude/index.html
https://doc.rust-lang.org/std/io/prelude/index.html
https://doc.rust-lang.org/std/prelude/v1/index.html
https://docs.rs/async-std/1.9.0/async_std/prelude/index.html
https://docs.rs/tokio/0.3.5/tokio/prelude/index.html

I found one example on why usage an imports in preludes considered as a bad practice in many advanced projects.
For example here, we have `aurora_engine::parameters`:
https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-tests/src/prelude.rs#L5
and here, we have `aurora_engine_types::parameters::*`:
https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-tests/src/prelude.rs#L8
**And when we use `crate::prelude::parameters::MetaCallArgs` here (and this is full path to single type and thus fine!) or some other type from `crate::prelude::parameters`, and there wouldn't be obvious, without fully qualified type path, which type from which module would be used, 'cause prelude hiding that, and this 'cause and lead us to type collisions for types with the same names and same parts in its path (if we would have such similarly named types):**
https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-tests/src/tests/meta_parsing.rs#L3

**Imports in prelude, without fully qualified path to types on a call site, covers and hiding path to types with same last parts in path or with the same names, which potentially could lead to type collisions and multiple type definitions, and problems even on linkage stage (with ld).**

And that's why the usage of imports from preludes considered as bad practice in many projects.

So, we would eliminate imports in prelude, and will use prelude for export of widely useful types/abstractions/functions/methods/etc., and versioning them, as library/crate API.

joshuajbouw:
Yeah as said in chat, its not good practice what we are doing, and doesn't make sense to reinforce it. Just needs to be fixed.

Issue #297, opened 2021-10-08
Expose NEAR's randomness host function as custom precompile


birchmd:
On Ethereum people need to use oracles for randomness (e.g. https://twitter.com/KamesCG/status/1446478366668955661). On NEAR there is a randomness host function we could expose as a custom precompile on Aurora, making it easier to access randomness.

This idea was proposed by @mikedotexe

mikedotexe:
Even better than my idea, great stuff.
Here's the function from the Rust SDK side:
https://docs.rs/near-sdk/3.1.0/near_sdk/env/fn.random_seed.html

birchmd:
Done in #368

PR #368, opened 2021-11-20
NEAR random numbers from solidity contract

mfornet:
Add Precompile that allows calling random numbers from solidity
contracts.

Closes https://github.com/aurora-is-near/aurora-engine/issues/297


joshuajbouw:
Do double check the tests as they are failing, thanks.

engine-precompiles/src/random.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub const ADDRESS: Address =
+        super::make_address(0xf8615118, 0x15955326b953fa97b6955a2f8020a4e9);
+}
+
+impl Precompile for RandomBytes {
+    fn required_gas(_input: &[u8]) -> Result<u64, ExitError> {
+        Ok(costs::RANDOM_BYTES_GAS)
+    }
+
+    #[cfg(feature = "contract")]
```
birchmd, line 35: Now that we have the standalone engine I think it is good to try to avoid implementations which require NEAR host functions. In this case we can simply store the seed as a field in the `RandomBytes` struct. This pushes the NEAR call up one level to the precompile initialization. Similarly there we can have the random seed be an input to creating precompiles, pushing it back another level. Ultimately the host function is only called at the boundary of the application (`engine/lib.rs` in the wasm contract), or in the case of the standalone engine we could derive the seed from the NEAR block information, and the value is passed down into the core logic.

I think it would be good if we could eliminate this `contract` feature from the precompiles crate entirely. This requires a little refactoring of the native precompiles implementation, but I think it is doable.

engine-precompiles/src/random.rs
```diff
@@ -0,0 +1,85 @@
...
+    ) -> EvmPrecompileResult {
+        let cost = Self::required_gas(input)?;
+        if let Some(target_gas) = target_gas {
+            if cost > target_gas {
+                return Err(ExitError::OutOfGas);
+            }
+        }
+
+        Ok(
+            PrecompileOutput::without_logs(cost, sdk::random_seed().to_fixed_bytes().to_vec())
```
birchmd, line 50: I think we will need some very clear documentation around this precompile. It is not a random bytes generator in the sense that repeated calls to this address in the same transaction will return the same value. This is a per-block entropy source which could then be used to create a random sequence.

Speaking of documentation, do we have any documentation of the Aurora custom precompiles? If not we should make an issue to write this.

engine-precompiles/src/random.rs
```diff
@@ -0,0 +1,85 @@
...
+use evm::{Context, ExitError};
+
+mod costs {
+    use crate::prelude::types::Gas;
+
+    // TODO(#51): Determine the correct amount of gas
+    pub(super) const RANDOM_BYTES_GAS: Gas = 0;
+}
+
+const ERR_TARGET_TOKEN_NOT_FOUND: &str = "Target token not found";
```
birchmd, line 17: Unused constant?

engine-precompiles/src/random.rs
```diff
@@ -0,0 +1,85 @@
...
+
+pub struct RandomBytes; //TransferEthToNear
```
birchmd, line 19: Outdated comment

birchmd (commented):
Thanks for taking initiative on this!

birchmd:
@mfornet please also retarget this PR to `develop`

engine-precompiles/src/random.rs
mfornet: Done.

engine-precompiles/src/random.rs
mfornet: I updated the docstring for now, but I agree that it would be super helpful to have a precompile documentation.

joshuajbouw:
84 files changes o.O.

mfornet:
Changed target branch to develop @birchmd @joshuajbouw

joshuajbouw:
> Changed target branch to develop @birchmd @joshuajbouw

Thanks! Will check tonight.

engine-sdk/src/lib.rs
```diff
@@ -41,6 +41,21 @@ pub fn log_utf8(bytes: &[u8]) {
         exports::log_utf8(bytes.len() as u64, bytes.as_ptr() as u64);
     }
 }
+/// Calls random seed produced by validators
+#[cfg(feature = "contract")]
+pub fn random_seed() -> H256 {
```
birchmd, line 46: I wonder if it makes sense to add this to `sdk::env::Env` instead of as a standalone sdk function.

This way we can have an implementation for it in the standalone engine too.

engine-tests/src/test_utils/random.rs
```diff
@@ -0,0 +1,81 @@
...
+        c.0
+    }
+}
+
+pub(crate) struct Random {
+    contract: solidity::DeployedContract,
+}
+
+impl Random {
+    pub fn random_seed(&self, runner: &mut AuroraRunner, signer: &mut Signer) -> Option<U256> {
```
birchmd, line 48: Seems unnecessary to return an `Option` here when we unwrap all the errors except the return length. Since this is only for tests I think it is fine to return simply `U256` (panicking if the result length is not 32 bytes).

As a side note, I don't think `U256` is the right type. Yes, a 32-byte array is returned but semantically I don't think we would normally interpret it as a number. `H256` or even just `[u8; 32]` would be better in my opinion.

engine-tests/src/test_utils/mod.rs
```diff
@@ -406,6 +407,11 @@ impl AuroraRunner {
             Ok(bytes)
         }
     }
+
+    pub fn with_random_seed(mut self, random_seed: Vec<u8>) -> Self {
```
birchmd, line 411: I think it would be nicer if this took `[u8; 32]` and turned it into a `Vec` when passing to the context. The seed is supposed to have a fixed size.

engine/src/engine.rs
```diff
@@ -311,8 +312,13 @@ struct StackExecutorParams {
 
 impl StackExecutorParams {
     fn new(gas_limit: u64, current_account_id: AccountId) -> Self {
+        let random_seed = sdk::random_seed().to_fixed_bytes().to_vec();
```
birchmd, line 315: Again here I think it is better to take the random seed as an argument. We want to push sdk access all the way out to `lib.rs`, or if it must happen in core logic then only through one of the sdk traits (in this case I think `sdk::env::Env` makes sense if you are going to use a trait).

engine-sdk/src/lib.rs
```diff
@@ -41,6 +41,21 @@ pub fn log_utf8(bytes: &[u8]) {
...
+        let bytes = H256::zero();
+        exports::read_register(0, bytes.0.as_ptr() as *const u64 as u64);
+        bytes
+    }
+}
+
+#[cfg(not(feature = "contract"))]
+pub fn random_seed() -> H256 {
+    H256::zero()
+}
```
birchmd, line 58: Regardless of whether we pass `random_seed` all the from `engine/src/lib.rs` into core logic, or if you implement it as part of the `Env` trait, I think this implementation should be deleted. We must have feature parity between the wasm engine and the standalone engine otherwise the debug tracing will sometimes return wrong results.

birchmd (changes requested):
Thanks, this is definitely a step in the right direction. But it does not go far enough. We must have feature parity between the wasm engine and the standalone engine otherwise the debug tracing will sometimes return wrong results. So the sdk access needs to be pushed all the way to the boundary of the application, or it needs to be mediated by an sdk trait.

engine-sdk/src/lib.rs
mfornet: I missed the Env trait, will implement it there instead. But I wonder what is the rule of thumb here. I was looking for the implementation of `sha256` and implementing `random_seed` next to it.

engine-tests/src/test_utils/random.rs
mfornet: Refactored everything to use `H256`.

engine-tests/src/test_utils/mod.rs
mfornet: Refactored everything to use `H256`.

engine/src/engine.rs
mfornet: Got the idea now. Hopefully this time I got it right. To make sure we don't have this problem in the future, we should consider changing visibility from `sdk` to not be used at this point.

engine-sdk/src/lib.rs
mfornet: Done

engine-sdk/src/lib.rs
birchmd: The difference with `sha256` is that we have a correct implementation both with and without the `contract` feature because it is a pure mathematical function. That is the core principle: we must have the same behaviour for all functions in the wasm contract as in the standalone engine.

birchmd (approved):
LGTM nice work!

joshuajbouw (commented):
Fine.

mikedotexe:
Wow, great work folks!

vans163:
Came here from the challenge on the NEARBuild, can there be a little more explanation for how this is truly random? IIUC
```
let result = runner.submit_transaction(&signer.secret_key, tx).unwrap();
```
The random_seed is just the hash of the signed block or transaction? Meaning if the runner or validator is compromised and collaborating, they can craft the seed? Maybe I am understanding wrong.

mikedotexe:
> The random_seed is just the hash of the signed block or transaction? Meaning if the runner or validator is compromised and collaborating, they can craft the seed? Maybe I am understanding wrong.

My understanding (from someone who does not work closely with `nearcore`) is that this is a verifiable random function. I think @mfornet can explain this really well, but in terms of the validator's influence, it's my understanding that they cannot simply "make up" some random bytes, it is verifiable.

However, at the moment, validators **can** take a look at the randomness and decide, "nah, this doesn't benefit me or my friends, and I don't want to include my work in the blockchain." So in that sense, validators could arguably give preferential treatment, but with a 2-second block time, it seems to me that other validators would be happy to process it normally.

So I feel like it's still quite resistant to tampering unless there's widespread validator collusion.

Again with the disclaimer that I don't know what I'm talking about, but have asked this question and am relaying these details from memory.

vans163:
> > The random_seed is just the hash of the signed block or transaction? Meaning if the runner or validator is compromised and collaborating, they can craft the seed? Maybe I am understanding wrong.
> 
> My understanding (from someone who does not work closely with `nearcore`) is that this is a verifiable random function. I think @mfornet can explain this really well, but in terms of the validator's influence, it's my understanding that they cannot simply "make up" some random bytes, it is verifiable.
> 
> However, at the moment, validators **can** take a look at the randomness and decide, "nah, this doesn't benefit me or my friends, and I don't want to include my work in the blockchain." So in that sense, validators could arguably give preferential treatment, but with a 2-second block time, it seems to me that other validators would be happy to process it normally.
> 
> So I feel like it's still quite resistant to tampering unless there's widespread validator collusion.
> 
> Again with the disclaimer that I don't know what I'm talking about, but have asked this question and am relaying these details from memory.

This is not exactly what I am referring to. It is not about verifiable randomness, but about if there is the ability (at some level) to craft an environment where the randomness can be tampered with.

For example if I understand right (which could be fully wrong), the random_seed is based on the hash of the wrapped transaction/block aurora produces.  So if there was say a coinflip contract that did something like this to get random number between 0 and 1.
```
seed_random().next(0,1)
```
I know based on looking at the next function for the random implementation that any hash at or over 0x7fffffff is a 1 and under  is a 0. 

So if I need a 1 to win, and am the validator / runner on aurora and I take a hash of the next block/tx I am producing and the seed_random is 0x7ffffffe, this will make me lose, so i include an extra "dummy" transaction into the block at the last moment to have it hash different, until I get something at or over 0xffffffff, making me win the dice game in the block I just validated/rolledup.

mfornet:
tl;dr The following is not possible

> so i include an extra "dummy" transaction into the block at the last moment to have it hash different, until I get something at or over 0xffffffff,

----
@vans163 random numbers on NEAR doesn't depend on the content of the block, only on the previous random number and the private key of the current block producer (BP) building the block. [Code](https://github.com/near/nearcore/blob/master/core/primitives/src/block.rs#L249-L250).

As Mike pointed earlier, this means that all a BP can do to affect the random number is not producing a block (or in case it produce several blocks in a row, drop a subset of them). This means that BP can bias the random number, but only have limited ability for this (changing the content of the block won't affect the output random number).

There is this proposal (https://github.com/near/NEPs/discussions/206) that address this issue.

Btw, note there are no such things a Aurora validators or Aurora BP, only NEAR BP (in some sense you may say that NEAR BP, are Aurora BP as well).

------

Short term impact of using random numbers while previous proposal is not live. Validators trying to bias random number, will leave traces, at least half of the time they will be dropping blocks. They put rewards and reputation at risk. If they start dropping too many blocks most likely BP will lost delegated stake and will have even less stake to participate or bias random numbers.

vans163:
This is very enlightening thank you, I did not realize how complex this could get.  It seems that NEP is critical to have then.  

Sorry if my understanding is poor but this is potentially worse then depending on the content of the block? If I am the validator and know I have a chance at getting chosen to produce the next block, I know the previous random number, and I know what next random number I am about to generate as I know my own private key.  I know exactly if I get chosen what the seed will be in the block I will produce, I can then choose to include/discard certain transaction?

mfornet:
> I know the previous random number, and I know what next random number I am about to generate as I know my own private key. I know exactly if I get chosen what the seed will be in the block I will produce, I can then choose to include/discard certain transaction?

This is correct. 

To implement a fair bet-like game it is required that the value committed is fixed before the random seed is produced. This can be done in multiple steps, using either cross-contract calls or submitting 2 different transactions. However achieving this is still challenging, given that validators that submit 2 txs in a row can commit, and reveal sequentially. I'll try to think if there is a way in the current system to go around this, otherwise we will need to wait until the **New randomness beacon** lands in NEAR mainnet.

vans163:
Yea because collaborating with 1 validator is pretty easy, especially now since the stake required is quite low.  But what that NEP proposes is that collaboration with 1/3 of the validators is required and that is much harder now, hard enough to make it worth using to avoid the hassle off the 2tx hash/reveal or other methods of achieving randomness.

Also I see the possible damage done by the current implementation is VERY low, gaming the system is not possible as the best we can do is include/exclude a transaction but we cant change the random_seed. So if say we have a booster pack NFT game, where you open a card pack and a ultra-rare rinosaur has a 1/100_000_000 chance of being pulled, the best we can do is not open the pack if we dont have the rinosaur, but we cant keep spoofing the block we are about to produce so it hashes to a random_seed that will make the pack contain the card.  Unless ofcourse on the law of averages we go through 160_000_000+ blocks on our validator until the random_seed is exactly what we want.

But for a coinflip game, or 50/50 random, the current implementation is definitely very gameable.

PR #374, opened 2021-11-22
Connector cleanup

mrLSD:
Cleanup Eth-connector and improve code quality.

Full description here: #359


engine/src/deposit_event.rs
```diff
@@ -1,11 +1,105 @@
...
+                .to_string()
+        } else {
+            message
+        };
+        let address_bytes = validate_eth_address(address)
+            .map_err(ParseEventMessageError::EthAddressValidationError)?;
+        // Second data section should contain Eth address
+        data.extend(address_bytes);
+        // Add `:` separator between relayer_id and data message
+        Ok([relayer_account_id.as_ref(), &hex::encode(data)].join(":"))
```
birchmd, line 99: I think it would be nice if there was a type like

```
struct FtOnTransferMessage {
    fee: Fee,
    address: Address,
    relayer_account_id: AccountId,
}
```

This type would be constructed in this function and passed around through the core logic. Only at the end where we are about to do the NEAR call that requires it to be a string, then we would serialize it in this way.

engine/src/deposit_event.rs
```diff
@@ -91,28 +184,43 @@ impl DepositedEvent {
...
+        let event_message_data: String = event.log.params[1].value.clone().to_string();
+
         let amount = event.log.params[2]
             .value
             .clone()
             .into_uint()
-            .ok_or(error::ParseError::InvalidAmount)?;
-        let fee = event.log.params[3]
+            .ok_or(error::ParseError::InvalidAmount)?
+            .as_u128();
```
birchmd, line 196: We should catch the overflow case here.

engine/src/parameters.rs
```diff
@@ -160,12 +161,12 @@ impl CallArgs {
         // made for flexibility and extensibility.
         if let Ok(value) = Self::try_from_slice(bytes) {
             Some(value)
-        // Fallback, for handling old input format,
-        // i.e. input, formed as a raw (not wrapped into call args enum) data structure with legacy arguments,
-        // made for backward compatibility.
+            // Fallback, for handling old input format,
+            // i.e. input, formed as a raw (not wrapped into call args enum) data structure with legacy arguments,
+            // made for backward compatibility.
```
birchmd, line 166: Random indentation change?

engine/src/connector.rs
```diff
@@ -10,61 +10,68 @@ use crate::parameters::{
...
-    pub fee: U256,
+    pub fee: Fee,
 }
 
 impl<I: IO + Copy> EthConnectorContract<I> {
-    pub fn get_instance(io: I) -> Self {
+    /// Init Eth-connector contract instance.
+    /// Load contract data from storage and init I/O handler.
+    /// Used as single point of contract access for various contract actions
+    pub fn init_instance(io: I) -> Self {
```
birchmd, line 74: Nit pick: I don't know if "initialize" is the right verb here. Maybe "create" would be better. I don't have strong feelings either way, but I don't there to be any confusion between this function and `init_contract` (which is only called once).

birchmd (changes requested):
Thanks for your work on this, it's certainly moving in the right direction!

My main concern is about changing the binary format in one of the messages. That kind of backwards-incompatible change could cause an issue in the future. Remember that we want to support replaying transactions in the standalone engine now, so we need to be cognizant of breaking changes.

mrLSD:
> My main concern is about changing the binary format in one of the messages.

From that point of view, it makes sense. I'll return to the old binary format.

An interesting fact, out of 32 bytes - 16 bytes are superfluous. since we don't use them.

engine/src/connector.rs
mrLSD: I thought about it. I thought that "creation" is not an accurate name. 
We load previously created data. Initialization more accurately reflects the essence of what is happening. From this point of view, it would be more successful to rename `init_contract` to `create_contract`.

engine/src/fungible_token.rs
```diff
@@ -317,7 +335,7 @@ impl<I: IO + Copy> FungibleTokenOps<I> {
         sender_id: &AccountId,
         receiver_id: &AccountId,
         amount: Balance,
-    ) -> (u128, u128) {
+    ) -> (Balance, Balance) {
```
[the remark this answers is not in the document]
mrLSD: I fully support it. However, the `internal_ft_resolve_transfer` function is related to the reference implementation of NEP-141. By optimizing it, we thereby move away from the reference implementation. This has its obvious pros and cons. 
@joshuajbouw let's do a final decision.

engine/src/deposit_event.rs
mrLSD: Actually, we don't need that. It's used **ONLY** for:
```
pub struct TransferCallCallArgs {
    pub receiver_id: AccountId,
    pub amount: Balance,
    pub memo: Option<String>,
    pub msg: String,
}
```

Mentioned logic for `prepare_message` solves the problem with early as possible validation data. We early prepare message string for TransferCall args. I can't see any reason to introduce a new type, because we never used it.

engine-types/src/types.rs
```diff
@@ -183,6 +295,10 @@ impl Wei {
     pub fn checked_add(self, rhs: Self) -> Option<Self> {
         self.0.checked_add(rhs.0).map(Self)
     }
+
+    pub fn into_u128(self) -> u128 {
+        self.0.as_u128()
+    }
```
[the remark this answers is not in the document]
mrLSD: Good point for discussion. Before that, we used `as_u128` instead. When it's overflow it will panics.
There is really long path for a lot of core functions without Result type.
So we can use `low_u128`, but it doesn't check overflow.
Or introduce Result type for a lot of core Engine functions.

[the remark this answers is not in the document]
birchmd: Is it that much work to introduce `Result`? Lots of functions already return `Result` now after the recent refactors, so hopefully adding a `Result` here would only bubble up to being a new error variant on some functions that are already returning errors.

engine/src/deposit_event.rs
birchmd: The reason to return structured data as a structured type instead of something opaque like a `String` is because it makes the logic more clear. Without reading the function implementation, you would never know that the `String` which is being returned is actually a structured thing.

The principle of pushing IO all the way to boundary of applications applies both to the Input and the Output sides of the IO. Yes, we want to validate input early and only pass strongly typed structures in our core logic. And similarly we want to hold structed data in strong types until the very end, where we may need to serialize it in some way to communicate with the outside world.

mrLSD: With this approach, we ensure that the data is prepared for sending in the correct way. There is no problem to add an additional type, and to it a method that would generate this binary data. However, it will play the role of code readability. No more. In backward compatibility conditions, the input and output messages do not change. The new type itself will have no function other than to make the code easier to read.

birchmd: Yes, exactly. And easier to read code is a good thing.

mrLSD: For that was reused and reorganized type `OnTransferMessageData`

birchmd: If `OnTransferMessageData` is now used for multiple purposes perhaps a more generic name is needed? Ie that struct is not always used with `ft_on_transfer` anymore, so its name is now misleading.

engine-types/src/types.rs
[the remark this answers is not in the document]
mrLSD: I spent some time analyzing where we can use it. The problem is in point related to the Balance key prefix. We use the same prefix for `u128` and `U256` balances. But I found a solution for that. It was committed.

engine-types/src/types.rs
```diff
@@ -104,6 +108,114 @@ impl Mul<EthGas> for u64 {
...
+}
+
+impl Mul<Fee> for Fee {
+    type Output = Fee;
+
+    fn mul(self, rhs: Fee) -> Self::Output {
+        Fee(self.0 * rhs.0)
+    }
+}
```
birchmd, line 165: I don't think this implementation should exist. A fee times a fee is not a fee, it's some kind of "fee area" (or similar nonsense concept).

When you're defining a new type you are giving semantics to a type; it's not a pure number anymore. Imagine if you were defining a `Meter` new type. This `impl Mul<Meter> for Meter` would not output `Meter`, it would output `SquareMeter` because areas are semantically different from lengths.

The reason it is important to not have implementations which break the semantics of the type is because otherwise it would be possible to write buggy logic. If you find yourself needing to do a `Fee * Fee` operation then the compiler should flag to you that it is not allowed because it means either you have assigned the types wrong (one of the arguments is actually a pure number, not a fee) or something is logically incorrect in the computation.

Note it is ok to have `impl Mul<u128> for Fee` because `u128` is still a pure number. This implementation is effectively a "scalar multiplication" (in mathematics jargon), or you could think about it as a short-cut for repeated addition (and of course `impl Add<Fee> for Fee` should output another `Fee`).

engine-types/src/types.rs
```diff
@@ -104,6 +108,114 @@ impl Mul<EthGas> for u64 {
...
+
+impl Div<Fee> for Fee {
+    type Output = Fee;
+
+    fn div(self, rhs: Fee) -> Self::Output {
+        Fee(self.0 / rhs.0)
+    }
+}
```
birchmd, line 189:
```suggestion
impl Div<Fee> for Fee {
    type Output = u128;

    fn div(self, rhs: Fee) -> Self::Output {
        self.0 / rhs.0
    }
}
```

When you divide some unit by itself you get back a pure number. For example, pi (the mathematical constant) is a pure number which is defined to be a ratio (division) of two lengths (circumference and diameter).

engine-types/src/types.rs
```diff
@@ -104,6 +108,114 @@ impl Mul<EthGas> for u64 {
...
+
+impl Div<Fee> for u128 {
+    type Output = Fee;
+
+    fn div(self, rhs: Fee) -> Self::Output {
+        Fee(self / rhs.0)
+    }
+}
```
birchmd, line 197: This implementation also should not exist. The output would be an "inverse fee" (or similar nonsense concept). See my rant about `Fee * Fee` for more details about why getting the semantics right is important.

engine-types/src/types.rs
```diff
@@ -104,6 +108,114 @@ impl Mul<EthGas> for u64 {
...
+
+impl Add<u128> for Fee {
+    type Output = Fee;
+
+    fn add(self, rhs: u128) -> Self::Output {
+        Fee(self.0 + rhs)
+    }
+}
```
birchmd, line 157: One could argue that these implementations should not exit. You could think of them as a short-cut for `fee + Fee::from(number)`, but I think it is better if the conversion from number to fee is not done implicitly. It may prevent logic bugs if wrapping numbers in the Fee type must always be done explicitly.

engine-types/src/types.rs
```diff
@@ -104,6 +108,114 @@ impl Mul<EthGas> for u64 {
...
+    fn from(fee: u128) -> Self {
+        Self(fee)
+    }
+}
+
+impl From<Fee> for u128 {
+    fn from(fee: Fee) -> Self {
+        fee.0
+    }
+}
```
birchmd, line 217: Nit: This is less important than my other comments, but I also think implementation should not exist. All throughout our core logic fees are fees and so will never need to be unwrapped back to the raw numeric type. It is only at the boundary where we need to pass the value out of our application where we will need this conversion. This means it should not be needed very often and therefore having a super convenient way to do it is not needed.

On the other hand, convenience can erode the safety of new types because it allows the developer to change between types without needing to think about it, even if they _should_ be thinking about it.

Instead of `impl From<Fee> for u128` I prefer to have a function like

```rust
impl Fee {
  pub fn raw(self) -> u128 { self.0 }
}
```

The name "raw" is meant to suggest that you are dropping the semantics of the new type to work directly with the underlying numeric type. And again this should only happen at the boundaries of our application.

engine-types/src/types.rs
```diff
@@ -314,10 +433,35 @@ impl<T> Stack<T> {
...
     str::from_utf8(inp).unwrap()
 }
 
+pub mod error {
+    use crate::{fmt, String};
+
+    #[derive(Eq, Hash, Clone, Debug, PartialEq)]
+    pub enum BalanceOverflowError {
+        Overflow,
+    }
```
birchmd, line 447:
```suggestion
    pub struct BalanceOverflowError;
```

An enum with a single variant seems unnecessarily complex to me compared with a simple unit struct.

engine/src/connector.rs
```diff
@@ -680,15 +617,15 @@ impl<I: IO + Copy> EthConnectorContract<I> {
             .return_output(&self.ft.get_accounts_counter().to_le_bytes());
     }
 
-    /// Save eth-connector contract data
+    /// Save eth-connector fungible toke contract data
```
birchmd, line 620:
```suggestion
    /// Save eth-connector fungible token contract data
```

engine/src/engine.rs
```diff
@@ -1136,12 +1141,14 @@ pub fn set_balance<I: IO>(io: &mut I, address: &Address, balance: &Wei) {
 }
 
 pub fn remove_balance<I: IO + Copy>(io: &mut I, address: &Address) {
-    let balance = get_balance(io, address);
+    // The `unwrap` is safe here because if the connector
+    // is implemented correctly then the "Eth on Aurora" wll never underflow and
```
birchmd, line 1145: Looks like the comment is incomplete

engine/src/fungible_token.rs
```diff
@@ -126,11 +142,12 @@ impl<I: IO + Copy> FungibleTokenOps<I> {
         }
     }
 
-    /// Balance of ETH (ETH on Aurora)
-    pub fn internal_unwrap_balance_of_eth_on_aurora(&self, address: EthAddress) -> Balance {
-        engine::get_balance(&self.io, &Address(address))
-            .raw()
-            .as_u128()
+    /// Balance of ETH (ETH on Aurora)###
```
birchmd, line 145:
```suggestion
    /// Balance of ETH (ETH on Aurora)
```

birchmd (changes requested):
I think this PR is nearly ready! In this round of comments I want you delete some of the implementations on the `Fee` new type because they do not properly follow the semantics of its arithmetic.

engine-types/src/types.rs
```diff
@@ -3,10 +3,14 @@ use borsh::{BorshDeserialize, BorshSerialize};
 
 use crate::fmt::Formatter;
 
+// TODO: introduce new Balance type for more strict typing
```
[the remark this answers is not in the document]
mrLSD: It will be new PR.

engine-types/src/types.rs
mrLSD: The fact is, we are not using multiplication at this time.

Otherwise, I disagree. A simple example - we can do Watts.
2 watts * 3 watts = 6 watts.

This means operating with a single entity, the result of which is the same entity.

engine-types/src/types.rs
mrLSD: We also do not use division. I can remove it.

engine-types/src/types.rs
mrLSD: Removed

engine-types/src/types.rs
mrLSD: Was removed

birchmd: Thanks for removing.

Watts times Watts is not Watts; it would be Watts^2, which is not a unit of power. When you multiply quantities you get a quantity which has the units of both inputs. When you add quantities, you and only add ones with the same units. This is an application of [dimensional analysis](https://en.wikipedia.org/wiki/Dimensional_analysis).

engine/src/engine.rs
```diff
@@ -1136,12 +1141,14 @@ pub fn set_balance<I: IO>(io: &mut I, address: &Address, balance: &Wei) {
 }
 
 pub fn remove_balance<I: IO + Copy>(io: &mut I, address: &Address) {
-    let balance = get_balance(io, address);
+    // The `unwrap` is safe here because if the connector
+    // is implemented correctly then the "Eth on Aurora" wll never underflow.
+    let balance = get_balance(io, address).try_into_u128().unwrap();
```
birchmd, line 1146: I think a better way to phrase this justification would be something like

"All ETH on Aurora originates from transfers over the bridge into the NEP-141 contract on NEAR. In particular, all ETH in the EVM will be held by the `aurora` account in that NEP-141 contract. Since the NEP-141 standard represents balances as u128, it means there can never be more than u128::MAX total ETH in the EVM (if someone were to try, the NEAR transaction would fail). Therefore there can never be an EVM address with more than u128::MAX either."

But I agree, the unwrap here is safe in practice.

birchmd (approved):
Thanks @mrLSD nice job.

joshuajbouw (approved):
Looks good to me. In the future, please do changes in different PRs. Makes it much easier for the reviewers to go through.

engine-types/src/types.rs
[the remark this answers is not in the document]
mrLSD: Actually, for that, I introduced a special method:
`into_u128`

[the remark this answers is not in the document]
mrLSD: I was a bit confused. 
```
impl From<Fee> for u128
```
It was removed. But that part of code marked in your comment as new code part.

engine-types/src/types.rs
```diff
@@ -183,6 +295,10 @@ impl Wei {
     pub fn checked_add(self, rhs: Self) -> Option<Self> {
         self.0.checked_add(rhs.0).map(Self)
     }
+
+    pub fn into_u128(self) -> u128 {
```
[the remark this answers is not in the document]
mrLSD: I believe we should have the same idiomatic way everywhere. Currently, we have the same approach `into_u64` for types:
* `NearGas`
* ` EthGas`

engine/src/connector.rs
```diff
@@ -120,88 +129,57 @@ impl<I: IO + Copy> EthConnectorContract<I> {
...
-            .map_err(|_| error::ParseEventMessageError::InvalidAccount)?;
-        if data.len() == 1 {
-            Ok(TokenMessageData::Near(account_id))
-        } else {
-            Ok(TokenMessageData::Eth {
-                receiver_id: account_id,
-                message: data[1].into(),
-            })
-        }
-    }
```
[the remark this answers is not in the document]
mrLSD: For early validation in `deposit_event`. It's guaranty data correctness early, also I introduce a new type for that and data consistency.

engine/src/connector.rs
```diff
@@ -120,88 +129,57 @@ impl<I: IO + Copy> EthConnectorContract<I> {
...
-                .ok_or(AddressValidationError::FailedDecodeHex)?
-                .to_string()
-        } else {
-            message
-        };
-        let address_bytes =
-            hex::decode(address).map_err(|_| AddressValidationError::FailedDecodeHex)?;
-        data.extend(address_bytes);
-        Ok([relayer_account_id.as_ref(), &hex::encode(data)].join(":"))
-    }
```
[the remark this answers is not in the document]
mrLSD: Same logic - early validation (as it is possible), and added new types for that.

PR #375, opened 2021-11-23
Feat(standalone): Storage backend

birchmd:
Lays out the necessary logic for the rocksdb backend. This implementation follows @mfornet 's suggestion (using labelled keys instead of snapshots + diffs). Diffs are still persisted because they might be useful for other things.

~I've added some basic tests, and I'll write more before this PR gets merged. But it's a relatively large PR so I wanted to get it out there to give others time to review it.~ Edit: I'm happy with the test coverage now.

~I also need to fix some clippy warnings that popped up because I wanted to import test utils into the standalone crate. Though maybe the better way to do it would be to put standalone tests into the engine-tests crate.~ Edit: moved standalone-storage into its own crate and put the tests for it in engine-tests; this will also enable correctness checks by running the same transaction under both the wasm contract and the standalone backed by the new storage layer.


engine-standalone-storage/src/engine_state.rs
```diff
@@ -0,0 +1,136 @@
...
+            bound_tx_position,
+            transaction_diff,
+            output,
+            db,
+        }
+    }
+
+    pub fn get_transaction_diff(&self) -> Diff {
+        self.transaction_diff.borrow().clone()
+    }
```
joshuajbouw, line 68: If we are returning a clone here, shouldn't it be stored owned?

engine-standalone-storage/src/engine_state.rs
```diff
@@ -0,0 +1,136 @@
...
+#[derive(Debug)]
+pub enum EngineStorageValue<'a> {
+    Slice(&'a [u8]),
+    Vec(Vec<u8>),
+}
+
+impl<'a> AsRef<[u8]> for EngineStorageValue<'a> {
+    fn as_ref(&self) -> &[u8] {
+        match self {
+            Self::Slice(slice) => slice,
```
joshuajbouw, line 17: This ends up being a double reference, which the compiler is going to drop anyways automatically, but normally clippy captures these types of issues in its current version. Would be good to add a single dereference.

engine-standalone-storage/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Get an object which represents the state of the engine at the given block hash,
+    /// after transactions up to (not including) the given transaction index.
+    /// The `input` is the bytes that would be present in the NEAR runtime (normally
+    /// not needed for standlone engine).
+    pub fn get_engine_storage<'db, 'input: 'db>(
```
joshuajbouw, line 148: `take_engine_storage_at_position` would be a better choice here, given what is happening in the method.

joshuajbouw (changes requested):
Just a couple of changes. All in all, looks correct as far as I can tell right now. Good start!

joshuajbouw:
The three storage lifetimes, thats more of a optimisation to prevent cloning of large storage states I take it?

engine-standalone-storage/src/engine_state.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    fn write_storage(&mut self, key: &[u8], value: &[u8]) -> Option<Self::StorageValue> {
+        let original_value = self.read_storage(key);
+
+        self.transaction_diff
+            .borrow_mut()
+            .modify(key.to_vec(), value.to_vec());
+
+        original_value
+    }
```
mfornet, line 119: Should we allow some option to write to storage without reading evicted value! I think this feature is not always necessary, and reading from storage can be slow if the value is not in the diff right?

mfornet: Same with `remove_storage`.

mfornet: I see this is part of IO trait, and probably that involve a bigger change. 

Don't take this comment as blocking, just something that passed through my mind while reading the code.

engine-standalone-storage/src/engine_state.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    fn read_storage(&self, key: &[u8]) -> Option<Self::StorageValue> {
+        if let Some(diff) = self.transaction_diff.borrow().get(key) {
+            return diff
+                .value()
+                .map(|bytes| EngineStorageValue::Vec(bytes.to_vec()));
+        }
+
+        let opt = self.construct_engine_read(key);
+        let mut iter = self.db.iterator_opt(rocksdb::IteratorMode::End, opt);
```
mfornet, line 100: Nice interface from rocksdb to get the previous value!

mfornet (commented):
The current implementation doesn't allow to iterate through ranges in the storage. One of the suggestions made in the notes were:

> Maintain the current state indexed by an empty key. This can provide range queries for the latest storage only.

Do we plan to support this as well?

engine-standalone-storage/src/engine_state.rs
birchmd: The interface for the `IO` trait is based on the [one from `NEAR` itself](https://nomicon.io/RuntimeSpec/Components/BindingsSpec/TrieAPI.html) where `write` also takes a register to store the old value.

I agree the current implementation is more overhead when the old value is not used. An alternative implementation is the following:

We add another variant to `EngineStorageValue` called `LazyDBRead` which holds the key. And its `AsRef` implementation actually performs the read to return the value. Then for `write_storage` and `remove_storage` we check if the key is in the diff, if so we capture the value as we do now, but if not this `LazyDBRead` is returned insteasd. This means no DB read is done unless the returned value is used. It works because we do not modify the DB during the lifetime of this `IO` instance (writes only change the Diff), so we can technically perform the write before reading the old value from the DB.

What do you think about that? It is more efficient (writes only require allocating memory to store the key, not DB reads unless the value is used), but more complex also. I'm not sure what level of optimization we need for the standalone engine. It probably has an inherent advantage in terms of performance because it does not require wasm interpretation.

engine-standalone-storage/src/engine_state.rs
birchmd: The `IO` instance only holds a reference to the `Diff` so that the instance can implement `Copy` (this allows us to move it between the engine and connector submodules without much overhead; some refactoring could possibly make it so this is not required, but for now it is).

We clone here so that the returned `Diff` can outlive the `IO` instance (the `Diff` will still be used after the `IO` instance is gone because we need to record the `Diff` to storage after we execute the transaction).

I don't think it makes sense to take the `Diff` from the `Storage` object directly (where it is not a reference) because semantically it is the `IO` instance that generates the `Diff`. Also this means under the hood the `Storage` implementation can reuse the same `Diff` allocation between multiple transactions, so I don't think the clone here is even that much overhead (if we somehow did not clone then we would need to still allocate new memory for the next `Diff` when we do a new transaction; here we allocate new memory to capture the `Diff` and reuse the previous memory in the next transaction).

joshuajbouw: As we had discussed earlier, thanks for the detailed response for others that may be curious.

engine-standalone-storage/src/lib.rs
birchmd: I don't agree with the `take` verb. I think of `take` as consuming `self`, where as this only requires a reference `&mut self`. Granted the `get` verb is maybe not quite right either because the returned object can mutate the underlying storage. How about we call it `access_engine_storage_at_position`?

engine-standalone-storage/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "engine-standalone-storage"
+version = "0.1.0"
+edition = "2018"
+authors = ["Aurora <hello@aurora.dev>"]
+description = "Aurora engine standalone library. Provides debugging capabilities."
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/375/commits/fcb58a1ba1ce514b1b8db7ae2bd0fc513b75e398

engine-standalone-storage/src/engine_state.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/375/commits/6b0b3939bae0ac609c5f6dcce6a6463aad6f6088

engine-standalone-storage/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+        self.db.put(storage_key, diff.try_to_bytes().unwrap())?;
+
+        let block_height = self.get_block_height_by_hash(tx_included.block_hash)?;
+        for (key, value) in diff.iter() {
+            let storage_key = construct_engine_key(key, block_height, tx_included.position);
+            self.db.put(storage_key, value.try_to_bytes().unwrap())?;
+        }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
birchmd: Yeah, good call on making sure all the keys are written together. [Rocksdb has a `WriteBatch`](https://github.com/facebook/rocksdb/wiki/Basic-Operations#atomic-updates) for this purpose. Done in https://github.com/aurora-is-near/aurora-engine/pull/375/commits/19118746ee728ff35f0c37595fe5e68c289d2f61

engine-standalone-storage/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    /// Get an object which represents the state of the engine at the given block hash,
+    /// after transactions up to (not including) the given transaction index.
+    /// The `input` is the bytes that would be present in the NEAR runtime (normally
+    /// not needed for standlone engine).
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/375/commits/14a77109128b733838dd660d984311aa555deae5

engine-standalone-storage/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+    TransactionHash = 0x03,
+    Diff = 0x04,
+    Engine = 0x05,
+}
+
+pub struct Storage {
+    db: DB,
+    engine_transaction: RefCell<Diff>,
+    engine_output: Cell<Vec<u8>>,
+}
```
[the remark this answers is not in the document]
birchmd: Yeah, good point. I renamed the engine state one to `EngineStateAccess`. Done in https://github.com/aurora-is-near/aurora-engine/pull/375/commits/ff9bc9ad954eefc4b9426eee56320426b9c0751f

engine-standalone-storage/src/engine_state.rs
birchmd: Yes, rocksdb is great for this purpose!

birchmd:
> The current implementation doesn't allow to iterate through ranges in the storage. One of the suggestions made in the notes were:
> 
> > Maintain the current state indexed by an empty key. This can provide range queries for the latest storage only.
> 
> Do we plan to support this as well?

I don't think I will support this. Having a "hot" state that is updated with every block seems like it will make it hard to get a consistent view of the state. I think if we need to do range queries it should be possible the get the relevant data from diffs alone. It might be a high latency operation, but it seems like use cases requiring range queries will be infrequent and not require instantaneous responses.

birchmd:
> The three storage lifetimes, thats more of a optimisation to prevent cloning of large storage states I take it?

The lifetimes are needed because the engine currently requires `IO + Copy` on its storage access trait bound. None of the rust smart pointers (`Box`, `Rc`, `Arc`, etc) implement `Copy`, so regular references needed to be used. Semantically there are three different references involved, with potentially distinct origins, and therefore three lifetimes. That said, in practice the implementation ends up with all three lifetimes being the same so we could simplify to a single lifetime. I kept the lifetimes independent though because I think it is an implementation detail that they all happen to come out the same and in the future we may wish to change the implementation such that the lifetimes are different.

birchmd:
@joshuajbouw @mfornet Thanks for the reviews! I've addressed all the comments. I've also added the last test I wanted to include for this feature. I think this PR is ready to merge once I have your approvals.

engine-standalone-storage/src/engine_state.rs
mfornet: This is an excellent suggestion, though not strictly necessary for now. I'm ok skipping it.

mfornet (approved):

engine-standalone-storage/src/lib.rs
joshuajbouw: Yeah, I like `acess_engine...` better. Lets go with that. I don't like using `get` as in Rust that heavily implies that it is taking itself as a reference, not a mut reference.

engine-standalone-storage/src/diff.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    pub fn try_from_bytes(bytes: &[u8]) -> Result<Self, std::io::Error> {
+        Self::try_from_slice(bytes)
+    }
+}
+
+impl Diff {
+    /// Compose two Diffs into a single one. If there is a conflict between them
+    /// then the value from the given (`other`) Diff is kept.
+    pub fn compose(&mut self, mut other: Self) {
```
joshuajbouw, line 41: Can we use the `append` name here? I hate to be nit picky, but if internally we are appending, I believe we should keep these verbage so that it is self describing what is going on internally.

engine-standalone-storage/src/lib.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/375/commits/01d802677043ec56ccf13ef9ed34c6f739c2e799

joshuajbouw (approved):
Just note, I believe there are a few optimisations here. But, we need to get this over the edge and its fine for now.

engine-standalone-storage/src/diff.rs
birchmd: Done

birchmd:
> Just note, I believe there are a few optimisations here. But, we need to get this over the edge and its fine for now.
> 
> Specifically with `write_storage`, we require `.to_vec()`, and taking in owned values as a reference even though we are taking ownership of them. Likewise, `write_storage_direct` shouldn't exist.

`write_storage_direct` makes sense in the NEAR-based `IO` implementation as a performance optimization because it means we don't need to copy values from a register just to send it back to the register again (the `StorageValue` type is a register ID). In this implementation I agree it doesn't add any value. But I also think it's harm is pretty minimal, so I doubt it'll cause any performance issue. If it does then the implementation can be adjusted to reduce cloning bytes.

PR #378, opened 2021-11-28
Chore: Cargo make

joshuajbouw:
Make files can be a bit messy and hard to read through the file. Luckily, a rust alternative exists which is quite clean and accomplishes the same goals in a Rust-friendly manner.

What this accomplishes is a replace of `make` with `cargo make`. Why this is important is because we are able to get a cleaner make file that can have even better control with the use of profiles. That would make it quite difficult to run the run deployments on the wrong binaries as it'll correctly read the appropriate ENV file.

On top of this, I had reworked the Github action workflows and I am getting no failures now. Which is good, less time spent with kickstarting it to work again. This also includes a monthly "sweep" to keep our footprint smaller.

In the future, this can be improved by dropping `default_to_workspace = false` and adding a Makefile.toml to every single crate in our workspace.

Do note, for whatever reason there was an efficiency being made on the `profile_erc20_get_balance()` test with this compilation which doesn't make itself immediately obvious. I have tried to figure out why this is but can not come up with a reason.  I request reviewers to please try to notice why this is.

This is low priority, and I have only been working on this while debugging precompiles.

## Follow up PRs

- Update hardhat as per https://github.com/aurora-is-near/aurora-engine/pull/378#discussion_r916772111
- Remove meta-call logic


.github/workflows/builds.yml
```diff
@@ -11,18 +11,23 @@ jobs:
     steps:
       - name: Clone the repository
         uses: actions/checkout@v2
-      - run: make release
-      - run: ls -lH mainnet-release.wasm
+      - name: Install cargo-make
```
strokovok, line 14: There's actually no need to do this. Since our runners are persistent, I will just update the runner image, and `cargo-make` will be forever available.

.github/workflows/scheduled_sweep.yml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+      - name: Clone the repository
+        uses: actions/checkout@v2
+      - name: Install cargo-make
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: --debug cargo-make
+      - name: Run sweep
+        run: cargo make sweep
```
strokovok, line 19: Could you describe what exactly sweep does? Which files are affected?

.github/workflows/tests.yml
```diff
@@ -8,53 +8,32 @@ on:
 
 name: Tests
 jobs:
-  test:
-    name: Test suite (mainnet, testnet, betanet)
+  test-mainnet:
+    strategy:
+      matrix:
+        network: [ mainnet, testnet ]
```
strokovok, line 14: Are you intentionally dropping the betanet?

.github/workflows/tests.yml
```diff
@@ -8,53 +8,32 @@ on:
 
 name: Tests
 jobs:
-  test:
-    name: Test suite (mainnet, testnet, betanet)
+  test-mainnet:
```
strokovok, line 11: It's actually not about mainnet only. I think keeping `test` will be ok

.github/workflows/tests.yml
```diff
@@ -8,53 +8,32 @@ on:
...
+    strategy:
+      matrix:
```
strokovok, line 13: Keep in mind that we've previously discovered that running tests sequentially gives us better execution time because of better cache hit

strokovok (changes requested):

.github/workflows/tests.yml
joshuajbouw: Yes.

.github/workflows/builds.yml
strokovok: I've [installed cargo-make to runners](https://github.com/aurora-is-near/devops-stuff/commit/9e8b7c9a299c1b833ece28f93c3afe78af05d3dc?diff=unified), so you can remove this step from all workflows

.github/workflows/scheduled_sweep.yml
joshuajbouw: This sweep simply cleans builds that are older than 30 days.

.github/workflows/builds.yml
joshuajbouw: Ok, I'll remove it.

.github/workflows/scheduled_sweep.yml
strokovok: Please name files location :) Because there's some chance that this workflow doesn't do what you intended

.github/workflows/builds.yml
strokovok: One place still left)

strokovok (changes requested):

.github/workflows/scheduled_sweep.yml
joshuajbouw: What ever is in the target build folder. If that gets cleaned up each time anyways, then this is unnecessary.

strokovok: Yep, it gets cleaned up anyways)

.github/workflows/tests.yml
```diff
@@ -9,7 +9,7 @@ on:
 name: Tests
 jobs:
   test:
-    name: Test suite (mainnet, testnet)
+    name: Test mainnet
```
strokovok, line 12: "Test suite"

joshuajbouw:
Will be coming to this as its a great improvement to our process.

Makefile.toml
```diff
@@ -0,0 +1,213 @@
...
+category = "Formatting"
+command = "${CARGO}"
+args = [
+    "fmt",
+    "--all",
+]
+
+[tasks.clippy]
+category = "Check"
+command = "${CARGO}"
```
mrLSD, line 69: Same question about constant command.

Makefile.toml
```diff
@@ -0,0 +1,213 @@
...
+    "--features",
+    "contract",
+    "--",
+    "-D",
+    "warnings",
+]
+
+[tasks.check-fmt]
+category = "Check"
+command = "${CARGO}"
```
mrLSD, line 82: Same question about constant command.

Makefile.toml
```diff
@@ -0,0 +1,213 @@
...
+
+[tasks.post-engine-build-env]
+category = "Tools"
+script = '''
+echo "Environment:"
+echo "    IS_RELEASE:           ${IS_RELEASE}"
+echo "    CARGO_FEATURES:       ${ACTUAL_FEATURES}"
+echo "    TARGET_WASM:          ${ACTUAL_WASM}"
+echo "    TARGET_DIR:           ${TARGET_DIR}"
+echo "    Extra build args:     ${@}"
```
mrLSD, line 113: Unclear how it works in that case

Makefile.toml
```diff
@@ -0,0 +1,213 @@
...
+category = "Build"
+script = '''
+cd etc/eth-contracts
+yarn
+yarn build
+'''
+
+[tasks.cargo-build-engine]
+category = "Build"
+command = "${CARGO}"
```
mrLSD, line 142: Same question about constant command.

Makefile.toml
```diff
@@ -0,0 +1,213 @@
...
+category = "Test"
+script = '''
+cd etc/eth-contracts
+yarn
+yarn test
+'''
+
+[tasks.test-workspace]
+category = "Test"
+command = "${CARGO}"
```
mrLSD, line 187: Same question about constant command.

README.md
```diff
@@ -21,22 +21,25 @@ Local   | `aurora.test.near`  | 1313161556 | 2.6.1
...
 
+- Node.js (v14+)
 - Rust nightly (2021-03-25) with the WebAssembly toolchain
-- GNU Make (3.81+)
+- cargo-make
 
 ```sh
 rustup install nightly-2021-03-25
 rustup target add wasm32-unknown-unknown --toolchain nightly-2021-03-25
+cargo install --force cargo-make
```
mrLSD, line 35: Why we still do not use stable Rust?

engine-tests/src/tests/erc20.rs
```diff
@@ -108,7 +108,7 @@ fn profile_erc20_get_balance() {
     assert!(result.is_ok());
 
     // call costs less than 3 Tgas
-    test_utils::assert_gas_bound(profile.all_gas(), 2);
+    test_utils::assert_gas_bound(profile.all_gas(), 3);
```
mrLSD, line 111: Why it was increased?

mrLSD (commented):
Great job!

Why gas cost was increased?

.env/mainnet.env
```diff
@@ -0,0 +1,8 @@
+CARGO_FEATURES = "mainnet"
+CARGO_FEATURES_TEST_BUILD = "mainnet,integration-test,meta-call"
+CARGO_FEATURES_TEST = "mainnet-test"
+NEAR_EVM_ACCOUNT = "aurora"
+TARGET_WASM = "aurora-mainnet.wasm"
+TARGET_WASM_DEBUG = "mainnet-debug.wasm"
+TARGET_WASM_TEST = "aurora-mainnet-test.wasm"
+NEAR_CLI = "near"
```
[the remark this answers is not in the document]
joshuajbouw: It can certainly be different. See https://docs.near.org/docs/tools/kurtosis-localnet for example.

Makefile.toml
```diff
@@ -0,0 +1,213 @@
...
+    { path = ".env/custom.env", profile = "custom" },
+    { path = ".env/local.env", profile = "development" },
+]
+
+[config]
+default_to_workspace = false
+skip_core_tasks = true
+
+[env]
+CARGO = "cargo"
```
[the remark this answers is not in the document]
joshuajbouw: It not always is.

Makefile.toml
joshuajbouw: It is the last flow that is ran so we can be visually certain that the engine is compiling correctly at the end of the build.

README.md
joshuajbouw: rjson needs to be removed, it uses nightly features currently.

engine-tests/src/tests/erc20.rs
joshuajbouw: I had found the reason actually. I missed a rust compiler flag.

joshuajbouw:
I believe that the gas costs got ever so slightly better due to the removal of meta-call. I'm not sure why it affected it, but it appears to have.

etc/eth-contracts/package.json
```diff
@@ -8,18 +8,18 @@
...
     "eslint-config-standard": "^17.0.0",
     "eslint-plugin-import": "^2.26.0",
-    "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-n": "^15.0.0",
+    "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^6.0.0",
     "eth-gas-reporter": "^0.2.25",
     "ethereum-waffle": "^3.4.4",
     "ethers": "^5.6.9",
     "hardhat": "^2.9.9",
```
hskang9, line 21:
```suggestion
    "hardhat": "^2.10.0",
```
For node 18.4.0 devs, hardhat 2.9.9 is not compatible. I saw them adding support [here](https://github.com/NomicFoundation/hardhat/actions/runs/2322266468), and it seems they released a new version. After updating it to 2.10.0,  the error below did not show up:

```
error hardhat@2.9.9: The engine "node" is incompatible with this module. Expected version "^12.0.0 || ^14.0.0 || ^16.0.0". Got "18.4.0"
error Found incompatible module.
```

hskang9 (changes requested):
Without hardhat upgrade, It looks good to me.

Makefile.toml
```diff
@@ -0,0 +1,212 @@
...
+dependencies = [
+    "build-contracts",
+    "build-engine",
+    "make-bin-directory",
+    "copy-engine-build",
+    "post-engine-build-env",
+]
+
+[tasks.build-test]
+env = { "RUSTFLAGS" = "${RUSTC_FLAGS_BUILD}", "CARGO_FEATURES" = "${CARGO_FEATURES_BUILD_TEST}", "WASM_FILE" = "${WASM_FILE_TEST}", "RELEASE" = "--release", "TARGET_DIR" = "release" }
```
hskang9, line 168: I ran `cargo make build-test` command after cloning and installed prerequisite, and it seems `Makefile.toml` cannot get `RUSTC_FLAGS_BUILD` environment variable.

```
[cargo-make] INFO - Execute Command: "cargo" "build" "--target" "wasm32-unknown-unknown" "--release" "--no-default-features" "--features=testnet" "-p" "aurora-engine" "-Z" "avoid-dev-deps"
error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `rustc - --crate-name ___ --print=file-names '${RUSTC_FLAGS_BUILD}' --target wasm32-unknown-unknown --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit status: 1)
  --- stderr
  error: multiple input filenames provided (first two filenames are `-` and `${RUSTC_FLAGS_BUILD}`)
```

After I added `RUSTC_FLAGS_BUILD` as `[env]` variable in the file, the command works. Is there something that I missed in prerequisite?

Makefile.toml
```diff
@@ -0,0 +1,212 @@
...
+category = "Build"
+run_task = "build-engine-flow"
+
+[tasks.build]
+env = { "RUSTFLAGS" = "${RUSTC_FLAGS_BUILD}", "CARGO_FEATURES" = "${CARGO_FEATURES_BUILD}", "RELEASE" = "--release", "TARGET_DIR" = "release" }
```
hskang9, line 173: Same with `build-test` command case.

etc/eth-contracts/package.json
joshuajbouw: This should be done as a follow up PR.

Makefile.toml
joshuajbouw: Good catch, will fix.

etc/eth-contracts/package.json
joshuajbouw: Noted in description.

.github/workflows/scheduled_sweep.yml
```diff
@@ -0,0 +1,14 @@
...
+name: Scheduled sweep
+jobs:
+  clippy:
+    name: Sweep
+    runs-on: [self-hosted, heavy]
+    steps:
+      - name: Clone the repository
+        uses: actions/checkout@v2
+      - name: Run sweep
+        run: cargo make sweep
```
strokovok, line 14: As we discussed before, there's no need to do sweep, because target folder is cleaned up each time anyway

strokovok (changes requested):

strokovok (approved):

PR #382, opened 2021-11-30
Feat(standalone): Structures and logic for keeping storage in sync with the blockchain

birchmd:
Note: This PR is on top of #379 so that PR should be reviewed and merged first. To review the changes for this PR only look at the last commit.

This PR introduces a high level message format for passing new blockchain information into the standalone storage engine. This will allow it to stay in sync with the blockchain. Note this PR does not yet make use of these data structures; in a future PR they will be constructed based on messages received from the NATS message broker (which is still a work in progress from what I understand from @artob and @JonathanLogan ).

To help guide reviewers, the core logic is found in the `consume_message` function.

### DIFF
--- a/engine-standalone-storage/src/sync/mod.rs
+++ b/engine-standalone-storage/src/sync/mod.rs
@@ -1,0 +1,232 @@
+use aurora_engine::{connector, engine, parameters};
+use aurora_engine_sdk::env::{self, Env};
+use aurora_engine_types::TryFrom;
+use borsh::BorshDeserialize;
+
+pub mod types;
+
+use types::{Message, TransactionKind};
+
+const AURORA_ACCOUNT_ID: &str = "aurora";
+
+pub fn consume_message(storage: &mut crate::Storage, message: Message) -> Result<(), error::Error> {
+    match message {
+        Message::Block(block_message) => {
+            let block_hash = block_message.hash;
+            let block_height = block_message.height;
+            let block_metadata = block_message.metadata;
+            storage
+                .set_block_data(block_hash, block_height, block_metadata)
+                .map_err(crate::Error::Rocksdb)?;
+            Ok(())
+        }
+
+        Message::Transaction(transaction_message) => {
+            // Failed transactions have no impact on the state of our database.
+            if !transaction_message.succeeded {
+                return Ok(());
+            }
+
+            let signer_account_id = transaction_message.signer;
+            let predecessor_account_id = transaction_message.caller;
+            let relayer_address = aurora_engine_sdk::types::near_account_to_evm_address(
+                predecessor_account_id.as_bytes(),
+            );
+            let transaction_position = transaction_message.position;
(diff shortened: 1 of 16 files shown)


engine-standalone-storage/src/sync/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+            let predecessor_account_id = transaction_message.caller;
+            let relayer_address = aurora_engine_sdk::types::near_account_to_evm_address(
+                predecessor_account_id.as_bytes(),
+            );
+            let transaction_position = transaction_message.position;
+            let near_tx_hash = transaction_message.near_tx_hash;
+            let block_hash = transaction_message.block_hash;
+            let block_height = storage.get_block_height_by_hash(block_hash)?;
+            let block_metadata = storage.get_block_metadata(block_hash)?;
+            let current_account_id = "aurora".parse().unwrap();
```
joshuajbouw, line 41: This is hard coded, probably should be a constant for now. Would be nice in the future to have configuration settings where this can be changed.

engine-standalone-storage/src/sync/types.rs
```diff
@@ -0,0 +1,53 @@
...
+pub enum TransactionKind {
+    Submit(EthTransactionKind),
+    Call(parameters::CallArgs),
+    /// Input here represents the EVM code used to create the new contract
+    Deploy(Vec<u8>),
+    DeployErc20(parameters::DeployErc20TokenArgs),
+    /// This type of transaction can impact the aurora state because of the bridge
+    FtOnTransfer(parameters::NEP141FtOnTransferArgs),
+    /// Bytes here will be parsed into `aurora_engine::proof::Proof`
+    Deposit(Vec<u8>),
```
joshuajbouw, line 52: Should include all the missing documentation here.

joshuajbouw (approved):
Mostly just notes. We need to get into the habit of documenting everything. However, I understand that we can come back and clean this up later as we do need to deliver.

engine-standalone-storage/src/sync/mod.rs
birchmd: Factored out as a constant in https://github.com/aurora-is-near/aurora-engine/pull/382/commits/f8b1611c75eb26dd65de024b1830d906b07f8f49

I agree having a config would be a little nicer. Not necessary for us right now because both testnet and mainnet use the `aurora` account, so I'll leave this for the time being. When we tie everything together in the in the standalone engine binary we will need some configuration for sure (eg storage directory, NATS URL, etc). At that point we can maybe include this kind of "environment" configuration too.

engine-standalone-storage/src/sync/types.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/382/commits/e2a520768b00a59ab24b50dc82a3895525868c0f

PR #383, opened 2021-12-02
Feat(standalone): EVM tracing via SputnikVM

birchmd:
This PR exposes the tracing functionality provided by SputnikVM to the standalone engine via a convenience function `traced_call`.

The tracing structures and convenience function are also separated into their own crate so that they can be used in the existing `engine-tests` as a dependency. This PR also includes a test showing how the tracing feature can be used, and validates the trace produced contains the expected data.

In a future PR we will need to perform the transformation of the SputnikVM events into the format expected by the geth tracing.


engine-standalone-tracing/src/sputnik.rs
```diff
@@ -0,0 +1,79 @@
+use std::cell::RefCell;
+use std::ptr::NonNull;
+use std::rc::Rc;
+
+// TODO: implement evm* tracing traits for `crate::trace::Logs`, and tie this `traced_call` function to the FFI.
```
joshuajbouw, line 5: Note TODO

joshuajbouw (approved):
Just note the TODO

engine-standalone-tracing/src/sputnik.rs
birchmd: Trait implementations are done in #384 

Hooking in with the main standalone binary will be done in a future PR (pending the decision on how the relayer will interact with standalone)

PR #384, opened 2021-12-03
Feat(standalone): Capture geth-like tracing from SputnikVM events

birchmd:
~This is a follow up PR to #383 that one should be reviewed and merged first. To see only the changes from this PR look only at the last commit.~

The purpose of this PR is to produce a `TransactionTrace` from the events emitted by SputnikVM.

More comprehensive testing is needed to make sure this conversion is always done correctly, but a basic sanity test present in this PR does pass. It would be nice if we had more examples of geth traces to compare against, does anyone know if there is a way we can get them?

In the meantime, I suggest merging this PR as a baseline and we can make iterative corrections to edge cases as more testing is done.

### DIFF
--- a/engine-standalone-tracing/src/sputnik.rs
+++ b/engine-standalone-tracing/src/sputnik.rs
@@ -1,9 +1,13 @@
+use aurora_engine_types::types::EthGas;
+use evm::{Capture, Opcode};
 use std::cell::RefCell;
 use std::ptr::NonNull;
 use std::rc::Rc;
 
-// TODO: implement evm* tracing traits for `crate::trace::Logs`, and tie this `traced_call` function to the FFI.
-#[allow(dead_code)]
+use crate::types::{
+    LogStorageKey, LogStorageValue, Logs, ProgramCounter, TraceLog, TransactionTrace,
+};
+
 /// Capture all events from SputnikVM emitted from within the given closure using the given listener.
 pub fn traced_call<T, R, F>(listener: &mut T, f: F) -> R
 where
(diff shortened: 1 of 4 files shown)


joshuajbouw:
I'll review this one the other two dependent PRs are in.

engine-standalone-tracing/src/types.rs
```diff
@@ -231,19 +252,19 @@ impl TransactionTrace {
     }
 
     /// Returns the return as a str reference.
```
joshuajbouw, line 254: Update the doc comment

engine-standalone-tracing/src/sputnik.rs
```diff
@@ -24,6 +30,165 @@ where
...
+                if !self.current.depth.is_zero() {
+                    // If the depth is not zero then an error must have occurred to
+                    // exit early.
+                    self.failed = true;
+                    self.output = return_value.to_vec();
+                }
+            }
+            Event::TransactCall { .. } => (), // no useful information
+            Event::TransactCreate { .. } => (), // no useful information
+            Event::TransactCreate2 { .. } => (), // no useful information
```
joshuajbouw, line 185: This could be represented simple as `_ => (),` to shorten the code here.

joshuajbouw (approved):
Approved as its all functionally fine. Do note my comments.

engine-standalone-tracing/src/sputnik.rs
birchmd: I like explicitly listing all the cases because then the compiler will complain at you if a new variant is added.

engine-standalone-tracing/src/types.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/384/commits/3bd3c9f1f2b34c5a12759c99f9c9235026c410d5

PR #386, opened 2021-12-06
Balance type refactoring

mrLSD:
The current implementation of the balance is an extremely simplified form:
```
pub type balance = u128;
```
It's just a type alias.

This PR introduces a strict type for handling the essence of balance.

This will allow the logic related to the balance to be made stricter, which will eliminate a whole class of errors both in the source code and in the business logic.


engine-types/src/types/wei.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn from(value: WeiU256) -> Self {
+        Wei(U256::from_big_endian(&value))
+    }
+}
+
+impl From<Balance> for Wei {
+    fn from(value: Balance) -> Self {
+        Wei(U256::from(value.into_u128()))
+    }
+}
```
birchmd, line 98: I'm a little concerned about existence of this implementation. I understand that inside the EVM we have the `Wei` newtype and in the connector (NEP-141 contract) we have the `Balance` newtype, and these are two representations of the same underlying asset.

However, if we are going to allow seamless converting `Balance` into `Wei` then I believe that we cannot use this `Balance` type for anything other than Aurora NEP-141 balances. For example, we should not use it to represent NEAR balances (eg when attaching NEAR to promises). It probably also needs a different name like `BridgedWei` or `NEP141Wei` or something similar so that it is clear this meant to represent that specific assent and no other 128-bit sized quantity.

birchmd (changes requested):
This PR is a bit hard to review with its changes being mixed in with the ones from #374 but here are some initial comments.

engine-types/src/types/balance.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+impl Mul<Balance> for Balance {
+    type Output = Balance;
+
+    fn mul(self, rhs: Balance) -> Self::Output {
+        Balance(self.0 * rhs.0)
+    }
+}
```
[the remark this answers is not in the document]
mrLSD: Still, an explanation is needed here. Because there are basic entities Add, Sub - in your vision, everything is fine with them. However, the single difference between Mul is execution priority, and its presence for a single entity is not mutually contradictory.

engine-types/src/types/wei.rs
mrLSD: We can introduce something like:
`pub const fn new_u128(amount: u128) -> Wei {...}`

engine-types/src/types/balance.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl Div<Balance> for Balance {
+    type Output = Balance;
+
+    fn div(self, rhs: Balance) -> Self::Output {
+        Balance(self.0 / rhs.0)
+    }
+}
```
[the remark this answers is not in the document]
mrLSD: Removed

engine-types/src/types/balance.rs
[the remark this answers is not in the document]
birchmd: If I have $5 and you have $10, how can we combine our money to create $50? We cannot. This is because multiplying two dollar amounts together does not make sense.

If there is a group of 10 friends, each of whom has $5, how can they combine their money together to create $50? They can pool their money to get $5 + $5 + $5 + $5 + $5 + $5 + $5 + $5 + $5 + $5 = 10 * $5 = $50. Note in this example that 10 is a pure number (not a dollar amount), which is why `Mul<u128> for Balance` would make sense as an implementation, it is a short-cut for repeated addition (or in the language of linear algebra, we would say this unit supports [scalar multiplication](https://en.wikipedia.org/wiki/Scalar_multiplication)).

Multiplication (and division) is fundamentally different than addition (and subtraction) when it comes to quantities that have some semantic meaning in the real world. The same is true for units of distance, time, mass, etc. You can add two quantities that have the same units to get a new quantity with those units (3 apples + 2 apples = 5 apples). You cannot multiply two quantities with the same units to get a new quantity with those units (3 apples * 2 apples = ??? what does it mean to multiply apples?).

To learn more about units and algebraic operations on them, check out [dimensional analysis](https://en.wikipedia.org/wiki/Dimensional_analysis).

engine-types/src/types/wei.rs
birchmd: Well, the more I think about this, the more I believe we actually should split `Balance` into multiple different newtype definitions. Right now we are using the same type for NEAR and `NEP141Wei` interchangeably which is not good. I think it would be safer if we had separate types so that we could never accidentally mix them up.

engine-types/src/types/balance.rs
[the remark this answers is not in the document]
mrLSD: But still correct: `5 apples * 2 = 10 apples`
In our case it's implementation for:
`Mul<u128> for Balance`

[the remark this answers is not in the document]
mrLSD: Actually I removed it.

engine-types/src/types/wei.rs
mrLSD: Do you mean:
```
pub struct FungibleTokenOps<I: IO> {
    /// Total ETH supply on Near (nETH as NEP-141 token)
    pub total_eth_supply_on_near: Balance,

    /// Total ETH supply on Aurora (ETH in Aurora EVM)
    pub total_eth_supply_on_aurora: NEP141Wei,

    /// The storage size in bytes for one account.
    pub account_storage_usage: StorageUsage,
    io: I,
}
```

mrLSD: And you meanl: `pub struct NEP141Wei(Wei);`?

engine-types/src/types/balance.rs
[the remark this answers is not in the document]
birchmd: Yes, `Mul<u128> for Balance` would have been fine, as I said in my rant. But I'm also just as happy to see it removed if we never use it anyways.

engine-types/src/types/wei.rs
birchmd: I think the most accurate usage of types would be something like this:

```
pub struct FungibleTokenOps<I: IO> {
    /// Total ETH supply on Near (nETH as NEP-141 token)
    pub total_eth_supply_on_near: NEP141Wei,

    /// Total ETH supply on Aurora (ETH in Aurora EVM)
    pub total_eth_supply_on_aurora: Wei,

    /// The storage size in bytes for one account.
    pub account_storage_usage: StorageUsage,
    io: I,
}
```

Because it uses the NEP-141 type for the quantity on NEAR and the EVM type for the quantity in the engine.

And definition of `NEP141Wei` would be `pub struct NEP141Wei(u128)`.

Additionally there would be another type, say `pub struct Yocto(u128)` (name inspired by the fact 1 NEAR = 10^24 yoctoNEAR, and 1 yoctoNEAR is the smallest unit), and this would be used in places where we expect NEAR tokens to be. For example: attached deposit for promises and storage registration.

mrLSD: One notice here. Currently, for `ft_transfer_call` we have input data:
```
pub struct TransferCallCallArgs {
    pub receiver_id: AccountId,
    pub amount: Balance,
    pub memo: Option<String>,
    pub msg: String,
}
```
Where `amount: Balance`. It's same amount for `eth_on_aurora` & `eth_on_near`.
I think we can't change it to `U256` due to incompatibility with the current format.

For other cases we can improve `Balances` types with your proposal.

engine-types/src/types/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub fee_amount: Wei,
+    pub fee_address: Address,
+    pub contract_address: Address,
+    pub value: Wei,
+    pub input: Vec<u8>,
+}
+
+pub struct StorageBalanceBounds {
+    pub min: Balance,
+    pub max: Option<Balance>,
```
birchmd, line 72: I think these should be `Yocto`

engine-types/src/types/wei.rs
```diff
@@ -0,0 +1,127 @@
+use crate::fmt::Formatter;
+use crate::types::balance::error;
+use crate::types::Balance;
+use crate::{Add, Display, Sub, U256};
+
+/// Wei compatible Borsh-encoded raw value to attach an ETH balance to the transaction
+pub type WeiU256 = [u8; 32];
+
+// the NEP-141 balance for the quantity on NEAR in the EVM (Wei)
```
birchmd, line 9:
```suggestion
// Type representing the NEP-141 balances of the eth-connector (ie Wei amounts that have been bridged to Near)
```

engine-types/src/types/wei.rs
birchmd: Need to `impl From<Balance> for Wei` now that `Balance` definitely does not represent any sort of Wei.

engine/src/connector.rs
```diff
@@ -160,7 +161,7 @@ impl<I: IO + Copy> EthConnectorContract<I> {
             return Err(error::DepositError::CustodianAddressMismatch);
         }
 
-        if event.fee.into_u128() >= event.amount {
+        if Balance::new(event.fee.into_u128()) >= event.amount {
```
birchmd, line 164: Should be `NEP141Wei`

engine/src/connector.rs
```diff
@@ -272,9 +273,9 @@ impl<I: IO + Copy> EthConnectorContract<I> {
             // Mint - calculate new balances
             self.mint_eth_on_near(
                 data.new_owner_id.clone(),
-                data.amount - data.fee.into_u128(),
+                data.amount - Balance::new(data.fee.into_u128()),
```
birchmd, line 276: Should be `NEP141Wei`

engine/src/connector.rs
```diff
@@ -272,9 +273,9 @@ impl<I: IO + Copy> EthConnectorContract<I> {
...
             )?;
-            self.mint_eth_on_near(data.relayer_id, data.fee.into_u128())?;
+            self.mint_eth_on_near(data.relayer_id, Balance::new(data.fee.into_u128()))?;
```
birchmd, line 278: Should be `NEP141Wei`

engine/src/connector.rs
```diff
@@ -595,11 +596,11 @@ impl<I: IO + Copy> EthConnectorContract<I> {
             .map_err(error::FtTransferCallError::MessageParseFailed)?;
 
         // Special case when predecessor_account_id is current_account_id
-        let fee = message_data.fee.into_u128();
+        let fee = Balance::new(message_data.fee.into_u128());
```
birchmd, line 599: We have `Fee` type, does it not make sense to use it here?

engine/src/deposit_event.rs
```diff
@@ -268,21 +268,22 @@ impl DepositedEvent {
...
         let event_message_data: String = event.log.params[1].value.clone().to_string();
 
-        let amount: u128 = event.log.params[2]
+        let amount = event.log.params[2]
             .value
             .clone()
             .into_uint()
             .ok_or(error::ParseError::InvalidAmount)?
             .try_into()
+            .map(Balance::new)
```
birchmd, line 277: Should be `NEP141Wei`

engine/src/engine.rs
```diff
@@ -1144,7 +1145,7 @@ pub fn set_balance<I: IO>(io: &mut I, address: &Address, balance: &Wei) {
 pub fn remove_balance<I: IO + Copy>(io: &mut I, address: &Address) {
     // The `unwrap` is safe here because if the connector
     // is implemented correctly then the "Eth on Aurora" wll never underflow.
-    let balance = get_balance(io, address).try_into_u128().unwrap();
+    let balance = Balance::new(get_balance(io, address).try_into_u128().unwrap());
```
birchmd, line 1148: Should be `NEP141Wei`

engine/src/fungible_token.rs
```diff
@@ -147,7 +148,9 @@ impl<I: IO + Copy> FungibleTokenOps<I> {
         &self,
         address: EthAddress,
     ) -> Result<Balance, crate::prelude::types::error::BalanceOverflowError> {
-        engine::get_balance(&self.io, &Address(address)).try_into_u128()
+        engine::get_balance(&self.io, &Address(address))
+            .try_into_u128()
+            .map(Balance::new)
```
birchmd, line 153: Should be `NEP141Wei`

engine/src/fungible_token.rs
```diff
@@ -284,7 +283,8 @@ impl<I: IO + Copy> FungibleTokenOps<I> {
     }
 
     pub fn ft_balance_of(&self, account_id: &AccountId) -> Balance {
```
birchmd, line 285: Should be `NEP141Wei`

engine/src/fungible_token.rs
```diff
@@ -347,9 +347,11 @@ impl<I: IO + Copy> FungibleTokenOps<I> {
         let unused_amount = match promise_result {
             PromiseResult::NotReady => unreachable!(),
             PromiseResult::Successful(value) => {
-                if let Some(unused_amount) =
+                if let Some(raw_unused_amount) =
                     parse_json(value.as_slice()).and_then(|x| (&x).try_into().ok())
                 {
+                    let unused_amount = Balance::new(raw_unused_amount);
```
birchmd, line 353: Should be `NEP141Wei`

engine/src/fungible_token.rs
```diff
@@ -420,13 +422,13 @@ impl<I: IO + Copy> FungibleTokenOps<I> {
...
         if let Some(balance) = self.get_account_eth_balance(&account_id) {
-            if balance == 0 || force {
+            if balance == ZERO_BALANCE || force {
                 self.accounts_remove(&account_id);
                 self.total_eth_supply_on_near -= balance;
                 let storage_deposit = self.storage_balance_of(&account_id);
                 let action = PromiseAction::Transfer {
                     // The `+ 1` is to cover the 1 yoctoNEAR necessary to call this function in the first place.
-                    amount: storage_deposit.total + 1,
+                    amount: (storage_deposit.total + Balance::new(1)).into_u128(),
```
birchmd, line 431: Should be `Yocto`

engine/src/fungible_token.rs
```diff
@@ -447,7 +449,7 @@ impl<I: IO + Copy> FungibleTokenOps<I> {
 
     pub fn storage_balance_bounds(&self) -> StorageBalanceBounds {
         let required_storage_balance =
-            Balance::from(self.account_storage_usage) * sdk::storage_byte_cost();
+            Balance::from(self.account_storage_usage as u128 * sdk::storage_byte_cost());
```
birchmd, line 452: Should be `Yocto`

engine/src/lib.rs
```diff
@@ -722,7 +722,7 @@ mod contract {
         let mut io = Runtime;
         let args = StorageDepositCallArgs::from(parse_json(&io.read_input().to_vec()).sdk_unwrap());
         let predecessor_account_id = io.predecessor_account_id();
-        let amount = io.attached_deposit();
+        let amount = Balance::new(io.attached_deposit());
```
birchmd, line 725: Should be `Yocto`

engine/src/lib.rs
```diff
@@ -844,12 +845,15 @@ mod contract {
         let args: ([u8; 20], u64, u64) = io.read_input_borsh().sdk_expect("ERR_ARGS");
         let address = Address(args.0);
         let nonce = U256::from(args.1);
-        let balance = U256::from(args.2);
+        let balance = Balance::from(args.2);
```
birchmd, line 848: Should be `NEP141Wei`

engine/src/parameters.rs
```diff
@@ -248,7 +248,7 @@ impl TryFrom<JsonValue> for NEP141FtOnTransferArgs {
         Ok(Self {
             sender_id: AccountId::try_from(value.string("sender_id")?)
                 .map_err(|_| JsonError::InvalidString)?,
-            amount: value.u128("amount")?,
+            amount: Balance::new(value.u128("amount")?),
```
birchmd, line 251: This one is right because we could be receiving any kind of NEP-141 token. You probably knew that, but I just wanted to make at least one positive comment.

engine/src/parameters.rs
```diff
@@ -371,7 +371,7 @@ impl TryFrom<JsonValue> for TransferCallCallArgs {
 
     fn try_from(v: JsonValue) -> Result<Self, Self::Error> {
         let receiver_id = AccountId::try_from(v.string("receiver_id")?)?;
-        let amount = v.u128("amount")?;
+        let amount = Balance::new(v.u128("amount")?);
```
birchmd, line 374: Should be `NEP141Wei`

engine/src/parameters.rs
```diff
@@ -444,7 +444,7 @@ impl TryFrom<JsonValue> for TransferCallArgs {
     fn try_from(v: JsonValue) -> Result<Self, Self::Error> {
         Ok(Self {
             receiver_id: AccountId::try_from(v.string("receiver_id")?)?,
-            amount: v.u128("amount")?,
+            amount: Balance::new(v.u128("amount")?),
```
birchmd, line 447: Should be `NEP141Wei`

engine/src/parameters.rs
```diff
@@ -488,7 +488,7 @@ impl TryFrom<JsonValue> for ResolveTransferCallArgs {
         Ok(Self {
             sender_id: AccountId::try_from(v.string("sender_id")?)?,
             receiver_id: AccountId::try_from(v.string("receiver_id")?)?,
-            amount: v.u128("amount")?,
+            amount: Balance::new(v.u128("amount")?),
```
birchmd, line 491: Should be `NEP141Wei`

birchmd (changes requested):
I tried to thoroughly check the semantics of how each value was being used and point out what the correct type should be based on that.

birchmd:
I can't comment on it directly because it's outside the diff, but I'd like to revisit the types on

```rust
#[derive(Debug, Default, BorshDeserialize, BorshSerialize)]
pub struct FungibleToken {
    /// Total ETH supply on Near (nETH as NEP-141 token)
    pub total_eth_supply_on_near: Balance,
    /// Total ETH supply on Aurora (ETH in Aurora EVM)
    pub total_eth_supply_on_aurora: Balance,
    /// The storage size in bytes for one account.
    pub account_storage_usage: StorageUsage,
}
```

Certainly `Balance` is now the wrong type because obviously these are Wei of some kind. So the simplest thing would be to change them to `NEP141Wei`. But I wonder if it makes sense for both of them to be `Wei` instead. I know the NEP-141 balances are 128-bit, but the total could be larger than that right? There is no reason for us to assume the total of all the accounts will fit in a number the same size as each individual balance.

Let's take a smaller example, say I have the array `[10u8, 20, 250]` and I want to compute the sum. If I try to use `u8` to represent the total then I will get an overflow. But if I use `u16` then it's all fine.

engine-types/src/types/balance.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    pub fn checked_add(self, rhs: Self) -> Option<Self> {
+        self.0.checked_add(rhs.0).map(Self)
+    }
+
+    /// Consumes `Fee` and returns the underlying type.
+    pub fn into_u128(self) -> u128 {
+        self.0
+    }
```
joshuajbouw, line 36: This should be `as_u128` as the inner value supports `Copy`. Taking self as a reference.

joshuajbouw (changes requested):

mrLSD:
> Certainly `Balance` is now the wrong type because obviously these are Wei of some kind. So the simplest thing would be to change them to `NEP141Wei`. But I wonder if it makes sense for both of them to be `Wei` instead. I know the NEP-141 balances are 128-bit, but the total could be larger than that right? There is no reason for us to assume the total of all the accounts will fit in a number the same size as each individual balance.

@birchmd  If we still believe that on the "Eth on Near' side, the NEP-141 standard is being implemented - we should separate types:
```
    /// Total ETH supply on Near (nETH as NEP-141 token)
    pub total_eth_supply_on_near: NEP141Wei,
    /// Total ETH supply on Aurora (ETH in Aurora EVM)
    pub total_eth_supply_on_aurora: Wei,
```

I think it is architecturally wrong to combine these two types into one. 

Or we abandon the idea of supporting NEP-141 standard, but then we will get a huge problem associated with compatibility with other applications that implement this standard and interact with Aurora.

engine/src/engine.rs
mrLSD: No, it is not. Should be `Wei`

mrLSD:
@birchmd wy should we associate 
```
pub struct StorageBalance {
    pub total: NEP141Wei,
    pub available: NEP141Wei,
}
```
to `Yocto` type? Seems it's part of NEP-141 protocol.

birchmd:
@mrLSD you're right, the NEP-141 spec does explicitly say the total is also a 128-bit number. So I agree the correct types are

```rust
    /// Total ETH supply on Near (nETH as NEP-141 token)
    pub total_eth_supply_on_near: NEP141Wei,
    /// Total ETH supply on Aurora (ETH in Aurora EVM)
    pub total_eth_supply_on_aurora: Wei,
```

One potential issue here is the serialization of `total_eth_supply_on_aurora`. If we make it 256-bit then we would also need: (1) a state migration for the next update. Or (2) we would need to write some backwards compatible serialization logic that could take either 128-bit or 256-bit. Or (3) we would need to compromise on the type and leave `total_eth_supply_on_aurora` as `NEP141Wei`. Since the `FungibleToken` struct only uses fixed size data types, we can easily differentiate whether 128 or 256 bits are stored for `total_eth_supply_on_aurora`, so I personally feel that (2) is the best option.

> Why should we associate `StorageBalance` to `Yocto` type?

Because storage deposits use NEAR tokens. See https://nomicon.io/Standards/StorageManagement.html

mrLSD:
> One potential issue here is the serialization of `total_eth_supply_on_aurora`. If we make it 256-bit then we would also need: (1) a state migration for the next update. Or (2) we would need to write some backwards compatible serialization logic that could take either 128-bit or 256-bit. Or (3) we would need to compromise on the type and leave `total_eth_supply_on_aurora` as `NEP141Wei`. Since the `FungibleToken` struct only uses fixed size data types, we can easily differentiate whether 128 or 256 bits are stored for `total_eth_supply_on_aurora`, so I personally feel that (2) is the best option.

@birchmd After our personal discussion of this issue, for the current PR I have implemented as follows:

1. It's for storage. At the same time, we preserve compatibility with the current solution. Those. we don't need migration now. 
```
   pub total_eth_supply_on_near: NEP141Wei,
   pub total_eth_supply_on_aurora: NEP141Wei,
```

2. For `FungibleTokenOps` implemented:
```
   pub total_eth_supply_on_near: NEP141Wei,
   pub total_eth_supply_on_aurora: Wei,
```
So we convert types Wei <--> NEP141Wei. 

3. All infrastructure is ready to change `FungibleToken::total_eth_supply_on_aurora` to `Wei` type. But it should be separated PR, because we should migrate our storage.

4. I'm not sure that for `total_eth_supply_on_aurora` we should support both types after migration. Current implementation for `FungibleTokenOps` is
```
   pub total_eth_supply_on_aurora: Wei
```
So that question is only about storage type and storage migration.

5. Currently, we do not have a mechanism for our contract migration. Seems it should be done first.

engine-types/src/types/wei.rs
```diff
@@ -0,0 +1,231 @@
...
+    fn from(value: Fee) -> Self {
+        Wei(U256::from(value.as_u128()))
+    }
+}
+
+impl From<Balance> for Wei {
+    fn from(value: Balance) -> Self {
+        Wei(U256::from(value.as_u128()))
+    }
+}
```
birchmd, line 181: Please remove this implementation.

engine-types/src/types/wei.rs
```diff
@@ -0,0 +1,231 @@
...
+
+/// Type casting from Wei compatible Borsh-encoded raw value into the Wei value, to attach an ETH balance to the transaction
+impl From<WeiU256> for Wei {
+    fn from(value: WeiU256) -> Self {
+        Wei(U256::from_big_endian(&value))
+    }
+}
+
+impl From<Fee> for Wei {
```
birchmd, line 171: Is this implementation needed? (Related to my question how how `Fee` is used)

engine-types/src/types/wei.rs
```diff
@@ -0,0 +1,231 @@
...
+        let serialized = Wei::new(U256::from(100)).try_to_vec().unwrap();
+        assert_eq!(serialized.len(), 32);
+
+        let balance = Wei::try_from_slice(&serialized).unwrap();
+        assert_eq!(balance, Wei::new(U256::from(100)));
+    }
+
+    #[test]
+    fn test_wei_serialized_len() {
+        let mut serialized = [0u8; 30];
```
birchmd, line 225: A little awkward here that the slice is not 32 bytes, but the test still passes. Makes me wonder if we should add a check the to the deserialize implementation that the slice is 32 bytes.

On the one hand, of course a number represented by fewer than 32 bytes could be cast into a U256, but on the other hand it means we could silently ignore some bugs if we know all valid serializations of this type produce 32 bytes.

engine/src/parameters.rs
```diff
@@ -237,6 +237,8 @@ pub struct BeginBlockArgs {
 #[derive(Debug, Clone)]
 pub struct NEP141FtOnTransferArgs {
     pub sender_id: AccountId,
+    /// Balance can be for Eth on Near and for Eth to Aurora
+    /// We use common `Balance` type without concrete specification.
```
birchmd, line 241: I think the comment here should actually say something like

"`ft_on_transfer` can be called with arbitrary NEP-141 tokens attached, therefore we do not specify a particular type Wei."

If it were true that this only ever received some kind of Wei (either `NEP141Wei` or `Wei`) then I would say that we should make this type `NEP141Wei`.

engine/src/fungible_token.rs
```diff
@@ -504,8 +504,10 @@ impl<I: IO + Copy> FungibleTokenOps<I> {
 
             self.internal_register_account(account_id);
             let refund = amount - min_balance;
-            if refund > 0 {
-                let action = PromiseAction::Transfer { amount: refund };
+            if refund > ZERO_YOCTO {
+                let action = PromiseAction::Transfer {
+                    amount: refund.as_u128(),
+                };
```
birchmd, line 510: Looks like `PromiseAction` enum should be updated to take `Yocto`

birchmd (changes requested):
Nice job! Just a few more comments remaining.

engine-types/src/types/wei.rs
mrLSD: I think you are right. `Fee(NEP141Wei)` looks like a good solution. Will change it.

engine-types/src/types/wei.rs
mrLSD: Yes, we need it.

mrLSD:
>  You haven't answered this question yet. Based on what I can see in the code it looks like the Fee is always in some kind of Wei (sometimes Wei, sometimes NEP141Wei). So I think it should be changed to Fee(NEP141Wei), and then it can easily be converted to either Wei type easily.


I think you are right. Fee(NEP141Wei) looks like a good solution. Will change it.

engine/src/fungible_token.rs
mrLSD: I've thought about it, but I guess it's not idiomatic. Promises operate on the raw types, since they are direct suppliers to `near-sdk`.

engine/src/parameters.rs
mrLSD: Comment changed.

engine/src/fungible_token.rs
birchmd: I think it should take `Yocto` because it is a typed wrapper over the raw functionality. At the interface where `PromiseAction` is actually converted into a NEAR promise is were we would call `as_u128`. Keeping strong typing throughout all our internal logic can prevent bugs.

joshuajbouw (approved):
Looks good :)

PR #387, opened 2021-12-08
Eth Address cleanup

mrLSD:
For more strict types related to Eth address we need modifications:
* modify `Address` type to own type
* Cleanup current `EthAddress` type with new `Address` type
* Implement `Borsh Ser/Deserializer` for `Address` type.


engine-types/src/lib.rs
```diff
@@ -7,6 +7,7 @@ pub mod account_id;
 pub mod parameters;
 pub mod storage;
 pub mod types;
+pub mod types_new;
```
joshuajbouw, line 10: Seems redundant? Why would we need this? Should just live in `types` ?

engine-types/src/parameters.rs
```diff
@@ -53,14 +54,14 @@ pub struct PromiseBatchAction {
...
 
 /// withdraw NEAR eth-connector call args
 #[derive(BorshSerialize, BorshDeserialize)]
 pub struct RefundCallArgs {
-    pub recipient_address: EthAddress,
-    pub erc20_address: Option<EthAddress>,
+    pub recipient_address: Address,
+    pub erc20_address: Option<Address>,
     pub amount: RawU256,
 }
```
joshuajbouw, line 67: Could these break API? Looks like a manual Serialize and Deserialize is needed.

engine/src/connector.rs
```diff
@@ -55,7 +54,7 @@ pub struct EthConnector {
     /// It used in the Deposit flow, to verify log entry form incoming proof.
     pub prover_account: AccountId,
     /// It is Eth address, used in the Deposit and Withdraw logic.
-    pub eth_custodian_address: EthAddress,
+    pub eth_custodian_address: Address,
```
joshuajbouw, line 57: Couldn't this cause issues?

joshuajbouw (changes requested):
I would double check to ensure that anything that is stored anywhere is the exact same format as it currently is. That and ensure that any external call args can certainly be parsed.

engine-types/src/lib.rs
mrLSD: After #386 merge, it will be renamed to `types`. In #386 it's a module, and `types.rs` removed.
To avoid merge conflicts I decided set some temporary name. It'll easily rename to `types` after #386 merge.

engine-types/src/parameters.rs
mrLSD: No breaking changes. For the new `Address` type was implemented custom `Borsh Ser/Deserialize`. And it was tested. `H160`, `U256` do not support `Borsh`. Previously it was `type Adress = H160`.
So we need `RawAddress, RawU256`. But now we can remove at all `RawAddress`.
The code is much cleaner and logically true.

After that it will be PR for Borsh traits wrapped types for `U256`.

engine/src/connector.rs
mrLSD: The idea was to remove `EthAdress`. It was `type EthAddress = [u8; 20]`.
But know we have a strong type for that. Type aliasing is sensitive and can't guaranty types consistency.
`Address` type solves that issue.

engine-types/src/lib.rs
mrLSD: For now, it was merged & deleted.

engine-types/src/types/address.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    #[test]
+    fn test_wrong_address_19() {
+        let serialized_addr = [0u8; 19];
+        let addr = Address::try_from_slice(&serialized_addr);
+        assert_eq!(
+            addr.unwrap_err().to_string(),
+            error::AddressError::IncorrectLength.to_string()
+        );
```
birchmd, line 116: Cleaner to use `matches!` instead of comparing strings.

engine-types/src/types/address.rs
```diff
@@ -0,0 +1,144 @@
...
+        Address::zero()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_address_serializer() {
```
birchmd, line 93: We should also add a test that checks the serialization of `Address` is the same as the serialization of `[u8; 20]`. This will ensure this newtype is backwards compatible with the raw types we were using previously.

engine-precompiles/src/native.rs
```diff
@@ -128,7 +130,7 @@ pub mod events {
 
     fn encode_address(a: Address) -> H256 {
         let mut result = [0u8; 32];
-        result[12..].copy_from_slice(a.as_ref());
+        result[12..].copy_from_slice(a.as_bytes());
```
birchmd, line 133: Could implement `AsRef<[u8]> for Address` to avoid this change.

engine-tests/src/tests/access_lists.rs
```diff
@@ -26,11 +26,13 @@ fn test_access_list_tx_encoding_decoding() {
         data: vec![0],
         access_list: vec![
             AccessTuple {
-                address: test_utils::address_from_hex("0x095e7baea6a6c7c4c2dfeb977efac326af552d87"),
+                address: test_utils::address_from_hex("0x095e7baea6a6c7c4c2dfeb977efac326af552d87")
```
birchmd, line 29: Should the `address` field of `AccessTuple` be the `Address` type instead of `H160`? Or does it make things awkward with passing this into SputnikVM if we do that?

engine-tests/src/tests/erc20_connector.rs
```diff
@@ -109,15 +109,16 @@ impl test_utils::AuroraRunner {
...
 
         result.check_ok();
 
-        Vec::<u8>::try_from_slice(result.value().as_slice())
+        let raw_address: [u8; 20] = Vec::<u8>::try_from_slice(result.value().as_slice())
             .unwrap()
             .try_into()
-            .unwrap()
+            .unwrap();
+        Address::from_slice(&raw_address)
```
birchmd, line 121: Could have a convenience function for taking `[u8; 20]` instead of converting to slice.

```rust
impl Address {
    pub fn from_array(array: [u8; 20]) -> Self { Self(H160(array)) }
}
```

engine-tests/src/tests/eth_connector.rs
```diff
@@ -98,12 +98,10 @@ fn init_contract(
...
 }
 
-fn validate_eth_address(address: &str) -> EthAddress {
+fn validate_eth_address(address: &str) -> Address {
     let data = hex::decode(address).unwrap();
     assert_eq!(data.len(), 20);
-    let mut result = [0u8; 20];
-    result.copy_from_slice(&data);
-    result
+    Address::from_slice(&data)
```
birchmd, line 104: Could use `Address::decode(address).unwrap()` instead and this function would be 1 line instead of 3.

engine-tests/src/tests/eth_connector.rs
```diff
@@ -597,15 +599,13 @@ fn test_deposit_with_0x_prefix() {
     use aurora_engine::deposit_event::TokenMessageData;
     let (master_account, contract) = init(CUSTODIAN_ADDRESS);
 
-    let eth_custodian_address: [u8; 20] = {
-        let mut buf = [0u8; 20];
+    let eth_custodian_address: Address = {
         let bytes = hex::decode(CUSTODIAN_ADDRESS).unwrap();
-        buf.copy_from_slice(&bytes);
-        buf
+        Address::from_slice(&bytes)
```
birchmd, line 604: Could use `Address::decode(CUSTODIAN_ADDRESS).unwrap()`

engine-tests/src/tests/eth_connector.rs
```diff
@@ -597,15 +599,13 @@ fn test_deposit_with_0x_prefix() {
...
     };
-    let recipient_address = [10u8; 20];
+    let recipient_address = Address::from_slice(&[10u8; 20]);
```
birchmd, line 606: Could use `Address::from_array()` function I mentioned earlier.

engine-types/src/types/mod.rs
```diff
@@ -1,21 +1,21 @@
...
+pub use address::*;
 pub use balance::*;
 pub use fee::*;
 pub use gas::*;
 pub use wei::*;
 
-pub type RawAddress = [u8; 20];
 pub type RawU256 = [u8; 32];
-// Big-endian large integer type.
+/// Big-endian large integer type.
```
birchmd, line 16: I think this comment actually applies to `RawU256`, and `Unformatted binary data of fixed length.` applies to `RawH256`.

engine/src/deposit_event.rs
```diff
@@ -258,12 +258,13 @@ impl DepositedEvent {
...
         let event = EthEvent::fetch_log_entry_data(DEPOSITED_EVENT, Self::event_params(), data)
             .map_err(error::ParseError::LogParseFailed)?;
-        let sender = event.log.params[0]
+        let raw_sender = event.log.params[0]
             .value
             .clone()
             .into_address()
             .ok_or(error::ParseError::InvalidSender)?
             .0;
+        let sender = Address::from_slice(&raw_sender);
```
birchmd, line 267: Could probably write as `Address::new(raw_sender)` because I think `into_address()` above would return `H160`.

engine/src/engine.rs
```diff
@@ -540,12 +542,12 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
                 )
             }
             CallArgs::V1(call_args) => {
-                let contract = Address(call_args.contract);
+                let contract = ADDRESS(call_args.contract.raw());
```
birchmd, line 545:
```suggestion
                let contract = call_args.contract;
```

engine/src/engine.rs
```diff
@@ -709,11 +723,11 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
             let mut message = args.msg.as_bytes();
             assert_or_finish!(message.len() >= 40, output_on_fail, self.io);
 
-            let recipient = Address(unwrap_res_or_finish!(
+            let recipient = ADDRESS(H160(unwrap_res_or_finish!(
```
birchmd, line 726: It seems likely to me that this expression could be simplified, but it's hard to tell because of the macro.

engine/src/lib.rs
```diff
@@ -309,7 +307,10 @@ mod contract {
             &io,
         )
         .sdk_unwrap();
-        engine.register_relayer(predecessor_account_id.as_bytes(), Address(relayer_address));
+        engine.register_relayer(
+            predecessor_account_id.as_bytes(),
+            Address::from_slice(&relayer_address),
```
birchmd, line 312: Could instead use `Address::from_array(relayer_address)` function I suggested earlier.

engine/src/lib.rs
```diff
@@ -476,33 +474,33 @@ mod contract {
     pub extern "C" fn get_code() {
         let mut io = Runtime;
         let address = io.read_input_arr20().sdk_unwrap();
-        let code = engine::get_code(&io, &Address(address));
+        let code = engine::get_code(&io, &Address::from_slice(&address));
```
birchmd, line 477: Could instead use `Address::from_array()` function I suggested earlier.

engine/src/lib.rs
```diff
@@ -476,33 +474,33 @@ mod contract {
...
         io.return_output(&code)
     }
 
     #[no_mangle]
     pub extern "C" fn get_balance() {
         let mut io = Runtime;
         let address = io.read_input_arr20().sdk_unwrap();
-        let balance = engine::get_balance(&io, &Address(address));
+        let balance = engine::get_balance(&io, &Address::from_slice(&address));
```
birchmd, line 485: Could instead use `Address::from_array()` function I suggested earlier.

engine/src/lib.rs
```diff
@@ -476,33 +474,33 @@ mod contract {
...
         io.return_output(&balance.to_bytes())
     }
 
     #[no_mangle]
     pub extern "C" fn get_nonce() {
         let mut io = Runtime;
         let address = io.read_input_arr20().sdk_unwrap();
-        let nonce = engine::get_nonce(&io, &Address(address));
+        let nonce = engine::get_nonce(&io, &Address::from_slice(&address));
```
birchmd, line 493: Could instead use `Address::from_array()` function I suggested earlier.

engine/src/lib.rs
```diff
@@ -854,13 +852,13 @@ mod contract {
 
         let mut io = Runtime;
         let args: ([u8; 20], u64, u64) = io.read_input_borsh().sdk_expect("ERR_ARGS");
-        let address = Address(args.0);
+        let address = Address::from_slice(&args.0);
```
birchmd, line 855: Could instead use `Address::from_array()` function I suggested earlier.

engine/src/parameters.rs
```diff
@@ -539,8 +539,8 @@ mod tests {
     #[test]
     fn test_roundtrip_view_call() {
         let x = ViewCallArgs {
-            sender: [1; 20],
-            address: [2; 20],
+            sender: Address::from_slice(&[1; 20]),
+            address: Address::from_slice(&[2; 20]),
```
birchmd, line 543: Could instead use `Address::from_array()` function I suggested earlier.

engine/src/parameters.rs
```diff
@@ -552,12 +552,12 @@ mod tests {
     #[test]
     fn test_call_args_deserialize() {
         let new_input = FunctionCallArgsV2 {
-            contract: [0u8; 20],
+            contract: Address::from_slice(&[0u8; 20]),
```
birchmd, line 555: Could instead use `Address::from_array()` function I suggested earlier.

engine/src/parameters.rs
```diff
@@ -552,12 +552,12 @@ mod tests {
...
             value: WeiU256::default(),
             input: Vec::new(),
         };
         let legacy_input = FunctionCallArgsV1 {
-            contract: [0u8; 20],
+            contract: Address::from_slice(&[0u8; 20]),
```
birchmd, line 560: Could instead use `Address::from_array()` function I suggested earlier.

engine/src/transaction/eip_1559.rs
```diff
@@ -71,7 +69,7 @@ pub struct SignedTransaction1559 {
 }
 
 impl SignedTransaction1559 {
-    pub fn sender(&self) -> Option<Address> {
+    pub fn sender(&self) -> Option<ethabi::Address> {
```
birchmd, line 72: Why return `ethabi::Address` instead of our `Address`?

engine/src/transaction/mod.rs
```diff
@@ -102,7 +102,7 @@ impl From<EthTransactionKind> for NormalizedEthTransaction {
                 access_list: tx.transaction.access_list,
             },
             Eip1559(tx) => Self {
-                address: tx.sender(),
+                address: tx.sender().map(|v| Address::from_slice(v.as_bytes())),
```
birchmd, line 105: `map` not necessary if `sender` returns our `Address` (see previous comment)

birchmd (commented):
Looks pretty good! Most of my comments are minor.

engine-types/src/types/address.rs
mrLSD: macros .unwrap_err() do not support methods call. So we need directly convert `error::AddressError::IncorrectLength.to_string()`. Because `error::AddressError::IncorrectLength` is Enum

engine-types/src/types/address.rs
```diff
@@ -0,0 +1,144 @@
...
+        Ok(Address::new(H160(result)))
+    }
+
+    pub fn as_bytes(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+
+    pub fn from_slice(raw_addr: &[u8]) -> Self {
+        Self::new(H160::from_slice(raw_addr))
+    }
```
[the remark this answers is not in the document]
mrLSD: `try_from_slice` should return `Result`. But in many cases we for that message usage we don't need `Result` type

engine-types/src/types/address.rs
```diff
@@ -0,0 +1,144 @@
...
+use borsh::{BorshDeserialize, BorshSerialize};
+
+/// Base Eth Address type
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Ord, PartialOrd)]
+pub struct Address(H160);
+
+#[allow(non_snake_case, dead_code)]
+pub const fn ADDRESS(addr: H160) -> Address {
+    Address(addr)
+}
```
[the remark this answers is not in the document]
mrLSD: No, we can't. `const fn` is not the same as calling `Address::new`. That method we call in another `const fn`. So we can't do that.

But more importantly, the constant method is about optimization. In many places, we use a 
compilation of code that converts to a constant value. Those. compilation level optimization. 

And we can't use the same `const name` as `type name`. Previously it was just alias to `H160`.

I decided to use more const-specific naming. 

But if we remove `const fn` it will performance regrattion.

engine-tests/src/tests/access_lists.rs
mrLSD: Yes. Because it is related to the Traits of SputnikVM.

engine-types/src/types/address.rs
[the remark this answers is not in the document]
birchmd: I'm not saying we should get rid of `const` optimization. I'm saying calling `ADDRESS` as a function is not idiomatic. Either make `const fn new` or simply make the field public (tuple constructors are always `const`).

engine-types/src/types/address.rs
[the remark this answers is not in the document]
birchmd: In many places we don't need to parse a slice at all (see my comment about having a `[u8; 20]` constructor). We should not have any public functions that panic because the can be used unsafely. Please remove this `from_slice` function and in places where it is used either construct from `[u8; 20]` directly or use `try_from_slice` (whichever is appropriate).

engine-tests/src/tests/erc20_connector.rs
mrLSD: Why not:
```
    pub fn from_array(array: &[u8; 20]) -> Self {
        Self(H160(*array))
    }
```

engine/src/deposit_event.rs
mrLSD: It can't

engine-tests/src/tests/erc20_connector.rs
birchmd: `[u8; 20]` implements `Copy` so passing by reference doesn't really save anything.

engine-types/src/types/address.rs
birchmd: `matches!` uses pattern matching under the hood, so it is exactly for comparing enum variants.

mrLSD: It's different type.
`std::io::Error` and `AddressError`. So we should is error message correct.

engine-types/src/types/address.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub fn raw(&self) -> H160 {
+        self.0
+    }
+
+    /// Encode address to string
+    pub fn encode(&self) -> String {
+        hex::encode(self.0.as_bytes())
+    }
+
+    pub fn decode(address: String) -> Result<Address, error::AddressError> {
```
birchmd, line 25:
```suggestion
    pub fn decode(address: &str) -> Result<Address, error::AddressError> {
```

I don't think ownership of the address string should be needed here, so can pass by reference instead.

engine-types/src/types/address.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        let addr = Address::try_from_slice(&serialized_addr).unwrap();
+        assert_eq!(
+            addr.encode(),
+            "096DE9C2B8A5B8c22cEe3289B101f6960d68E51E".to_lowercase()
+        );
+    }
+
+    #[test]
+    fn test_address_deocde() {
```
birchmd, line 112:
```suggestion
    fn test_address_decode() {
```

engine/src/engine.rs
```diff
@@ -523,15 +527,15 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
         args: CallArgs,
         handler: &mut P,
     ) -> EngineResult<SubmitResult> {
-        let origin = self.origin();
+        let origin = Address::new(self.origin());
         match args {
             CallArgs::V2(call_args) => {
-                let contract = Address(call_args.contract);
+                let contract = Address::new(call_args.contract.raw());
```
birchmd, line 533:
```suggestion
                let contract = call_args.contract;
```

birchmd (approved):
Nice work!
I left a couple more comments, but looks good to me over all.

joshuajbouw (approved):
Looks good now :D

PR #388, opened 2021-12-09
Chore: Remove betanet

joshuajbouw:
As per direction of no longer support betanet, this PR simply removes betanet from our library. This was done as no one actually uses betanet, so we will no longer support it.


birchmd (approved):
LGTM

Issue #381, opened 2021-11-30
nETH ft_transfer_call doesn't attach enough gas to ft_on_transfer


evgenykuzyakov:
The `aurora` contract on NEAR side attaches fixed number of `10 TGas` only:

https://github.com/aurora-is-near/aurora-engine/blob/d020887ca3a9390843e3f6ea034e46e3d11c0f1f/engine/src/fungible_token.rs#L13

The standard implementation is uses the following logic, which allows to provide more than `10 TGas`:

https://github.com/near/near-sdk-rs/blob/7676fda0bc7e93ae809f66878f51144c98bcc974/near-contract-standards/src/fungible_token/core_impl.rs#L180

Even though it's not ideal due to limit on the `msg` size, it allows to provide larger amount of gas for the remote contract.

joshuajbouw:
In our pipeline.

mrLSD:
@evgenykuzyakov thanks for that point.
I see, by the link summary is `30 TGas`.

PR #389, opened 2021-12-09
Increase NEAR Gas for ft_on_transfer

mrLSD:
PR related to the issue: #381

Increased NEAR Gas:
* `deposit` to 80 Tgas
* `ft_on_transfer` 30 TGas

## Solution
The main solution is not to fix the amount of gas for promise call `ft_on_transfer`. For this, a formula is introduced according to which we can use more gas, depending on how much gas was initially released. 

`GAS_FOR_FT_ON_TRANSFER = prepaid_gas - GAS_FOR_FT_TRANSFER_CALL - GAS_FOR_RESOLVE_TRASNFER`

* added new Sdk and Env method - `prepaid_gas`
* introduced a new constant `GAS_FOR_FT_TRANSFER_CALL = 35 TGas`. This constant indicates how much gas is allocated to make a call `ft_transfer_call`.
* To invoke a ft_on_transfer` promise, a formula for gas counting has been introduced: `GAS_FOR_FT_ON_TRANSFER = prepaid_gas - GAS_FOR_FT_TRANSFER_CALL - GAS_FOR_RESOLVE_TRASNFER`

This will allow the use of dedicated gas for this method.

### DIFF
--- a/engine-sdk/src/near_runtime.rs
+++ b/engine-sdk/src/near_runtime.rs
@@ -392,7 +392,7 @@ pub(crate) mod exports {
         fn account_balance(balance_ptr: u64);
         pub(crate) fn attached_deposit(balance_ptr: u64);
         pub(crate) fn prepaid_gas() -> u64;
-        pub(crate) fn used_gas() -> u64;
+        fn used_gas() -> u64;
         // ############
         // # Math API #
         // ############
--- a/engine-sdk/src/near_runtime.rs
+++ b/engine-sdk/src/near_runtime.rs
@@ -238,8 +238,8 @@ impl crate::env::Env for Runtime {
         }
     }
 
-    fn prepaid_gas(&self) -> u64 {
-        unsafe { exports::prepaid_gas() }
+    fn prepaid_gas(&self) -> NearGas {
+        NearGas::new(unsafe { exports::prepaid_gas() })
     }
 }
 
@@ -392,7 +392,7 @@ pub(crate) mod exports {
         fn account_balance(balance_ptr: u64);
         pub(crate) fn attached_deposit(balance_ptr: u64);
         pub(crate) fn prepaid_gas() -> u64;
-        fn used_gas() -> u64;
+        pub(crate) fn used_gas() -> u64;
         // ############
         // # Math API #
         // ############
--- a/engine-sdk/src/near_runtime.rs
+++ b/engine-sdk/src/near_runtime.rs
@@ -237,6 +237,10 @@ impl crate::env::Env for Runtime {
             bytes
         }
     }
+
+    fn prepaid_gas(&self) -> u64 {
+        unsafe { exports::prepaid_gas() }
+    }
 }
 
 impl crate::promise::PromiseHandler for Runtime {
--- a/engine-standalone-storage/src/relayer_db/mod.rs
+++ b/engine-standalone-storage/src/relayer_db/mod.rs
@@ -81,6 +81,7 @@ where
         block_timestamp: env::Timestamp::new(0),
         attached_deposit: 0,
         random_seed: H256::zero(),
+        prepaid_gas: 0,
     };
     let mut handler = crate::promise::Noop;
 
--- a/engine-standalone-storage/src/relayer_db/mod.rs
+++ b/engine-standalone-storage/src/relayer_db/mod.rs
@@ -2,6 +2,7 @@ use aurora_engine::engine;
 use aurora_engine::transaction::EthTransactionKind;
 use aurora_engine_sdk::env::{self, Env};
 use aurora_engine_types::account_id::AccountId;
+use aurora_engine_types::types::NearGas;
 use aurora_engine_types::H256;
 use postgres::fallible_iterator::FallibleIterator;
 
@@ -81,7 +82,7 @@ where
         block_timestamp: env::Timestamp::new(0),
         attached_deposit: 0,
         random_seed: H256::zero(),
-        prepaid_gas: 0,
+        prepaid_gas: NearGas::new(0),
     };
     let mut handler = crate::promise::Noop;
 
--- a/engine-standalone-storage/src/relayer_db/mod.rs
+++ b/engine-standalone-storage/src/relayer_db/mod.rs
@@ -2,7 +2,6 @@ use aurora_engine::engine;
 use aurora_engine::transaction::EthTransactionKind;
 use aurora_engine_sdk::env::{self, Env, DEFAULT_PREPAID_GAS};
 use aurora_engine_types::account_id::AccountId;
-use aurora_engine_types::types::NearGas;
 use aurora_engine_types::H256;
 use postgres::fallible_iterator::FallibleIterator;
 
--- a/engine-standalone-storage/src/relayer_db/mod.rs
+++ b/engine-standalone-storage/src/relayer_db/mod.rs
@@ -1,6 +1,6 @@
 use aurora_engine::engine;
 use aurora_engine::transaction::EthTransactionKind;
-use aurora_engine_sdk::env::{self, Env};
+use aurora_engine_sdk::env::{self, Env, DEFAULT_PREPAID_GAS};
 use aurora_engine_types::account_id::AccountId;
 use aurora_engine_types::types::NearGas;
 use aurora_engine_types::H256;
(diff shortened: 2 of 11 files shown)


engine/src/fungible_token.rs
```diff
@@ -11,8 +11,10 @@ use crate::prelude::{
 };
 use aurora_engine_sdk::io::{StorageIntermediate, IO};
 
+/// Gas for `resolve_transfer`: 5 TGas
 const GAS_FOR_RESOLVE_TRANSFER: NearGas = NearGas::new(5_000_000_000_000);
-const GAS_FOR_FT_ON_TRANSFER: NearGas = NearGas::new(10_000_000_000_000);
+/// Gas for `ft_on_transfer` transfer: 25 TGas + GAS_FOR_RESOLVE_TRANSFER
+const GAS_FOR_FT_ON_TRANSFER: NearGas = NearGas::new(30_000_000_000_000);
```
birchmd, line 17: I don't think this is what we want.

If I understood correctly, it's not that we want to just increase the amount we allocate for this call. Rather, we want to send all the leftover gas to this call, only reserving enough for our own resolve callback afterwards.

birchmd (changes requested):

engine/src/fungible_token.rs
mrLSD: In any case, for the deposit, we have to manually increase the gas.

joshuajbouw (changes requested):
This isn't the solution that was shown.

joshuajbouw:
Please also describe the rational and the solution to the issue outlined in #381.

engine-sdk/src/near_runtime.rs
```diff
@@ -388,7 +392,7 @@ pub(crate) mod exports {
         fn account_balance(balance_ptr: u64);
         pub(crate) fn attached_deposit(balance_ptr: u64);
         pub(crate) fn prepaid_gas() -> u64;
-        fn used_gas() -> u64;
+        pub(crate) fn used_gas() -> u64;
```
birchmd, line 395: This change doe not look necessary

engine-standalone-storage/src/relayer_db/mod.rs
```diff
@@ -81,6 +82,7 @@ where
         block_timestamp: env::Timestamp::new(0),
         attached_deposit: 0,
         random_seed: H256::zero(),
+        prepaid_gas: NearGas::new(0),
```
birchmd, line 85: I think the default value for this should be 300_000_000_000_000. It is the maximum allowed prepaid gas and the most common value for transactions in practice as far as I know. You could probably even define a constant for this value so that you can simply reference it in all the places we use it

```rust
// env.rs

pub const DEFAULT_PREPAID_GAS: NearGas = NearGas::new(300_000_000_000_000);
```

engine-standalone-storage/src/sync/mod.rs
```diff
@@ -46,6 +47,8 @@ pub fn consume_message(storage: &mut crate::Storage, message: Message) -> Result
                 block_timestamp: block_metadata.timestamp,
                 attached_deposit: transaction_message.attached_near,
                 random_seed: block_metadata.random_seed,
+                // TODO: fix it?
```
birchmd, line 50:
```suggestion
```

engine-tests/src/test_utils/standalone/mocks/mod.rs
```diff
@@ -44,6 +46,7 @@ pub fn default_env(block_height: u64) -> aurora_engine_sdk::env::Fixed {
         block_timestamp: aurora_engine_sdk::env::Timestamp::new(0),
         attached_deposit: 0,
         random_seed: H256::zero(),
+        prepaid_gas: NearGas::new(0),
```
birchmd, line 49:
```suggestion
        prepaid_gas: DEFAULT_PREPAID_GAS,
```

See comment above.

engine-tests/src/tests/standalone/sanity.rs
```diff
@@ -30,6 +30,7 @@ fn test_deploy_code() {
         block_timestamp: aurora_engine_sdk::env::Timestamp::new(0),
         attached_deposit: 0,
         random_seed: H256::zero(),
+        prepaid_gas: NearGas::new(0),
```
birchmd, line 33:
```suggestion
        prepaid_gas: DEFAULT_PREPAID_GAS,
```

See first comment

engine/src/connector.rs
```diff
@@ -250,6 +250,7 @@ impl<I: IO + Copy> EthConnectorContract<I> {
         predecessor_account_id: AccountId,
         current_account_id: AccountId,
         data: FinishDepositCallArgs,
+        gas: NearGas,
```
birchmd, line 253:
```suggestion
        prepaid_gas: NearGas,
```

engine/src/connector.rs
```diff
@@ -472,6 +474,7 @@ impl<I: IO + Copy> EthConnectorContract<I> {
         predecessor_account_id: AccountId,
         current_account_id: AccountId,
         args: TransferCallCallArgs,
+        gas: NearGas,
```
birchmd, line 477:
```suggestion
        prepaid_gas: NearGas,
```

engine/src/fungible_token.rs
```diff
@@ -295,6 +298,7 @@ impl<I: IO + Copy> FungibleTokenOps<I> {
         memo: &Option<String>,
         msg: String,
         current_account_id: AccountId,
+        gas: NearGas,
```
birchmd, line 301:
```suggestion
        prepaid_gas: NearGas,
```

engine/src/fungible_token.rs
```diff
@@ -11,8 +11,10 @@ use crate::prelude::{
 };
 use aurora_engine_sdk::io::{StorageIntermediate, IO};
 
+/// Gas for `resolve_transfer`: 5 TGas
 const GAS_FOR_RESOLVE_TRANSFER: NearGas = NearGas::new(5_000_000_000_000);
-const GAS_FOR_FT_ON_TRANSFER: NearGas = NearGas::new(10_000_000_000_000);
+/// Gas for `ft_on_transfer`
+const GAS_FOR_FT_TRANSFER_CALL: NearGas = NearGas::new(25_000_000_000_000);
```
birchmd, line 17: Question: how did you determine this value?

birchmd (commented):
Nice! Looks pretty good to me. I think we should use a more descriptive variable name than simply `gas` and I think we should have a constant for the default prepaid gas value (mostly used in the standalone engine), and after that it should be good to go

engine-sdk/src/near_runtime.rs
mrLSD: It was after tests. Will remove.

engine/src/fungible_token.rs
mrLSD: Experimentally by tests.

mrLSD: But before that it was:  `GAS_FOR_FINISH_DEPOSIT - GAS_FOR_FT_ON_TRANSFER = 50 TGas - 10 TGas = 40 TGas`.
Currently: `GAS_FOR_FT_TRANSFER_CALL + GAS_FOR_FT_ON_TRANSFER = 30 TGas`
So `GAS_FOR_FT_ON_TRANSFER` was increased anyway.

birchmd: Yes, I agree it is clear that `ft_on_transfer` is getting much more gas now, and that is good.

What I am asking about is if our own `ft_transfer_call` will ever use more than 25 Tgas. If it does then creating the promise will fail because we won't have enough prepaid gas to cover it all.

mrLSD: I thinking about that. To guarantee the same work as it does now, this constant must be: `35 TGas`

mrLSD:
> And you're confident it is enough to cover all mainnet use cases? Did you include some extra just to be on the safe side?
 @birchmd 

I changed to `35 TGas`. So we can guaranty the same workflow with gas, as in the current contract.
But from Evgeny Kuzyakiv sample code, that constant is `25 TGas`.

We should decide carefully that value.

birchmd (approved):
LGTM
Nice work

joshuajbouw (approved):
Looks good to me here.

joshuajbouw:
Please squash & rebase so that I can merge. Thanks.

engine-sdk/src/near_runtime.rs
joshuajbouw: Shouldn't clippy be catching this? If not, I know that it can be configured. I'll add an issue about making a clippy file.

PR #390, opened 2021-12-09
Fix(engine): Original_storage bug fix; more tracing tests

birchmd:
The primary goal of this PR was to add more tests for our geth tracing feature. This is accomplished, and also uncovered a bug in the engine.

The implementation of `original_storage` was incorrect in always returning `None`. The original value is accessible because SputnikVM does not change storage values until after the transaction is complete (it stores intermediate writes in memory). This incorrect implementation was causing one of the example transactions I chose to check the trace of to not be replayed correctly (it was assigning the wrong gas cost to one of the `SSTORE` steps because the original value was listed as missing).

This PR fixed this `original_storage` bug along with adding the new tracing tests.

The tests also uncovered a bug in how I was interpreting the `memory_cost` from SputnikVM, and fixes that as well.

Note: by simply changing the `original_storage` implementation without taking into account any backwards compatibility, it means the debug tracing of old transactions may differ slightly from what happened when they were originally executed on NEAR. Personally I think this is acceptable because debug tracing is generally most useful for recent transactions only. Therefore, this bug will have been fixed in the engine by the time anyone wants to debug a transaction and they will see consistent results.

Note: fixing the `original_storage` bug causes the 1inch and uniswap gas costs to increase a little. This may have cause more transactions to hit the NEAR gas limit if they were already close to it. However, I believe it still necessary to make this change because we want our EVM implementation to be as accurate as possible. Hopefully this change will not break anything because the increase is small. And if it does, the problem should be resolved relatively quickly because another wasm cost decrease is coming to mainnet soon.


joshuajbouw (approved):
Looks great. Good find.

I believe that we should be fine for the difference for now between old txs and new txs. Though, it really does suck :(.

The small increase in our gas limit is not good, but it is completely unavoidable as this is a requirement.

PR #391, opened 2021-12-09
Feat(standalone): Add debug tracing of remaining gas values

birchmd:
Geth traces include a `gas` value which counts down from the gas limit. In this PR we capture this information from the SputnikVM events and validate it is the same as the values in example traces from etherscan.


joshuajbouw (approved):
Looks good to me.

mrLSD (approved):
LGTM!

PR #398, opened 2021-12-14
Clippy & rustfmt file preliminary list

mrLSD:
# Clippy & rustfmt file preliminary list

Inroduce `.rustfmt.toml`, `.clippy.toml` files with preliminary configuration list options.

## Documenattion

* https://rust-lang.github.io/rustfmt/
* https://github.com/rust-lang/rustfmt
* https://github.com/rust-lang/rust-clippy
* https://rust-lang.github.io/rust-clippy/master/index.html


joshuajbouw:
Can you please split this into two PRs? That way we know which is changing what.

joshuajbouw:
My method for coming up with the list below is purely ones that will limit panics we may encounter, as well as those that help us attain better code quality.

I'll add more comments but lets start with Nursery for now. Its best to add these in one by one so we know per commit which is changing what.

## Nursery

The nursery ones are those that are best practices coming from the Rust language nursery. imo, most of these should be enabled.

- `empty_line_after_outer_attr`, code quality
- `equatable_if_let`, better practice
- `fallible_impl_from`, limits panics
- `index_refutable_slice`, limits panics
- `option_if_let_else`, code quality
- `redundant_pub_crate`, code quality ease of knowing what types actually are private
- `string_lit_as_bytes`, code quality
- `use_self`, code quality and consistency
- `useless_let_if_seq`, leads to idiomatic code

joshuajbouw:
Here are some more:

## Pendantic

- `cast_lossless`, prevents possible issues with `u8 are u64` etc. Requires `From` instead.
- `cast_possible_truncation`, prevents possible issues with `u64 as u8` etc.
- `cast_possible_wrap`, prevents possible issues such as `u32::MAX as i32;` would in fact yield `-1`.
- `checked_conversions`, prevents error prone conversions. Requires `From` instead and `is_ok`.
- `match_on_vec_items`, catches potential panics.
- `wildcard_imports`, we just shouldn't be doing this at all.

joshuajbouw:
## Restriction
- `as_conversions`, we shouldn't use as conversions. Can lead to dangerous cohersions.
- `expect_used`, disallows potential panics. I'm iffy on this one though. @birchmd ? There are times when things just simply can't fail. And just leads to code bloat.
- `get_unwrap`, I would be very surprised if we did this to be honest.
- `unwrap_used`, we shouldn't be using unwraps, anywhere.

clippy.toml
```diff
@@ -0,0 +1,502 @@
...
+#to_digit_is_some
+#toplevel_ref_arg
+#try_err
+#unnecessary_fold
+#unnecessary_lazy_evaluations
+#unnecessary_mut_passed
+#unsafe_removed_from_name
+#unused_unit
+#unusual_byte_groupings
+#unwrap_or_else_default
```
andrcmdr, line 476: Unwrap or else is lazy by its nature, is most cases not useful to make return in a deterministic way. I'd add just one more ## to mark this setting. =)

andrcmdr (approved):
@joshuajbouw proposed a lot of useful settings.

I like the exhaustive settings list (for clippy and fmt).
During writing a code usage of `cargo check/clippy/build/fmt` frequently and before commit is a good practice. I usually use this via [shell scripts](https://github.com/aurora-is-near/borealis-indexer/blob/main/make.sh) for tuning development ops.
And while gathering local (or via CI) feedback of clippy we can gradually continuously tune these clippy settings to fit our needs. This will improve team habits and code quality.

Tuning of `clippy` and `fmt` is not a fast process (and shouldn't be, there're a lot of useful settings in different situations!) and I'm on a side of gradual approach for settings improvements, to fit into exact project and team ('cause it's hugely depends on a team habits, tasks, agreements on code quality and standards).

mrLSD:
Added clippy options to `Makefile`. Most of them `clippy.toml` not support.

Not supported options:
* equatable_if_let
* index_refutable_slice

@joshuajbouw when we use `as_conversions` almost everywhere we should use:
```
.try_into().unwrap()
```
because we do not expect Result type. Do we need it?

aleksuss:
Is it still actual? @mrLSD

PR #399, opened 2021-12-14
rustfmt file preliminary list

mrLSD:
# rustfmt file preliminary list

Inroduce `.rustfmt.toml` file with preliminary configuration list options.

## Documenattion

* https://rust-lang.github.io/rustfmt/
* https://github.com/rust-lang/rustfmt


joshuajbouw:
I don't think anymore needs to be added to this. Just be sure to run `cargo fmt` and also ensure that the CI is updated. I would expect the CI to actually fail. Not sure why it isn't.

aleksuss:
@mrLSD I think it could be closed because `reorder_imports` and `reorder_modules` are true by default, and `use_field_init_shorthand` is already checked by clippy.

PR #408, opened 2022-01-06
Feat(tests): Mock touching trie nodes

birchmd:
From mainnet data we know the average trie depth for Aurora Engine IO operations is 10. In this PR we wrap the `MockedExt` struct to include this hardcoded trie touching value. This allows us to get more realistic estimates of gas costs from our local tests. Notably, with this change the gas costs of 1inch and uniswap transactions in our regression tests go up, and the fraction of cost due to wasm execution goes down. The new values are more in line with what we expect based on real mainnet transactions.

### DIFF
--- a/engine-tests/src/test_utils/mocked_external.rs
+++ b/engine-tests/src/test_utils/mocked_external.rs
@@ -1,0 +1,179 @@
+use near_vm_logic::mocks::mock_external::MockedExternal;
+
+pub const MAINNET_AVERAGE_TRIE_DEPTH: u64 = 10;
+
+#[derive(Clone)]
+pub(crate) struct MockedExternalWithTrie {
+    pub underlying: MockedExternal,
+    trie_node_count: std::cell::Cell<u64>,
+}
+
+impl MockedExternalWithTrie {
+    pub fn new(ext: MockedExternal) -> Self {
+        Self {
+            underlying: ext,
+            trie_node_count: std::cell::Cell::new(0),
+        }
+    }
+
+    fn increment_trie_node_count(&self, amount: u64) {
+        let cell_value = self.trie_node_count.get();
+        self.trie_node_count.set(cell_value + amount);
+    }
+}
+
+impl near_vm_logic::External for MockedExternalWithTrie {
+    fn storage_set(&mut self, key: &[u8], value: &[u8]) -> Result<(), near_vm_logic::VMLogicError> {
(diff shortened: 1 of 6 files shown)


engine-tests/src/test_utils/mocked_external.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    fn increment_trie_node_count(&self, amount: u64) {
+        let cell_value = self.trie_node_count.get();
+        self.trie_node_count.set(cell_value + amount);
+    }
+}
+
+impl<'b> near_vm_logic::External for MockedExternalWithTrie<'b> {
+    fn storage_set(&mut self, key: &[u8], value: &[u8]) -> Result<(), near_vm_logic::VMLogicError> {
+        self.increment_trie_node_count(10);
```
mfornet, line 24: Should we make named constants out of this values. So it is clear where are they coming from!

joshuajbouw (approved):
This is sweet!

engine-tests/src/test_utils/mocked_external.rs
```diff
@@ -0,0 +1,176 @@
+use near_vm_logic::mocks::mock_external::MockedExternal;
+
+pub(crate) struct MockedExternalWithTrie<'a> {
+    underlying: &'a mut MockedExternal,
```
mfornet, line 4: Is there any particular motivation to use a reference here and build this object from the Mocked External all the time, instead of storing in the AuroraRunner object the MockedExternalWithTrie from the beginning (without requiring a reference to the `underlying` object).

I see the only place we are using `self.ext` without building the `MockedExternalWithTrie` is in the one_shot scenario, and even probably there it makes sense to use MockedExternalWithTrie as well.

mfornet (approved):
Excellent tool. 

As discussed previously with you, I was trying to mimic the exact same behaviour and cost for profiling the txs. While I understand now this is very hard, since it requires full knowledge of the trie structure, this is a good starting point for increasing the accuracy of our tests.

engine-tests/src/test_utils/mocked_external.rs
birchmd: I thought about having `MockedExternalWithTrie` own the underlying `MockedExternal`. Using a reference seemed like a more flexible approach, but you're right it probably just adds more complexity and we will always want to use this with `AuroraRunner` (and the derived one-shot runner). I will change it as you have suggested.

PR #410, opened 2022-01-07
Fix: Only allow owner to call deploy_upgrade

birchmd:
This is not a big deal because we never use the `stage_upgrade` functionality. But it was pointed out to us that `deploy_upgrade` should also be `owner_only` for security reasons if we did ever use that upgrade path. This PR adds this restriction.


sept-en (approved):
Thanks for fixing this!

joshuajbouw (approved):
Looks good

PR #413, opened 2022-01-12
Additional log data for Submit

mrLSD:
Add in logging which will help transaction debugging for gas costs

### DIFF
--- a/engine/src/engine.rs
+++ b/engine/src/engine.rs
@@ -215,6 +215,7 @@ pub enum DeployErc20Error {
     Engine(EngineError),
     Register(RegisterTokenError),
 }
+
 impl AsRef<[u8]> for DeployErc20Error {
     fn as_ref(&self) -> &[u8] {
         match self {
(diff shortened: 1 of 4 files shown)


engine/src/engine.rs
```diff
@@ -860,7 +861,12 @@ pub fn submit<I: IO + Copy, E: Env, P: PromiseHandler>(
         .address
         .ok_or(EngineErrorKind::InvalidSignature)?;
 
-    sdk::log!(crate::prelude::format!("signer_address {:?}", sender).as_str());
+    sdk::log!(crate::prelude::format!(
+        "tx_hash: 0x{}",
+        hex::encode(aurora_engine_sdk::keccak(transaction_bytes))
```
birchmd, line 866: I don't think this is a good change. We can see by the minor bump in the regression tests that the cost is non-zero and I don't think any value is added by logging this. This hashing could be done by looking at the data in the NEAR transaction directly in whatever offline analysis we are doing, no need to do it in the engine.

birchmd (changes requested):

engine/src/engine.rs
mrLSD: @joshuajbouw

joshuajbouw: Yeah, good point. Lets close this for now.

PR #418, opened 2022-01-18
Move engine transactions module to engine-transactions crate

mrLSD:
Move engine `transactions` module to `engine-transactions` crate.

Transactions types should be independent of `engine` crate. It will be easy to reuse only transactions types.

### DIFF
--- a/engine-standalone-storage/Cargo.toml
+++ b/engine-standalone-storage/Cargo.toml
@@ -17,6 +17,7 @@ crate-type = ["lib"]
 aurora-engine = { path = "../engine", default-features = false, features = ["std"] }
 aurora-engine-types = { path = "../engine-types", default-features = false, features = ["std"] }
 aurora-engine-sdk = { path = "../engine-sdk", default-features = false, features = ["std"] }
+aurora-engine-transactions = { path = "../engine-transactions", default-features = false }
 borsh = { version = "0.8.2" }
 evm-core = { git = "https://github.com/aurora-is-near/sputnikvm.git", default-features = false }
 rocksdb = "0.16.0"
--- a/engine-standalone-storage/Cargo.toml
+++ b/engine-standalone-storage/Cargo.toml
@@ -2,7 +2,7 @@
 name = "engine-standalone-storage"
 version = "0.1.0"
 edition = "2018"
-authors = ["Aurora <hello@aurora.dev>"]
+authors = ["Aurora Labs <hello@aurora.dev>"]
 description = "Aurora engine standalone storage library. Provides the storage backend used by the standalone engine."
 homepage = "https://github.com/aurora-is-near/aurora-engine"
 repository = "https://github.com/aurora-is-near/aurora-engine"
--- a/engine-standalone-storage/Cargo.toml
+++ b/engine-standalone-storage/Cargo.toml
@@ -17,7 +17,7 @@ crate-type = ["lib"]
 aurora-engine = { path = "../engine", default-features = false, features = ["std"] }
 aurora-engine-types = { path = "../engine-types", default-features = false, features = ["std"] }
 aurora-engine-sdk = { path = "../engine-sdk", default-features = false, features = ["std"] }
-aurora-engine-transactions = { path = "../engine-transactions", default-features = false }
+aurora-engine-transactions = { path = "../engine-transactions", default-features = false, features = ["std"] }
 borsh = { version = "0.8.2" }
 evm-core = { git = "https://github.com/aurora-is-near/sputnikvm.git", default-features = false }
 rocksdb = "0.16.0"
--- a/engine-tests/Cargo.toml
+++ b/engine-tests/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "aurora-engine-tests"
 version = "1.0.0"
-authors = ["NEAR <hello@near.org>"]
+authors = ["Aurora Labs <hello@aurora.dev>"]
 edition = "2018"
 description = ""
 documentation = ""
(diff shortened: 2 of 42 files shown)


engine-transactions/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "aurora-engine-transactions"
+version = "1.0.0"
+authors = ["NEAR <hello@near.org>"]
```
birchmd, line 4: Do we not have our own author email?

engine-transactions/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+[dependencies]
+aurora-engine-types = { path = "../engine-types", default-features = false }
+aurora-engine-sdk = { path = "../engine-sdk", default-features = false }
+aurora-engine-precompiles = { path = "../engine-precompiles", default-features = false }
+evm = { git = "https://github.com/aurora-is-near/sputnikvm.git", default-features = false }
+rlp = { version = "0.5.0", default-features = false }
+hex = { version = "0.4", default-features = false, features = ["alloc"] }
+
+[features]
+std = ["aurora-engine-types/std", "aurora-engine-precompiles/std"]
```
birchmd, line 24: Probably should also have

```
"evm/std", "rlp/std", "hex/std"
```

assuming each of those crates also have a `std` feature.

engine/Cargo.toml
```diff
@@ -19,6 +19,7 @@ crate-type = ["cdylib", "rlib"]
 aurora-engine-types = { path = "../engine-types", default-features = false }
 aurora-engine-sdk = { path = "../engine-sdk", default-features = false }
 aurora-engine-precompiles = { path = "../engine-precompiles", default-features = false }
+aurora-engine-transactions = { path = "../engine-transactions", default-features = false }
```
birchmd, line 22: Should also include this new crate in the `std` features list below (outside the diff, so I can't comment on it directly)

engine-transactions/src/lib.rs
```diff
@@ -1,5 +1,9 @@
-use crate::prelude::types::{Address, Wei};
-use crate::prelude::{vec, TryFrom, Vec, H160, U256};
+#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(not(feature = "std"), feature(alloc_error_handler))]
+#![cfg_attr(feature = "log", feature(panic_info_message))]
```
birchmd, line 3: Are these features needed?

birchmd (commented):
Looks pretty good to me, just some minor comments.

engine-transactions/Cargo.toml
joshuajbouw: We should update that, its like that throughout our library. Update it here for now, then we can address the rest in a further PR.

engine-transactions/src/lib.rs
joshuajbouw: They shouldn't be, unless we are using something from std.

engine-tests/Cargo.toml
```diff
@@ -17,6 +17,7 @@ aurora-engine = { path = "../engine", default-features = false, features = ["std
 aurora-engine-types = { path = "../engine-types", default-features = false, features = ["std"] }
 aurora-engine-sdk = { path = "../engine-sdk", default-features = false, features = ["std"] }
 aurora-engine-precompiles = { path = "../engine-precompiles", default-features = false, features = ["std"] }
+aurora-engine-transactions = { path = "../engine-transactions", default-features = false }
```
joshuajbouw, line 20: For consistency, we should retain `features = ["std]`

joshuajbouw (changes requested):

engine-transactions/src/lib.rs
mrLSD: We use `std` for tests. Without that, we can't run tests.

engine-standalone-storage/Cargo.toml
```diff
@@ -17,6 +17,7 @@ crate-type = ["lib"]
 aurora-engine = { path = "../engine", default-features = false, features = ["std"] }
 aurora-engine-types = { path = "../engine-types", default-features = false, features = ["std"] }
 aurora-engine-sdk = { path = "../engine-sdk", default-features = false, features = ["std"] }
+aurora-engine-transactions = { path = "../engine-transactions", default-features = false }
```
birchmd, line 20:
```suggestion
aurora-engine-transactions = { path = "../engine-transactions", default-features = false, features = ["std"] }
```

engine-transactions/src/lib.rs
```diff
@@ -1,5 +1,8 @@
-use crate::prelude::types::{Address, Wei};
-use crate::prelude::{vec, TryFrom, Vec, H160, U256};
+#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(not(feature = "std"), feature(alloc_error_handler))]
```
birchmd, line 2: Should be able to delete this too I think

mrLSD: I can't tests without that:
```
error[E0152]: found duplicate lang item `panic_impl`
  --> engine/src/lib.rs:36:1
   |
36 | pub unsafe fn on_panic(info: &::core::panic::PanicInfo) -> ! {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: the lang item is first defined in crate `std` (which `aurora_engine_transactions` depends on)
   = note: first definition in `std` loaded from /Users/anymacstore/.rustup/toolchains/nightly-2021-08-01-x86_64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/lib/libstd-ccdba070b604d7bc.rlib
   = note: second definition in the local crate (`aurora_engine`)

error[E0152]: found duplicate lang item `oom`
  --> engine/src/lib.rs:59:1
   |
59 | pub unsafe fn on_alloc_error(_: core::alloc::Layout) -> ! {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: the lang item is first defined in crate `std` (which `aurora_engine_transactions` depends on)
   = note: first definition in `std` loaded from /Users/anymacstore/.rustup/toolchains/nightly-2021-08-01-x86_64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/lib/libstd-ccdba070b604d7bc.rlib
   = note: second definition in the local crate (`aurora_engine`)

For more information about this error, try `rustc --explain E0152`.
error: could not compile `aurora-engine` due to 2 previous errors
```

joshuajbouw (approved):

PR #420, opened 2022-01-20
Avoid using wasmer directly

matklad:
On the nearcore's side, contract runtime team frequently makes
substantial changes to wasmer and wants to change those changes with
aurora. If aurora has a direct dependency to wasmer, we need to apply
more Cargo.toml patches to test this out.

Moreover, wasmer is not needed for this particular test -- we don't need
to compile WASM to machine code (which is what wasmer is doing), we just
need to run some analysis on wasmer. That's a much simpler thing to do
conceptually, and something `walrus` crate is typically used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -213,7 +213,7 @@ dependencies = [
  "serde_json",
  "sha3 0.9.1",
  "tempfile",
- "wasmer-near",
+ "walrus",
 ]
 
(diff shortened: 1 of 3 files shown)


birchmd (approved):
Thanks @matklad !

joshuajbouw (approved):
Looks great! Thanks man!

PR #421, opened 2022-01-20
use default git attributes for Cargo.lock

matklad:
No super-strong feelings here, I can work-around on my side! But yeah,
the current setup feels a bit odd to me, as it's quite unusual. 

git&github generally handle Cargo.lock reasonably nowdays by default.
Our current explicit setting prevent changes to Cargo.lock being visible
in some tools
(https://marketplace.visualstudio.com/items?itemName=kahole.magit in my
case),  which makes working on upgrading dependencies harder.

Changes to Cargo.lock are the most important ones, as one-line
Cargo.lock addition might bring in thousands of lines of new code. They
generally should be audited with above-average scrunity, not ignored.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-Cargo.lock linguist-generated=true -diff


birchmd (approved):
Makes sense to me.
I think this attribute was ported over from NEAR when we first made our own repo.
I know NEAR has also since removed it, but we hadn't gotten around to doing it yet ourselves.

joshuajbouw (approved):
Wasn't aware we were doing this. Thanks.

PR #422, opened 2022-01-20
switch to rebranded blake2

matklad:
This pulls
https://github.com/near/near-blake2/commit/cebb7df79608a7058017940830d37c37d55d21fe
which just renames blake2. It's parent, 736ff60, is what we are using
currently. There are some extra commits on top of that in master, but I
decided not to pull them in just yet, to minimize change.

The primary motivation here is making it easy to upgrade other things:

makes that tricky, as there's no longer `blake2` package at that URL!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -122,7 +122,6 @@ dependencies = [
  "aurora-engine-sdk",
  "aurora-engine-types",
  "base64 0.13.0",
- "blake2 0.9.1 (git+https://github.com/near/near-blake2.git)",
  "borsh 0.8.2",
  "byte-slice-cast",
  "ethabi",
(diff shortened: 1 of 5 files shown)


birchmd (approved):
LGTM!

joshuajbouw (approved):
Thanks, been meaning to get around to this eventually. Will fix out libs that use the aurora-engine without touching Cargo.lock manually hah!

matklad:
Is there anything I should do to make this mergable?

birchmd:
I guess if you don't see the option to merge it could be due to the repo permissions. I'll merge all your approved PRs now.

matklad:
Yeah, I don't see the option (and I probably shouldn't: I'd rather be a contributor than a maintainer :) ), so I assumed it was waiting on some extra review. Should've pinged earlier, sorry about that!

PR #424, opened 2022-01-24
Fix(engine): Return revert error message during contract deploy

birchmd:
@0x3bfc pointed out that presently the Engine always returns the address the contract would be deployed at, regardless of if the deploy execution was successful or not. In this PR we ensure to pass the error message on in the case of failure.

### DIFF
--- a/engine-precompiles/Cargo.toml
+++ b/engine-precompiles/Cargo.toml
@@ -19,8 +19,8 @@ base64 = { version = "0.13.0", default-features = false, features = ["alloc"] }
 blake2 = { git = "https://github.com/near/near-blake2.git", version = "0.9.1", default-features = false }
 borsh = { version = "0.8.2", default-features = false }
 bn = { package = "aurora-bn", git = "https://github.com/aurora-is-near/aurora-bn.git", default-features = false }
-evm = { git = "https://github.com/aurora-is-near/sputnikvm.git", default-features = false }
-evm-core = { git = "https://github.com/aurora-is-near/sputnikvm.git", default-features = false }
+evm = { git = "https://github.com/aurora-is-near/sputnikvm.git", rev = "177e253ad947b7186bfa6920843075556b60ff54", default-features = false }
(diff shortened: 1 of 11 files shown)


engine-precompiles/Cargo.toml
```diff
@@ -19,8 +19,8 @@ base64 = { version = "0.13.0", default-features = false, features = ["alloc"] }
 blake2 = { git = "https://github.com/near/near-blake2.git", version = "0.9.1", default-features = false }
 borsh = { version = "0.8.2", default-features = false }
 bn = { package = "aurora-bn", git = "https://github.com/aurora-is-near/aurora-bn.git", default-features = false }
-evm = { git = "https://github.com/aurora-is-near/sputnikvm.git", default-features = false }
-evm-core = { git = "https://github.com/aurora-is-near/sputnikvm.git", default-features = false }
+evm = { git = "https://github.com/aurora-is-near/sputnikvm.git", rev = "177e253ad947b7186bfa6920843075556b60ff54", default-features = false }
```
birchmd, line 22: Switching to use our EVM fork for now, we can return to the upstream master branch once https://github.com/rust-blockchain/evm/pull/91 is merged.

engine-precompiles/src/lib.rs
```diff
@@ -47,9 +47,9 @@ impl PrecompileOutput {
     }
 }
 
-impl From<PrecompileOutput> for evm::executor::PrecompileOutput {
+impl From<PrecompileOutput> for executor::stack::PrecompileOutput {
```
birchmd, line 50: These changes are unrelated to the core goal of this PR to properly handle reverts during contract deploy, but were needed after getting up to date with the latest upstream changes of SputnikVM (and getting up to date there was needed for this fix because I needed to make a small change to SputnikVM for this to work properly).

joshuajbouw (approved):

0x3bfc (approved):
LGTM .. Thanks @birchmd 👍

PR #427, opened 2022-01-27
chore: upgrade to the latest nearcore version

matklad:
builds on top of #425

### DIFF
--- a/engine-tests/Cargo.toml
+++ b/engine-tests/Cargo.toml
@@ -37,11 +37,11 @@ serde_json = "1"
 hex = { version = "0.4.3", default-features = false }
 near-sdk = { git = "https://github.com/aurora-is-near/near-sdk-rs.git", rev = "5e58722bd61d9d24ae6293326146c751f0a814fb" }
 near-sdk-sim = { git = "https://github.com/aurora-is-near/near-sdk-rs.git", rev = "5e58722bd61d9d24ae6293326146c751f0a814fb" }
-near-crypto = { git = "https://github.com/near/nearcore.git", rev = "0c9ad79a18e431f843e6123cf4559f9c2c5dd228" }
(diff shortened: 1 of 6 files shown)


engine-tests/Cargo.toml
```diff
@@ -37,11 +37,11 @@ serde_json = "1"
 hex = { version = "0.4.3", default-features = false }
 near-sdk = { git = "https://github.com/aurora-is-near/near-sdk-rs.git", rev = "5e58722bd61d9d24ae6293326146c751f0a814fb" }
 near-sdk-sim = { git = "https://github.com/aurora-is-near/near-sdk-rs.git", rev = "5e58722bd61d9d24ae6293326146c751f0a814fb" }
```
matklad, line 39: Not that this intentionally doesn't update the sdk. It already is the case that nearcore used by sdk is different from the nearcore used by engine-tests (that is, there are already two versions of `near-vm-runner` in Cargo.lock today).

engine-tests/src/tests/erc20_connector.rs
```diff
@@ -522,11 +522,11 @@ mod sim_tests {
                 &aurora.contract,
                 &aurora,
             ),
-            (INITIAL_ETH_BALANCE - ETH_EXIT_AMOUNT).into()
+            (INITIAL_ETH_BALANCE - ETH_EXIT_AMOUNT) as u128
         );
         assert_eq!(
             nep_141_balance_of(exit_account_id.as_str(), &aurora.contract, &aurora),
-            ETH_EXIT_AMOUNT.into()
+            ETH_EXIT_AMOUNT as u128
```
matklad, line 529: Note: this is needed because some crate added a new PartialEq impl, and the `.into` call became ambiguous.

matklad:
rebased!

joshuajbouw (approved):

PR #429, opened 2022-01-27
Benchmark contract performing only pure arithmetic and memory operations

birchmd:
Reproduces in Solidity the [contract used by the nearcore contract runtime](https://github.com/near/nearcore/blob/83fc0f7d6b212bacc49f058e7400743de3e59783/runtime/runtime-params-estimator/test-contract/src/lib.rs#L1031-L1043) team to benchmark wasm. This fulfils a request from @matklad to have such a contract.

### DIFF
--- a/engine-tests/src/tests/sanity.rs
+++ b/engine-tests/src/tests/sanity.rs
@@ -136,6 +136,47 @@ fn test_log_address() {
 }
 
 #[test]
+fn test_solidity_pure_bench() {
+    let (mut runner, mut signer, _) = initialize_transfer();
+    runner.wasm_config.limit_config.max_gas_burnt = u64::MAX;
+
+    let constructor = test_utils::solidity::ContractConstructor::force_compile(
+        "src/tests/res",
+        "target/solidity_build",
+        "bench.sol",
+        "Bencher",
+    );
+
+    let nonce = signer.use_nonce();
+    let contract = runner.deploy_contract(
+        &signer.secret_key,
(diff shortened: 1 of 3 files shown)


engine-tests/src/tests/sanity.rs
```diff
@@ -135,6 +135,47 @@ fn test_log_address() {
...
+
+    let nonce = signer.use_nonce();
+    let contract = runner.deploy_contract(
+        &signer.secret_key,
+        |c| c.deploy_without_constructor(nonce.into()),
+        constructor,
+    );
+
+    // Number of iterations to do
+    let loop_limit = 100_000;
```
birchmd, line 158: I know you said you wanted 10^6 iterations @matklad , but you can already see how ridiculous the costs are at just 10^5 😅

matklad (approved):
<3

matklad:
Yeah, turned out to be _extremely_ useful! See discussion here: https://near.zulipchat.com/#narrow/stream/295306-nearinc.2Fcontract-runtime/topic/Aurora.20profiling/near/269722435

I suggest merging this -- it feels like it can be very useful to keep the raw perf of sputnik in check. Though, I'd probably reduce the number of iterations to just 10_000 -- an 8s test is probably overly long.

birchmd:
Thanks @matklad ! I have switched to 10k iterations. This PR will be merged today

joshuajbouw (approved):
Looks good. Thanks.

PR #432, opened 2022-02-04
Fix(engine): Require chain_id

birchmd:
I don't think anyone has ever submitted any transaction to Aurora without a `chain_id`, so this should not break anything, and it should close potential shenanigans  since we do not expect any legitimate user to not provide a `chain_id`.


sept-en (approved):
LGTM!

joshuajbouw (approved):

joshuajbouw:
Will prep new release, this should make it in soon.

mfornet (approved):

PR #435, opened 2022-02-07
chore: make sure aurora is using the latest cost config

matklad:
The test needs adjustment because non-wasm cost falls from

17_548_272_829_581

to

14_461_725_193_038

Ie, as the total cost is lower, wasm fraction is now bigger.

Haven't looked into which specific cost change caused decrease.


birchmd (approved):
LGTM

joshuajbouw (approved):

PR #436, opened 2022-02-07
Implement native erc20 aurora tokens connector

karim-en:
This pull request is about implementing bridging native aurora erc20 tokens to near's nep141.
- Implemented `EvmErc20Locker.sol` contract to lock erc20 tokens which calls `ExitToNear` Precompile
- Handle withdraw call from tokens factory contract to unlock locked erc20 tokens.
Note: the tokens factory for native aurora tokens is part of [rainbow-token-connector](https://github.com/aurora-is-near/rainbow-token-connector/pull/94) repo.

Sequence description: 

- Init:
1. Call `Aurora::deploy_erc20_locker()` to deploy the locker.
2. Deploy the [token factory for native aurora tokens](https://github.com/aurora-is-near/rainbow-token-connector/pull/94).
3. Call `Aurora::set_native_erc20_factory(factory: AccountId)`

- Deploy nep141 token
1. Call `BridgeTokenFactory::deploy_bridge_token(address: String)` only once for new token.
2. Call `BridgeTokenFactory::update_metadata(token: String)`  to set the metadata.

- Aurora -> NEAR:
Call `EvmErc20Locker::lockToken(address ethToken, uint256 amount, bytes memory recipient)`.

- NEAR -> Aurora:
Call `BridgeToken::withdraw(amount: U128, recipient: String)`


sept-en:
@karim-en even though it's a draft PR, please note that `EvmErc20Locker.sol` should be located in the separate repo (e.g. `rainbow-token-connector`)

karim-en:
@sept-en ok, I can move it and remove the `deploy_erc20_locker`, but let me explain my point of view.
I preferred to keep the deploy the locker from the Aurora contract to avoid extra complexity which means extra vanurability/issue.
Let's compare those options:
- Deploy from Aurora: we need just to connect to the near rpc and call deploy locker.
- Deploy outside Aurora:
1. Connect to Aurora RPC and get eths to cover the gas usage.
2. Convert near Aurora account id to EVM address and deploy the locker using this address.
3. Set the address of the erc20_locker.

In addition to this, it will be easier to cover the locker functionality using the Aurora engine unit tests.
And actually, in my subjective opinion, this contract has a hard link to the EVM due to the precompile call and aurora address, which made sense to keep as a part of this repo.

sept-en:
@karim-en this is one-time usage of the function by design and Aurora key is not required in the deployment process at all (only needed to link the factory account after the deployment).
Also, it was the same process for other connectors too.

> In addition to this, it will be easier to cover the locker functionality using the Aurora engine unit tests.
You can still implement unit tests even without having these locker-specific functions in Aurora.

> And actually, in my subjective opinion, this contract has a hard link to the EVM due to the precompile call and aurora address, which made sense to keep as a part of this repo.
All other connectors, especially ether one, are quite connected to Aurora too, but we don't list them here.

Also, my main point is not about the place of keeping the smart contract, but more about removing locker-specific functions from Aurora interface. There's no need to expand the interface if it's not required.

Let's keep the contract here for now, as the PR is still in the draft mode, and discuss further with other guys too when it becomes ready-to-review.

engine-precompiles/src/native.rs
```diff
@@ -220,6 +223,56 @@ fn get_nep141_from_erc20(erc20_token: &[u8]) -> AccountId {
     .unwrap()
 }
 
+#[cfg(feature = "contract")]
+fn get_native_erc20_state() -> NativeErc20EngineState {
```
birchmd, line 227: We're going to need to work on this. In order for the standalone engine to be able to execute all Aurora transactions exactly the same as on NEAR we cannot have some functions that only with the the NEAR Runtime.

The way we have been doing this for other precompiles is to pass any state needed to the precompile at the time the Engine object is created. That approach may not quite work here, we may need to pass the full `IO` object instead, but I'll need to review more carefully to see.

engine-tests/src/tests/native_erc20_connector.rs
```diff
@@ -0,0 +1,441 @@
+// Simulation tests for exit to NEAR precompile.
+// Note: `AuroraRunner` is not suitable for these tests because
+// it does not execute promises; but `near-sdk-sim` does.
+mod sim_tests {
```
birchmd, line 4: Would be cool to have at least one test that runs through the standalone engine as well.

birchmd (changes requested):
I haven't looked at the logic of this feature itself very closely yet, but I know we will need to make this work with the standalone engine as well so I'm only going to flag that issue for now.

engine/src/engine.rs
```diff
@@ -641,6 +667,76 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
...
+        let output: Vec<u8> = match result {
+            Ok(result) => match result {
+                TransactionStatus::Succeed(ret) => ret,
+                _other => Vec::new(),
+            },
+            Err(e) => return Err(e),
+        };
+
+        Ok(ethabi::decode(output_types, output.as_slice())
+            .unwrap()
```
mrLSD, line 694: Please do not use `unwrap`. Wrap as error result.

engine/src/engine.rs
```diff
@@ -641,6 +667,76 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
...
+
+    pub fn get_erc20_metadata(&self, erc20_address: Address) -> NativeErc20Metadata {
+        let name: String = self
+            .view_with_selector(
+                self.origin,
+                erc20_address,
+                &[6, 253, 222, 3],
+                &[ethabi::ParamType::String],
+            )
+            .unwrap()
```
mrLSD, line 707: Do not use `unwrap`. Return `Result` type.

engine/src/engine.rs
```diff
@@ -641,6 +667,76 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
...
+                self.origin,
+                erc20_address,
+                &[49, 60, 229, 103],
+                &[ethabi::ParamType::Uint(8)],
+            )
+            .unwrap()
+            .into_uint()
+            .unwrap()
+            .try_into()
+            .unwrap();
```
mrLSD, line 731: Same issue. Do not use `unwrap`.

engine/src/engine.rs
```diff
@@ -837,6 +933,52 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
...
+
+    pub fn unlock_erc20_token<P: PromiseHandler>(
+        &mut self,
+        predecessor_account_id: &AccountId,
+        args: &UnlockErc20TokenArgs,
+        current_account_id: &AccountId,
+        handler: &mut P,
+    ) {
+        let native_erc20_state = get_native_erc20_state(&self.io).unwrap_or_default();
+        assert_eq!(
```
mrLSD, line 945: I think `assert`, and `panics`  here is not good. Pls return `Result` type.

engine/src/engine.rs
```diff
@@ -975,6 +1117,24 @@ pub fn compute_block_hash(chain_id: [u8; 32], block_height: u64, account_id: &[u
...
+        None => Err(EngineStateError::NotFound),
+        Some(bytes) => NativeErc20EngineState::try_from_slice(&bytes.to_vec())
+            .map_err(|_| EngineStateError::DeserializationFailed),
+    }
+}
+
+pub fn set_native_erc20_state<I: IO>(io: &mut I, state: NativeErc20EngineState) {
+    io.write_storage(
+        &bytes_to_key(KeyPrefix::Config, NATIVE_ERC20_BRIDGE_STATE_KEY),
+        &state.try_to_vec().expect("ERR_SER"),
```
mrLSD, line 1134: Please, do not `panic`. Return `Result` type and more explicit error message.

mrLSD (changes requested):

engine/src/engine.rs
karim-en: @mrLSD  Should I also fix all panics in this file like this https://github.com/aurora-is-near/aurora-engine/blob/develop/engine/src/engine.rs#L998? 
Because it is weird that panic is not recommended in the engine but part of the code is already used panic.

mrLSD: I think, sure.

birchmd: I think we should consider what could cause the error to happen and whether or not it is "fatal". In the case of "fatal" errors it is ok to panic because it is better to crash quickly instead of continue running in an inconsistent state. On the other hand, normal errors should be passed via the return value because they could potentially be handled upstream by the caller.

In this particular example, I think the panic is ok because I think the only way for `try_to_vec` to fail is if it could not allocate enough memory to serialize the object. If the Engine is running out of memory it is probably best to simply shut it down.

engine-precompiles/src/native.rs
```diff
@@ -220,6 +223,56 @@ fn get_nep141_from_erc20(erc20_token: &[u8]) -> AccountId {
...
+        ))
+        .expect(ERR_NATIVE_ERC20_STATE_NOT_FOUND);
+
+    use borsh::BorshDeserialize;
+    NativeErc20EngineState::try_from_slice(&bytes.to_vec())
+        .expect(ERR_NATIVE_ERC20_STATE_DESERIALIZATION_FAILED)
+}
+
+#[cfg(all(feature = "contract"))]
+fn create_refund_promise(
```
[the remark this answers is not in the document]
karim-en: @birchmd  Yes, it doesn't relay directly, but it is only used by `fn run` method which is behind the `contract` feature, removing this will produce an "unused" warning, also in this case we need to remove the already used feature here https://github.com/aurora-is-near/aurora-engine/blob/c72399765735f7407397ec21b60e5fb609c44dc4/engine-precompiles/src/native.rs#L12

engine-precompiles/src/native.rs
karim-en: @birchmd hmm, could you please give more details about this. As I see we already have access to the `IO` here https://github.com/aurora-is-near/aurora-engine/blob/bridge-native-aurora-tokens/engine-precompiles/src/native.rs#L215, so I used this code as an example. I don't fully get your point.

birchmd: Hm, I see. I thought we had eliminated all the `contract` feature gates when we did the standalone engine work previously. I guess I must have missed this one. If all this native precompile stuff is already gated behind the `contract` feature then following the same pattern here is ok for now just so that we don't block this feature. But please add a TODO and create a corresponding issue assigned to me so that I can sort this out after your PR is merged.

engine-precompiles/src/native.rs
```diff
@@ -22,6 +23,9 @@ use evm::backend::Log;
 use evm::{Context, ExitError};
 
 const ERR_TARGET_TOKEN_NOT_FOUND: &str = "Target token not found";
+const ERR_NATIVE_ERC20_STATE_NOT_FOUND: &str = "The native ERC20 state not found";
```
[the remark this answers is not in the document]
joshuajbouw: We haven't gone through an error overhaul yet. Should be as you stated.

joshuajbouw:
Possible to schedule a call so that I can have a run-through on this to ensure that the logic is sound?

karim-en:
@joshuajbouw yes, sure

joshuajbouw:
Just be aware everyone, need your fullest attention and responsibility when going through this PR. There was a scheduled call for tomorrow so that we can go through it.

mfornet (commented):
One observation is that this connector can be implemented completely without changing anything in the `aurora-engine`. This can be achieved via several smart contracts on Aurora and NEAR, and using `call` endpoint in `aurora` contract. I'm all for this option.

Cons: Users can't send the tokens from Aurora address to NEAR account signing though metamask, but ...

i. if the tokens are for itself it is ok to assume they will have a NEAR account with NEAR tokens.
ii. if the tokens are for someone else, they can send the tokens to the implicit Aurora address for that person (using metamask), and the target finalize the transaction on their own via a NEAR tx.

I'm all for not adding this features on the engine level, but on the smart contract level. Long term this code becomes a liability, and we are even considering moving to other engine backend which is significantly more complex after these changes.

joshuajbouw:
I love that suggestion @mfornet. Lets do it!

paouvrard:
Thoughts from the rainbowbridge product perspective.
- The usecase of sending to a custom recipient not owned by the sender is often done when sending to exchanges, so we can’t assume the recipient will be able to take action for receiving tokens.
- The solution should down the road allow direct transfers from Aurora to Ethereum using only a MetaMask wallet.
- The solution should down the road allow direct transfers between Aurora silos using only a MetaMask wallet: this will allow the widget functionality of depositing tokens from one silo to another using the widget and without leaving the dApp.

@mfornet I don't think we can assume Aurora users have a NEAR wallet when making evm to evm transfers, maybe it is doable with event relayers.

In the current `bridged aurora ERC-20` to `near NEP-141` transfers, the user sometimes has to pay storage fees for receiving tokens on NEAR which we should try not to reproduce here. Now with native Aurora ERC-20 the user even has one extra Approve step before locking tokens:
1- Pay storage fees if needed (NEAR Wallet)
2- Approve the locker contract to send tokens. (MetaMask)
3- Transfer tokens. (MetaMask)
Considering the extra Approve step I think we should aim to remove step 1 and not require NEAR wallets at all for transferring native Aurora tokens to NEAR/Ethereum/AuroraSilo.  We should now be able to more easily pay storage fees for the users because we have full control over the NEP-141 tokens minted from Aurora.

mfornet:
At this point we can close this PR since we are building a solution that doesn't require any change on the engine.

PR #437, opened 2022-02-07
test: make uniswap benchmarks reproducible

matklad:
Not sure if it's important for other tests that this is random.


birchmd (approved):
The fact that the signing account is random is not important to the test.

I assume you noticed this because of small discrepancies in gas usage due to different signing keys being used. Out of curiosity, do you know what the variance was when the key was randomly generated? Would be interesting to know what kind of variance users could expect based on their account.

matklad:
Yeah, this was to make it easier to see if the changes to evm affect gas costs. I didn't note the variance exactly, but I think it was somewhere between 10% and 1%

matklad:
@birchmd anything blocking the merge?

PR #438, opened 2022-02-09
Fix(engine): Cache generation values to avoid repeated state reads

birchmd:
We spend a lot of NEAR gas on IO, so saving reads is important. Since Sputnik caches changes to state in memory until the end of the transaction, I think that means the generation will be a constant for each address for the duration of a transaction. This means we can cache the `generation` values to avoid hitting the NEAR state multiple times.

This should save around 15% of our costs on average (as of today where the touching trie node cost is very high -- this caching might matter less in the future).


joshuajbouw (approved):
Very nice! This will be quite the improvement to throughput and reduced gas costs.

PR #439, opened 2022-02-10
optimize is_account_empty

matklad:
By making use of `&&` laziness, we save some IO operations, which adds
up to 4% for the uniswap test_uniswap_input_multihop.

### DIFF
--- a/engine/src/engine.rs
+++ b/engine/src/engine.rs
@@ -1213,10 +1213,9 @@ pub fn get_storage<I: IO>(io: &I, address: &Address, key: &H256, generation: u32
 }
 
 pub fn is_account_empty<I: IO>(io: &I, address: &Address) -> bool {
-    let balance = get_balance(io, address);
-    let nonce = get_nonce(io, address);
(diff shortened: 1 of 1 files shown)


birchmd (approved):
Nice!

matklad:
Also, a related question: it seems that aurora reads always hit host functions, and it seems that many of those reads are going to be repeated (it seems that evm's gas counter essetially multiplies io by 3 in some cases). Did you experiment with adding an in-memory cache for storage?

birchmd:
> how should I fix that?

Should be fixed in https://github.com/aurora-is-near/aurora-engine/commit/81ab4e03ecf3bc080387ee816a9c5a8a7f43f6f0

> Did you experiment with adding an in-memory cache for storage?

https://github.com/aurora-is-near/aurora-engine/pull/438 Caches one particular value we know we do not need to repeatedly read. We have not tried a more general cache though. Could be worth looking into. I am a little hesitant about a general cache because I do not know how much memory we might take up with something like that.

PR #445, opened 2022-02-15
Fix(engine): Upgrade to latest SputnikVM

birchmd:
Upgrading to latest SputnikVM includes optimizations which make our contract use 15%-20% less NEAR gas.


joshuajbouw:
Epic!

PR #454, opened 2022-02-25
Test: Reproduce GdASJ3KESs8VegpFECTveCwLQp8fxw8yvsauNEmGb6pZ gas failure

birchmd:
This test reproduces a transaction from testnet:
https://explorer.testnet.near.org/transactions/GdASJ3KESs8VegpFECTveCwLQp8fxw8yvsauNEmGb6pZ

It hit the NEAR gas limit even after the 2.4 engine release and limit increase to 300 Tgas. The purpose of having it here is to be able to track it's performance as we continue to optimize the Engine.

The test is somewhat inscrutable because the data was directly pulled from the Engine contract on testnet, but according to the partner that submitted the transaction, the high level description of what is happening is as follows:

"flashswap from uniswapv pool with call back to liquidate the user on compound and swap back the seized asset to payback the pool"


engine-tests/src/tests/uniswap.rs
```diff
@@ -54,7 +54,7 @@ fn test_uniswap_exact_output() {
 
     let (_result, profile) =
         context.add_equal_liquidity(LIQUIDITY_AMOUNT.into(), &token_a, &token_b);
-    test_utils::assert_gas_bound(profile.all_gas(), 58);
+    test_utils::assert_gas_bound(profile.all_gas(), 59);
```
joshuajbouw, line 57: Why does this change?

birchmd: Wasn't needed. The test was failing for me locally, but it was just a difference between my local solidity artifacts and the one in CI. Recompiling them locally fix it. Change removed from diff.

joshuajbouw (approved):

PR #457, opened 2022-03-02
Use EthTransactionKind reference to serialize into bytes

birchmd:
It doesn't make sense that turning `EthTransactionKind` into `Vec<u8>` consumes the object when the implementation only requires a reference. Changing the definition to `From<&EthTransactionKind>` keeps the implementation the same and allows more flexibility for users of the `aurora-engine-transactions` library.


joshuajbouw (approved):
Makes sense.

PR #458, opened 2022-03-02
Fix(Engine): Transactions to the zero address are not the same as transactions with empty to field

birchmd:
Fixes the issue raised in https://github.com/aurora-is-near/aurora-relayer/issues/197

Essentially, if the Engine treats transactions to the zero address the same as transactions with an empty `to` field, then two incorrect things happen

1. Transactions to the zero address create a new contract (this does not happen on Ethereum)
2. The Engine does not correctly derive the sender of transactions to the zero address, causing incorrect state changes and mismatch with the explorer.

This PR fixes these problems by letting the zero address be a possible value of the `to` field.

### DIFF
--- a/engine-tests/src/tests/sanity.rs
+++ b/engine-tests/src/tests/sanity.rs
@@ -16,6 +16,24 @@ const TRANSFER_AMOUNT: Wei = Wei::new_u64(123);
 const GAS_PRICE: u64 = 10;
 
 #[test]
+fn test_transaction_to_zero_address() {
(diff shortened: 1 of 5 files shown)


joshuajbouw:
As per discussion, it is best to:
* Add after a certain block height
* Use the revert

However, what worries me is the overhead on this. Is this negligible? Perhaps we should always cache the block height

joshuajbouw (approved):
I do want to take more time to explore what the original issue was, but the conversations of then are long long gone, but this looks good.

PR #462, opened 2022-03-09
Feat(Engine): Precompiles for predecessor_account_id and current_account_id

birchmd:
Related to https://github.com/aurora-is-near/aurora-engine/discussions/461

This PR provides custom precompiles to access the `predecessor_account_id` and `current_account_id` NEAR host functions from the Aurora EVM. See the discussion above for the motivation of adding these.

### DIFF
--- a/engine-precompiles/src/account_ids.rs
+++ b/engine-precompiles/src/account_ids.rs
@@ -1,0 +1,124 @@
+use super::{EvmPrecompileResult, Precompile};
+use crate::prelude::types::{Address, EthGas};
+use crate::PrecompileOutput;
+use aurora_engine_types::account_id::AccountId;
+use evm::{Context, ExitError};
+
+mod costs {
+    use crate::prelude::types::EthGas;
+
+    // TODO(#51): Determine the correct amount of gas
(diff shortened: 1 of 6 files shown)


joshuajbouw (changes requested):
This should fit in the `native.rs` file. The original idea was to group extra EVM and NEAR precompiles together into that file.

birchmd:
> This should fit in the `native.rs` file.

Given the amount of boilerplate needed for these precompiles I think it's nicer when each rust source file has only a few very closely related precompiles defined in it. Otherwise I think the files would become large and difficult to scroll through.

> The original idea was to group extra EVM and NEAR precompiles together into that file.

By this reasoning `random.rs` should be there as well, but it is currently separate (which again I think is better from a readability perspective). Perhaps a good compromise here would be to make submodules within `native` (say `native::account_ids`, `native::exit`, `native::random`) so that each definition could be in its own source file, but all of them would still be grouped in one module. If you like that idea I think it should be done as a follow-up PR though so that we are not adding new code and reorganizing existing code in the same PR.

joshuajbouw:
I like that compromise, but even then we should be refactoring the rest to state which is Ethereum's and which are our own.

joshuajbouw (approved):

mfornet (approved):

PR #463, opened 2022-03-10
Test: NEAR native benchmark compared with EVM

birchmd:
Implements a NEAR native version of the Solidity pure arithmetic benchmark to compare EVM execution cost with direct wasm cost. It is amazing that executing in the EVM takes 1500 Tgas, but with wasm directly it is less than 1 Tgas!

This gives some motivation for trying to insert pure wasm "short-cuts" into some of the common EVM execution paths. What specifically we will try to short-cut and how that will be implemented inside SputnikVM will be future work.


mfornet:
Epic discovery 🔥 

However, let's keep in mind that this is pure number crunching!

joshuajbouw (approved):
Holy crap!

PR #469, opened 2022-03-15
Test for isContract function

birchmd:
Related to ongoing support case.

Need to figure out why one transaction fails and one passes.

Passing transaction: https://aurorascan.dev/tx/0x4a538652e2f4ebda9b0f16a9af81dc894d0214f24425091099f153e36af32de4
Failing transaction: https://testnet.aurorascan.dev/tx/0xbe76feda421508f13a43d1013f4c8bd0e665e15df2187028bdf5e93bb58f24c0

This test show that the `isContract` function used in these transactions seems to work correctly in isolation, it itself may not be the problem.


joshuajbouw (approved):
Interesting find. I'll dig into it a bit as well.

EDIT: Since the change is block height dependent, would be good to have that, else we would be using the old logic.

birchmd:
> Since the change is block height dependent, would be good to have that, else we would be using the old logic.

The change is not block height dependent in the wasm artifact (which is what being used in this test)

PR #471, opened 2022-03-16
Feat(standalone): handle ft_transfer_call transactions

birchmd:
This is needed to ensure the standalone engine is catching all state changes on Aurora.

### DIFF
--- a/engine-standalone-storage/src/sync/mod.rs
+++ b/engine-standalone-storage/src/sync/mod.rs
@@ -8,7 +8,10 @@ use types::{Message, TransactionKind};
 
 const AURORA_ACCOUNT_ID: &str = "aurora";
 
-pub fn consume_message(storage: &mut crate::Storage, message: Message) -> Result<(), error::Error> {
+pub fn consume_message(
+    storage: &mut crate::Storage,
+    message: Message,
+) -> Result<ConsumeMessageOutcome, error::Error> {
     match message {
         Message::Block(block_message) => {
             let block_hash = block_message.hash;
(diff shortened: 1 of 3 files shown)


engine-standalone-storage/src/sync/mod.rs
```diff
@@ -183,11 +225,28 @@ pub fn consume_message(storage: &mut crate::Storage, message: Message) -> Result
...
+pub enum ConsumeMessageOutcome {
+    BlockAdded,
+    FailedTransactionIgnored,
+    TransactionIncluded(TransactionIncludedOutcome),
+}
+
+pub struct TransactionIncludedOutcome {
+    pub hash: aurora_engine_types::H256,
+    pub info: crate::TransactionIncluded,
+    pub diff: crate::Diff,
```
joshuajbouw, line 247: Nit, I shudder when I see `crate::` in the code, better to just include them at the top.

engine-standalone-storage/src/sync/mod.rs
```diff
@@ -122,6 +125,45 @@ pub fn consume_message(storage: &mut crate::Storage, message: Message) -> Result
...
+                            args,
+                            env.prepaid_gas,
+                        )?;
+                        let env = {
+                            let mut tmp = env.clone();
+                            // Since this would be executed as a callback, the predecessor_account_id
+                            // is now equal to the current_account_id
+                            tmp.predecessor_account_id = env.current_account_id;
+                            tmp
+                        };
```
joshuajbouw, line 145: Fine for now, but could be optimised possible with a function that does exactly this and takes in the account_id as a parameter.

joshuajbouw (approved):

engine-standalone-storage/src/sync/mod.rs
birchmd: So you would like to see it written as something like the following?

```rust
let env = env.clone().with_predecessor_account_id(env.current_account_id);
```

birchmd:
Turns out this is not sufficient, so I'll close this and make another PR which covers all the additional cases I need in the standalone.

PR #473, opened 2022-03-18
Fix(precompiles): Allow exit precompiles to work in the standalone engine

birchmd:
To fully verify the standalone engine processes transactions exactly the same as what is deployed on chain, including `SubmitResult` return values, the standalone must be able to run the exit precompiles because they create EVM logs (and therefore have an observable effect on the `SubmitResult`).

### DIFF
--- a/engine-precompiles/src/lib.rs
+++ b/engine-precompiles/src/lib.rs
@@ -21,11 +21,12 @@ use crate::bn128::{Bn128Add, Bn128Mul, Bn128Pair};
 use crate::hash::{RIPEMD160, SHA256};
 use crate::identity::Identity;
 use crate::modexp::ModExp;
-use crate::native::{ExitToEthereum, ExitToNear};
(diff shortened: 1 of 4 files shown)


mfornet (approved):

joshuajbouw (approved):
Looks good. I also do like some of the better code practices this change brings. Let's just be sure 100% that this is working given the last failure. So let's run it on local, make sure it's good to go, then let's go forward with it. Is there a possibility to change something on NEAR upstream to prevent what happened last time in our test framework...?

birchmd:
> Let's just be sure 100% that this is working

This PR doesn't add or remove any functionality from the wasm artifact, only refactors to allow running in standalone. So it should not have any negative impact on the contract. But see https://github.com/aurora-is-near/aurora-engine/pull/477 for an update to our tests that checks for the view call failure.

PR #474, opened 2022-03-18
Standalone and wasm ecrecover implementations differ

birchmd:
While testing the Borealis engine by replaying mainnet history I found that [this transaction](https://aurorascan.dev/tx/0xd3ef40dfe6fc6daf25c86dc09cb1749a3ddef08acaccfd223599afc18cd08cb6) gave a different result in the standalone than it did when it executed on chain.

The on-chain execution failed with an error message about failing to validate a signature using `ecrecover`, whereas the standalone execution passed without error (ie `ecrecover` did not fail to validate the signature). This means the implementation of the precompile `0x01` is different in the standalone than in the wasm contract. This must be resolved.

So far in this PR I have only created a unit test to show there is a problem; I have not implemented a fix yet.

### DIFF
--- a/engine-precompiles/src/secp256k1.rs
+++ b/engine-precompiles/src/secp256k1.rs
@@ -53,7 +53,7 @@ fn internal_impl(hash: H256, signature: &[u8]) -> Result<Address, ExitError> {
     Err(ExitError::Other(Borrowed(sdk::ECRecoverErr.as_str())))
 }
 
-pub(super) struct ECRecover;
+pub struct ECRecover;
 
 impl ECRecover {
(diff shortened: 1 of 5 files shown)


mfornet:
Is this ready for review already?

joshuajbouw (approved):

mfornet (approved):

joshuajbouw (commented):
Sorry, didn't mean to hit approve on this one as its missing some pending tests that will be inc.

mfornet:
> Sorry, didn't mean to hit approve on this one as its missing some pending tests that will be inc.

feel free to move it back to draft then. In general this looks good; but it would be great to have more test coverage.

PR #477, opened 2022-03-28
Fix(Engine): Predecessor id precompile works in view calls

birchmd:
This PR addresses the issue of the `predecessor_account_id` precompile failing in view calls. It does this by creating a new instance of the `Env` trait where it inserts default values for the methods that are not defined in NEAR during view calls. This instance of `Env` is used in the Engine's `view` method. To verify this fix works, our tests are also updated to put the NEAR runtime in view-call mode when executing a read-only call. As additional insurance, the call to `Env::predecessor_account_id` is made lazy; instead of calling the method when constructing the `Precompiles` struct, instead the `Env` reference itself is passed to the precompile, so that only when the `predecessor_account_id` precompile is invoked does that method get called. This means if someone accidentally calls another Engine method besides `view` as a view call, it will not necessarily fail if they do not need the `predecessor_account_id`.

### DIFF
--- a/engine-sdk/src/near_runtime.rs
+++ b/engine-sdk/src/near_runtime.rs
@@ -365,6 +365,53 @@ impl crate::promise::PromiseHandler for Runtime {
     }
 }
 
+/// Some host functions are not usable in NEAR view calls.
+/// This struct puts in default values for those calls instead.
+pub struct ViewEnv;
+
+impl crate::env::Env for ViewEnv {
+    fn signer_account_id(&self) -> AccountId {
+        AccountId::new("system").unwrap()
+    }
+
+    fn current_account_id(&self) -> AccountId {
+        unsafe {
+            exports::current_account_id(Runtime::ENV_REGISTER_ID.0);
+        }
+        Runtime::read_account_id()
+    }
+
+    fn predecessor_account_id(&self) -> AccountId {
+        AccountId::new("system").unwrap()
+    }
+
+    fn block_height(&self) -> u64 {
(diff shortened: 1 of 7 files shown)


joshuajbouw (approved):

engine-sdk/src/near_runtime.rs
```diff
@@ -365,6 +365,53 @@ impl crate::promise::PromiseHandler for Runtime {
...
+    }
+
+    fn block_timestamp(&self) -> crate::env::Timestamp {
+        let ns = unsafe { exports::block_timestamp() };
+        crate::env::Timestamp::new(ns)
+    }
+
+    fn attached_deposit(&self) -> u128 {
+        1
+    }
```
sept-en, line 399: Is there any particular reason to set `attached_deposit` as 1? Even though it's not used anywhere, I'd expect to see 0 here as otherwise, it looks a little bit misleading. Please let me know if I'm mistaken.

```suggestion
    fn attached_deposit(&self) -> u128 {
        0
    }
```

birchmd: I chose 1 because some functions ([for example `withdraw`](https://github.com/aurora-is-near/aurora-engine/blob/master/engine/src/lib.rs#L568)) require it to run and I wanted to make sure those functions could be called as well using this struct (as of now that isn't possible because there is no way for the environment to check if it is a view call or not, but maybe in the future there will be). There is no risk here because view calls do not commit to the chain, so there is no way for someone to abuse that 1 yoctoNEAR.

PR #479, opened 2022-03-30
Feat(Engine): Add custom precompile for NEAR prepaid_gas

birchmd:
Adds another custom precompile for Aurora. This one exposes the `prepaid_gas` NEAR host function. It may be useful for Aurora-based cross contract calls in the future.

### DIFF
--- a/engine-precompiles/src/prepaid_gas.rs
+++ b/engine-precompiles/src/prepaid_gas.rs
@@ -14,7 +14,7 @@ pub const ADDRESS: Address = crate::make_address(0x536822d2, 0x7de53629ef1f84c60
 mod costs {
     use crate::prelude::types::EthGas;
 
-    // TODO(#51): Determine the correct amount of gas
+    // TODO(#483): Determine the correct amount of gas
     pub(super) const PREPAID_GAS_COST: EthGas = EthGas::new(0);
 }
 
--- a/engine-precompiles/src/prepaid_gas.rs
+++ b/engine-precompiles/src/prepaid_gas.rs
@@ -1,0 +1,72 @@
+use super::{EvmPrecompileResult, Precompile};
+use crate::prelude::types::{Address, EthGas};
+use crate::PrecompileOutput;
+use aurora_engine_sdk::env::Env;
+use aurora_engine_types::{vec, U256};
+use evm::{Context, ExitError};
+
+/// predecessor_account_id precompile address
+///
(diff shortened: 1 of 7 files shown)


mfornet (approved):

engine-precompiles/src/prepaid_gas.rs
```diff
@@ -0,0 +1,72 @@
...
+/// predecessor_account_id precompile address
+///
+/// Address: `0x536822d27de53629ef1f84c60555689e9488609f`
+/// This address is computed as: `&keccak("prepaidGas")[12..]`
+pub const ADDRESS: Address = crate::make_address(0x536822d2, 0x7de53629ef1f84c60555689e9488609f);
+
+mod costs {
+    use crate::prelude::types::EthGas;
+
+    // TODO(#51): Determine the correct amount of gas
```
joshuajbouw, line 17: Would this be part of #51 todo ... ? Or just copied over. We do need to somehow determine the gas of this. If it is free, what is stopping someone from spamming this?

joshuajbouw (changes requested):

engine-precompiles/src/prepaid_gas.rs
birchmd: It was copy-pasted from the same TODO that exists on all the other NEAR-related precompiles (exit precompiles, account IDs, random). I didn't double-check that 51 was an issue (which is what I assumed) as opposed to a merged PR. All the NEAR-related precompiles currently cost 0 EVM gas and I assumed we already had an issue to do the proper estimation based on their cost in NEAR gas.

I'll create this issue and update all the numbers. But I think it should be considered out of scope for this PR. Yes, someone could call these precompiles in a loop, but it wouldn't be entirely free because calling the precompile itself costs EVM gas (cost for the `CALL` opcode for example), and the NEAR gas cost of these is pretty low, so I doubt having these be 0 EVM gas is that far off what the "correct" cost would be.

birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/479/commits/198a898091a09fe81fef3ef10f6f9e288ec122f4

joshuajbouw (approved):
While I do approve this PR, we should certainly have a system to calculate gas costs of these new Aurora-specific precompiles as a follow up before we go live.

PR #481, opened 2022-04-01
Feat(standalone): Persist transaction data

birchmd:
To enable the debug tracing API where the only input is the transaction hash, the standalone engine must persist the transaction data (ie the bytes given as input to the engine), and the transaction context (NEAR data like `predecessor_id`). This is introduced in this PR. Additionally, this PR introduces a read-only execution function. This is useful for debug tracing as well because we do not want to modify the standalone engine's DB when running trace requests.

The diff for this PR is fairly large, but most of the changes are not interesting. For example the `Borshable*` structs are just necessary boiler plate because there are not `borsh` instances for the ethereum types (and I thought it made sense to be consistent in the serialization format used in the engine's DB, so using a serialization format that is supported by the ethereum types didn't seem like a good option). And in `engine-standalone-storage/src/sync/mod.rs` the majority of the diff is simply moving some code around to avoid duplication between the read-only and committing execution modes (this should be more obvious if you tell GitHub to ignore whitespace when displaying the diff).

### DIFF
--- a/engine-standalone-storage/src/error.rs
+++ b/engine-standalone-storage/src/error.rs
@@ -5,6 +5,7 @@ use crate::TransactionIncluded;
 #[derive(Debug, PartialEq, Clone)]
 pub enum Error {
     BlockNotFound(H256),
+    Borsh(String),
     NoBlockAtHeight(u64),
     TransactionNotFound(TransactionIncluded),
     TransactionHashNotFound(H256),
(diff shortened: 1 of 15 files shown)


mfornet (approved):
Good catch! (requiring transaction data rather than just transaction hash)

joshuajbouw (approved):

PR #484, opened 2022-04-06
Fix near-blake2 dependency

mrLSD:
Issue related to `blake2_f` function compatibility.

Related to:
* [EIP-152](https://eips.ethereum.org/EIPS/eip-152)
* [Blake2 RFC-7693](https://datatracker.ietf.org/doc/html/rfc7693#section-3.2)

Realted to *EIP-152* `compress` function should accept parameters: `h, m, t, f`

### DIFF
--- a/engine-precompiles/Cargo.toml
+++ b/engine-precompiles/Cargo.toml
@@ -16,7 +16,7 @@ autobenches = false
 aurora-engine-types = { path = "../engine-types", default-features = false }
 aurora-engine-sdk = { path = "../engine-sdk", default-features = false }
 base64 = { version = "0.13.0", default-features = false, features = ["alloc"] }
-near-blake2 = { git = "https://github.com/near/near-blake2.git", version = "0.9.1", default-features = false }
+aurora-blake2 = { git = "https://github.com/aurora-is-near/aurora-blake2.git", version = "0.9.1", default-features = false }
 borsh = { version = "0.8.2", default-features = false }
 bn = { package = "aurora-bn", git = "https://github.com/aurora-is-near/aurora-bn.git", default-features = false }
 evm = { git = "https://github.com/aurora-is-near/sputnikvm.git", rev = "37448b6cacd98b06282cff5a559684505c29bd2b", default-features = false }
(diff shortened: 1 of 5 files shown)


joshuajbouw (changes requested):
While this does work, do remember the discussion we had and the requirements that we should have aurora-blake2 to be that commit, then we use our own library for that.

mrLSD:
At the moment after the research, it became obvious that the `compress` upstream `blake2` function does not meet the requirements for `EIP-152`.
What does it mean:
* We can't directly set parameters: `h, m, t, f`
* We can't set the `round` parameter
* Implementation looks like a "black box" - if we compare m with the reference implementation of `[Blake2 RFC-7693](https://datatracker.ietf.org/doc/html/rfc7693#section-3.2)`

`near-blake2` (exact commit) solve that issue.

joshuajbouw:
As we had discussed yeah, it's quite a bit unfortunate. The RFC is actually intended to be quite open. It is always scary to have to alter a cryptographic function to fit a requirement :S. Would be good to point that out to the authors, and figure out a way forward.

engine-precompiles/Cargo.toml
```diff
@@ -16,7 +16,7 @@ autobenches = false
 aurora-engine-types = { path = "../engine-types", default-features = false }
 aurora-engine-sdk = { path = "../engine-sdk", default-features = false }
 base64 = { version = "0.13.0", default-features = false, features = ["alloc"] }
-near-blake2 = { git = "https://github.com/near/near-blake2.git", version = "0.9.1", default-features = false }
+near-blake2 = { git = "https://github.com/near/near-blake2.git", rev = "cebb7df79608a7058017940830d37c37d55d21fe", default-features = false }
```
joshuajbouw, line 19: This needs to be changed to Aurora's

joshuajbouw (changes requested):

artob:
Task linked: [CU-290q9nc Engine v1.5.2](https://app.clickup.com/t/290q9nc)

artob:
Task linked: [CU-290q9nc Engine v1.5.2](https://app.clickup.com/t/290q9nc)

joshuajbouw (approved):

PR #488, opened 2022-04-15
Engine optimization: cache all reads from NEAR state

birchmd:
Previously I had tried something like this and it did not help across all transactions (some transactions with little IO were more expensive as a result), which is why I had gone for the cache to only eliminate duplicated reads.

Retrying the naïve caching approach has reduced transaction costs for heavy IO transactions, and left other transactions the same cost. I am not sure exactly what is the difference from last time; it could be the additional wasm work is less costly now since wasm opcode cost has been reduced on NEAR. But regardless, this optimization should be pushed now. In particular, it has a large impact on the test reproducing `8ru7VEA`, taking it down by 100 Tgas (almost to within the 300 Tgas limit).

We should revisit this caching after the NEAR upgrade to have trie node caching in nearcore itself since our reading of the state should be much cheaper at that point.


joshuajbouw (commented):
I am curious if this can be reduced further with the use of references in the `BTreeMap` instead of cloning address. I'm assuming that you have tried that?

Looks good. The `tgas_used` round up makes sense as well.

Do note that we should remove the `DupCache` if it isn't used anymore, and rename the file. 

This is a great improvement.

birchmd:
> use of references in the BTreeMap instead of cloning address

I do not think this is possible due to the lifetime needed on keys of the map.

> Do note that we should remove the DupCache if it isn't used anymore, and rename the file.

Done in https://github.com/aurora-is-near/aurora-engine/pull/488/commits/0ee5e0b9e0fca72fff010592ab118a08d26c8c67

joshuajbouw (approved):
Looks good. Good work!

PR #492, opened 2022-04-21
Add EVM_ prefix to contract errors

RomanHodulak:


birchmd (changes requested):
I do not think this is a good change. We should not be changing the output of the engine itself because we are beholden to the Ethereum protocol (which does not include prepending error messages from contracts). Remember that we are not the only consumer of the output of the engine.

If this tagging is needed internally for some reason then this transformation should be done in that infrastructure when it parses the engine output not in the engine itself.

joshuajbouw:
Yeah, I'm going to go with Michael's opinion here in regards to this. Sorry @RomanHodulak. Was a good exercise anyways.

PR #496, opened 2022-04-27
Feat(standalone): thread-safety and serde

birchmd:
These changes are needed for the Borealis Engine.

1. Removing the `RefCell` from `Storage` makes it thread safe; this allows it to be used in a multi-threaded application more easily
2. Deriving `serde` traits on engine types allows Borealis to use `serde_json` to send responses to `eth_call` requests.


joshuajbouw (approved):
Looks good.

PR #497, opened 2022-04-27
Refactor repro tests to have less boilerplate

birchmd:
I'll be adding more tests reproducing gas errors to our tests. This change simply factors out code that will be common to all the tests to make it easier to add the new ones.


sept-en (approved):
It's great to see such PRs. Looks really neat!

P.S. Just some minds out of loud:  perhaps, if you accept a struct containing these config fields in `repro_common`, this would make the code even more readable. But this is definitely not necessary given that it's not needed to be used that often.

joshuajbouw (approved):
Great.

birchmd:
Thanks for the suggestion @sept-en ! Done in https://github.com/aurora-is-near/aurora-engine/pull/497/commits/239e0552a81545cbe53a9045d0c5170579864ba6

PR #500, opened 2022-05-03
Fix(standalone): include possible execution errors in the TransactionIncludedOutcome 

birchmd:
This fixes an issue in the standalone engine where it would not return the full `TransactionIncludedOutcome` object on very old testnet blocks due to the current engine code not being able to read the old storage layout (before we introduced the version prefix). With this change the error will be included in the `TransactionIncludedOutcome` response so that the caller can handle the error in the same way they would handle any other `SubmitResult` error.


joshuajbouw (approved):
Looks better. Good job.

PR #501, opened 2022-05-03
Fix(connector): Return an error when storage cannot be read instead of panicking

birchmd:
This prevents an issue in the standalone engine where it can crash on old testnet blocks where the storage layout is different from the current code.

### DIFF
--- a/engine/src/engine.rs
+++ b/engine/src/engine.rs
@@ -1282,7 +1282,7 @@ pub fn remove_balance<I: IO + Copy>(io: &mut I, address: &Address) {
     EthConnectorContract::init_instance(*io)
         .map(|mut connector| {
             // The `unwrap` is safe here because (a) if the connector
-            // is implemented correctly then the total supply wll never underflow and (b) we are passing
+            // is implemented correctly then the total supply will never underflow and (b) we are passing
             // in the balance directly so there will always be enough balance.
             connector.internal_remove_eth(address, balance).unwrap();
         })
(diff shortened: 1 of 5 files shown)


engine/src/engine.rs
```diff
@@ -1277,12 +1277,16 @@ pub fn set_balance<I: IO>(io: &mut I, address: &Address, balance: &Wei) {
...
-    // is implemented correctly then the total supply wll never underflow and (b) we are passing
-    // in the balance directly so there will always be enough balance.
+    // Apply changes for eth-connector. We ignore the `StorageReadError` intentionally since
+    // if we cannot read the storage then there is nothing to remove.
     EthConnectorContract::init_instance(*io)
-        .internal_remove_eth(address, balance)
-        .unwrap();
+        .map(|mut connector| {
+            // The `unwrap` is safe here because (a) if the connector
+            // is implemented correctly then the total supply wll never underflow and (b) we are passing
```
mfornet, line 1285:
```suggestion
            // is implemented correctly then the total supply will never underflow and (b) we are passing
```

mfornet (approved):

engine/src/engine.rs
mfornet: this is an older typo :) but it is ok to fix it.

joshuajbouw (approved):
Much better practice. We should handle the rest of the `expect!`s like this so we can return a better error in light of issues. Might even reduce the contract size, I believe.

PR #514, opened 2022-05-25
Self-contained wasm contract for testing 5bEgfRQ

birchmd:
NEAR [requested](https://near.zulipchat.com/#narrow/stream/315147-aurora.2Fengine/topic/Benchmarks/near/283106730) we provide them with a self-contained wasm contract they can run easily in and NEAR runtime so that they can test changes they are making.

This PR delivers such a wasm binary. Within the aurora-engine repo, the wasm artifact can be compiled as follows:

```
aurora-is-near/aurora-engine$ make release
aurora-is-near/aurora-engine$ cd etc/self-contained-5bEgfRQ/
aurora-is-near/aurora-engine/etc/self-contained-5bEgfRQ$ cargo build --target wasm32-unknown-unknown --release
```

Then you can see the wasm binary

```
aurora-is-near/aurora-engine/etc/self-contained-5bEgfRQ$ ls -lh target/wasm32-unknown-unknown/release/self_contained_5bEgfRQ.wasm
```

To execute the benchmark, use the `run` method.

The benchmark itself is based on the [mainnet transaction](https://explorer.mainnet.near.org/transactions/5bEgfRQ5TSJfN9XCqYkMr9cgBLToM7JmS1bNzKpDXJhT) which we have measured as using [over 750 Tgas](https://github.com/aurora-is-near/aurora-engine/blob/680e1958be1f0a2acd099c2cfc48090b3095bb59/engine-tests/src/tests/repro.rs#L92), and most of which is from wasm compute as opposed to state access, therefore it is ok to use it for this self-contained use case where the wasm contract is intentionally stateless (more accurately the necessary state is baked into the binary itself to make it self-contained with a single `run` method).


nagisa:
This works great.

joshuajbouw (approved):

Issue #200, opened 2021-07-27
Get off of `no_std` and back onto `std`.


joshuajbouw:
This has been on the back of my mind for a bit but after the major and very odd Cargo bug which cost me countless hours this week, I feel that enough is enough with trying to stick with a reduced byte size at the cost of development hours and resources.

For example, we had to make libraries like https://github.com/aurora-is-near/aurora-bn to stay `no_std` compatible which cost quite a bit of time. There also is the considerable amount of hours that multiple team members, I included, faced, with trying to get Serde to work for NEP-141 and `JSON` returns.

These all cost a considerable amount of resources and time and it's certainly in excess of the additional cost of the extra bytes that it'll cost to go full `std`. This also was recommended to us by I think Max and/or Bowen from nearcore.

I had done a simple change to become compatible with `std` which only yielded a roughly 10% increase to the WASM binary size which I believe is fairly negligible.

We should discuss this and have a call but I know that all engine members are on board with this as we all have had issues. That does not mean we are not capable, it's just that the extra time it takes to stay `no_std` has cost us a fair amount of time over and over again.

austinabell:
Did you guys run into issues using [no_std Serde](https://serde.rs/no-std.html) or do you mean just trying to match serde with `rjson`?

There is also [miniserde](https://github.com/dtolnay/miniserde) if you want to optimize for code size, which doesn't come with `no_std` support but I'll play with it soon and can let you know if it works in general cases.

joshuajbouw:
Interesting. Thanks for the heads up. I'll look at it!

PR #515, opened 2022-05-26
Build(deps): Upgrade `libsecp256k1` version 0.3.5 => 0.7.0

RomanHodulak:
This resolves security issues with `Signature::parse` functions

See https://github.com/paritytech/libsecp256k1/commit/b525d5d318d9672a40250c1725fa1bb3156688b7

Closes https://app.clickup.com/t/2c7n5yy

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,7 +111,7 @@ source = "git+https://github.com/aurora-is-near/aurora-bn.git#8f1743884061981cac
 dependencies = [
  "byteorder",
  "getrandom 0.2.2",
- "rand 0.8.3",
+ "rand 0.8.5",
  "serde",
 ]
 
@@ -165,7 +165,7 @@ dependencies = [
  "ripemd160",
  "serde",
  "serde_json",
- "sha2 0.9.5",
+ "sha2",
  "sha3 0.9.1",
 ]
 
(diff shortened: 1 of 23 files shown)


joshuajbouw:
Just note that this only fixes it for local precompiles, not whats currently on NEAR. Those too would need to be looked at on nearcore as we use its host function in production.

RomanHodulak:
@joshuajbouw Ok, good to know.

Right now I'm dealing with an issue that upgrading `libsecp256k1` to `0.7.0` triggers some error that looks as if some dependency was dependent on `std` even with the `std` feature turned off.

joshuajbouw:
yes the library is non-std, so you would need to disable it in the Cargo.toml.

RomanHodulak:
Yes but I mean even when std is disabled

RomanHodulak:
I'm certain that the issue is within `aurora-blake2` dependencies.

One of which is `crypto-mac`.

And that is dependent on `subtle` and that is somehow being dependent on the `std`.

Looking at the `crypto-mac` [repository](https://github.com/RustCrypto/traits), this crate has been [discontinued and superseded](https://github.com/RustCrypto/traits/pull/819) by the `digest` crate.

joshuajbouw:
I'm updating the status to being blocked in favour of moving to std lib first. This should be done rather urgently.

See: #200

joshuajbouw:
Needs rebasing

joshuajbouw (approved):

PR #528, opened 2022-06-07
Chore(precompile): Include blake2 compression directly

joshuajbouw:
This is a better improvement since we only need the compression algorithm from [RFC 7693](https://datatracker.ietf.org/doc/html/rfc7693). 

This way it's more portable and easily maintained. It is tested with our usual blake2 tests. In theory, we may see some slight gas reduction and certainly binary reduction. We also don't need to maintain another library as this specific algorithm is straightforward.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,16 +105,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "aurora-blake2"
-version = "0.9.1"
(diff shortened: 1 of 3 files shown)


RomanHodulak:
Nicely reduced

RomanHodulak (approved):

PR #530, opened 2022-06-08
Add std wasm32 feature

mrLSD:
## Added feature `std-wasm32`.

By introducing this compilation option (feature), we are able to include the standard rust library.

#### Measured binary size (release):

* Without `std-wasm32`: 1018479
* With `std-wasm32`: 1020786 (+2.04%)
* With  `std-wasm32` and all other std features: 1184810 (+18.15%)

## Reason

There are some dependencies we use which we had to fork and make non-std. The problem is that we need to be active and update these libraries with upstream. Critical issues on these upstream libraries may not easily notify us of any issues, and we may be left with a critical bug for some time. 

The second issue is that we can be limited in the dependencies that we want to use. At times, a task which should take a day could take a week at times, for example.

Reverting to std will increase gas usage slightly, but will make the engine more maintainable.


joshuajbouw (changes requested):
Theres a bit more to be done. I wouldn't add it as a feature.Just make it only std.

These need to be removed under features in each Cargo.toml: 
`std = ["borsh/std", "evm/std", "primitive-types/std", "rlp/std", "sha3/std", "ethabi/std", "logos/std", "bn/std", "aurora-engine-types/std", "rjson/std", "aurora-engine-precompiles/std", "aurora-engine-transactions/std"]`

Also, any dependency that has `default-features = false` needs to be removed as we won't be going back to `no_std`.

engine-precompiles/Cargo.toml
```diff
@@ -27,15 +27,18 @@ ripemd160 = { version = "0.9.1", default-features = false }
...
 [dev-dependencies]
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 rand = "0.7.3"
 
 [features]
+<<<<<<< HEAD
+=======
 std = ["aurora-engine-types/std", "aurora-engine-sdk/std", "borsh/std", "bn/std", "evm/std", "evm-core/std", "libsecp256k1/std", "ripemd160/std", "sha2/std", "sha3/std", "ethabi/std"]
+>>>>>>> fd50b9b8d463196f0c228e67fd5ca83ac8dba675
```
sept-en, line 41: Looks like the merge conflict isn't resolved

joshuajbouw (approved):

joshuajbouw:
Slapping do not merge on it for now as I would prefer not to have it release until next release.

joshuajbouw:
Please resolve conflicts and squash.

PR #533, opened 2022-06-10
Fix(engine): commit compiled erc20 contract to the engine repo

### LINKED ISSUE #465: EvmErc20.bin: No such file or directory (os error 2)
When pulling aurora-engine as a library dependency compilation fails since it tries to load a file during compilation that doesn't exist: `EvmErc20.bin`

```
error: couldn't read /Users/marcelo/.cargo/git/checkouts/aurora-engine-7e378a275ccc464a/0c8db05/engine/src/../../etc/eth-contracts/res/EvmErc20.bin: No such file or directory (os error 2)
    --> /Users/marcelo/.cargo/git/checkouts/aurora-engine-7e378a275ccc464a/0c8db05/engine/src/engine.rs:1081:26
     |
1081 |     let erc20_contract = include_bytes!("../../etc/eth-contracts/res/EvmErc20.bin");
     |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this error originates in the macro `include_bytes` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Is there anyway to explicitly add this file to the package that cargo downloads while pulling engine as a dependency?

As a short term solution I manually created this file (empty) to unblock me.

birchmd:
Fixes #465


Makefile
```diff
@@ -78,7 +78,7 @@ target/wasm32-unknown-unknown/debug/aurora_engine.wasm: Cargo.toml Cargo.lock $(
 etc/eth-contracts/res/EvmErc20.bin: $(shell find etc/eth-contracts/contracts -name "*.sol") etc/eth-contracts/package.json
 	cd etc/eth-contracts && yarn && yarn build
 
-etc/eth-contracts/artifacts/contracts/test/StateTest.sol/StateTest.json: $(shell find etc/eth-contracts/contracts -name "*.sol") etc/eth-contracts/package.json
+etc/eth-contracts/artifacts/contracts/test/StateTest.sol/SelfDestruct.json: $(shell find etc/eth-contracts/contracts -name "*.sol") etc/eth-contracts/package.json
```
mfornet, line 81: Out of curiosity, why is self destruct added here (and stateTest.json) removed. Same about other changes in this file with regard with new additions.

etc/eth-contracts/res/EvmErc20V2.hex
```diff
@@ -0,0 +1 @@
...f4578283fd5b833567ffffffffffffffff8082111561120b578485fd5b61121787838801610ff0565b9450602086013591508082111561122c578384fd5b5061123986828701610ff0565b925050604084013560ff8116811461124f578182fd5b809150509250925092565b60006020828403121561126b578081fd5b5035919050565b60008060408385031215611284578182fd5b50508035926020909101359150565b600081518084526112ab816020860160208601611366565b601f01601f19169290920160200192915050565b600082516112d1818460208701611366565b9190910192915050565b6000600160f81b82526bffffffffffffffffffffffff198560601b1660018301528360158301528251611315816035850160208701611366565b91909101603501949350505050565b6000602082526110936020830184611293565b6000821982111561134a5761134a6113cd565b500190565b600082821015611361576113616113cd565b500390565b60005b83811015611381578181015183820152602001611369565b83811115610cec5750506000910152565b600181811c908216806113a657607f821691505b602082108114156113c757634e487b7160e01b600052602260045260246000fd5b50919050565b634e487b7160e01b600052601160045260246000fd5b634e487b7160e01b600052604160045260246000fd5b6001600160a01b038116811461140e57600080fd5b5056fea26469706673582212209fb2164fbe4cd313d0df109ec605706fc70da77ce5ebc7b3cfb80b30848bfa8b64736f6c63430008030033
```
mfornet, line 1: Do we need to keep in git V2?

mfornet (approved):

etc/eth-contracts/res/EvmErc20V2.hex
birchmd: V2 is not currently used, but presumably someday it will be when we eventually upgrade the ERC-20 tokens. It didn't seem harmful to commit them to the repo early.

Makefile
birchmd: As it turns out the Makefile was only working by accident previously. The file `StateTest.json` does not exist because the contract in `StateTest.sol` is called `SelfDestruct`. The Rust tests need to have the Solidity test contracts compile for them to work. That compilation did happen previously with only `mainnet-test.wasm` as the `make` requirement because the wasm artifacts depend on `EvmErc20.bin`, and when that file is not kept in the repo then it must be generated by compiling all the Solidity contracts (including the tests). However, with `EvmErc20.bin` committed to the repo, `make` can skip that Solidity compilation step when producing the wasm artifacts (of course this is intentional since we want the engine to be able to compile as a normal Cargo dependency). Therefore we must explicitly say in the Makefile that the Rust tests depend on the Solidity test contract artifacts as well. Of course this dependency was always true it just so happened that it didn't need to be expressed before because of the Solidity compilation triggered earlier in the build process.

joshuajbouw (approved):

PR #534, opened 2022-06-14
Add Backstage metadata

vzctl:
@sept-en Could you please review the precompiles piece and add description?
@joshuajbouw please check and feel free to add more info!

### DIFF
--- a/engine-precompiles/.catalog-info.yaml
+++ b/engine-precompiles/.catalog-info.yaml
@@ -3,6 +3,8 @@ apiVersion: backstage.io/v1alpha1
 kind: Component
 metadata:
   name: aurora-engine-precompiles
+  description: |-
+    EVM precompiles for the Aurora engine
   tags:
     - precompiles
     - contract
@@ -21,6 +23,8 @@ apiVersion: backstage.io/v1alpha1
 kind: Component
 metadata:
   name: aurora-engine-precompiles-native
+  description: |-
+    Aurora and NEAR specific precompile additions
   tags:
     - precompiles
     - contract
--- a/engine-precompiles/.catalog-info.yaml
+++ b/engine-precompiles/.catalog-info.yaml
@@ -1,0 +1,37 @@
+---
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: aurora-engine-precompiles
+  tags:
+    - precompiles
+    - contract
+    - near
+  links: []
+  annotations:
+    aurora.dev/security-tier: "1"
+spec:
+  owner: engine-team
+  type: contract
+  lifecycle: production
+  system: aurora-engine
+  interactsWith: []
+---
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: aurora-engine-precompiles-native
+  tags:
+    - precompiles
+    - contract
+    - near
+  links: []
+  annotations:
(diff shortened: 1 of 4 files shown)


joshuajbouw (approved):

joshuajbouw:
Does anything more need to be added to this @vzctl ? I'm not too familiar with what could be added.

engine/.catalog-info.yaml
```diff
@@ -0,0 +1,20 @@
+---
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: aurora-engine
+  title: Aurora (engine)
+  tags:
```
joshuajbouw, line 7:
```suggestion
  description: |-
    The Aurora engine
  tags:
```

engine-precompiles/.catalog-info.yaml
```diff
@@ -0,0 +1,37 @@
+---
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: aurora-engine-precompiles
+  tags:
```
joshuajbouw, line 6:
```suggestion
  description: |-
    EVM precompiles for the Aurora engine
  tags:
```

joshuajbouw (changes requested):

joshuajbouw:
The `engine-standalone-storage` and `engine-standalone-tracing` currently is owned by the Protocol team. I figure that will change in the future, however.

`engine-sdk` is owned by us, and serves as our SDK for interfacing with NEAR. It is a component.

`engine-transactions` is owned by us and includes all the different kinds of ETH transactions.

`engine-types` is a shared types library that all other modules use.

vzctl:
All looks good to me! @sept-en is it good to go from your side?

etc/eth-contracts/.catalog-info.yaml
```diff
@@ -0,0 +1,21 @@
+---
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: aurora-engine-token
+  description: |-
+    Aurora token deployed on the Aurora network
```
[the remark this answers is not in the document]
sept-en: I'd say rather

```suggestion
    Aurora ERC-20 contract for bridged tokens
```

etc/eth-contracts/.catalog-info.yaml
```diff
@@ -0,0 +1,21 @@
...
+  links: []
+  annotations:
+    aurora.dev/security-tier: "1"
+spec:
+  owner: engine-team
+  type: contract
+  lifecycle: production
+  system: aurora-engine
+  deployedAt:
+    - contract:aurora/mainnet/0x8bec47865ade3b172a928df8f990bc7f2a3b9f79
```
sept-en, line 20: This lists only one particular contract for $AURORA token. Meanwhile, there are plenty of such tokens deployed. How should we deal with this case?

sept-en (changes requested):

etc/eth-contracts/.catalog-info.yaml
vzctl: It's a very good point and we will need to implement some discovery mechanism to ingest all deployed contracts automatically from the chain.

Let's keep this out of scope for now - a bit too complicated to solve in the short term.

sept-en (approved):

joshuajbouw (approved):

PR #536, opened 2022-06-21
Fix(engine): Correctly account for changes in total supply of ETH on Aurora

birchmd:
This PR fixes a bug where the total supply on Aurora (that is the amount of ETH available in the EVM) was incorrectly lowered if a self-destructed address had a beneficiary other than itself. In that case the funds were simply moved from one address to another without any being lost even though an account was deleted.

The fix is implemented by aggregating the changes to the ETH supply in each item of the `Apply` loop done at the end of an EVM transaction. In the case of a self-destruct with beneficiary, this accounting will see the tokens removed from the deleted account, but added to the beneficiary's account, thus correctly recognizing this as net zero change. If there was no beneficiary (eg the contract specifies itself), then there will be no other `Apply` entry to offset this loss and the accounting will correctly note a net loss of ETH.

A test is included to illustrate this behaviour.

As a side-effect of this change, the `Apply` loop now looks up the values of balance and nonce from before the EVM transaction was executed (this is needed to see the changes in ETH balance used for accounting). Using these old values, the engine can save some NEAR gas because it does not need to write the same value to storage again. This results in a net decrease in the engine's gas usage.

The PR is somewhat large, but most of the changes are in the tests. The main logic changes are in `engine/src/engine.rs` and `engine/src/accounting.rs`.

### DIFF
--- a/engine/src/engine.rs
+++ b/engine/src/engine.rs
@@ -1686,7 +1686,7 @@ impl<'env, J: IO + Copy, E: Env> ApplyBackend for Engine<'env, J, E> {
                 // In tests we have convenience functions that can poof addresses with ETH out of nowhere.
                 #[cfg(all(not(feature = "integration-test"), feature = "contract"))]
                 {
-                    unreachable!()
+                    panic!("ERR_INVALID_ETH_SUPPLY_INCREASE");
                 }
             }
         }
--- a/engine/src/engine.rs
+++ b/engine/src/engine.rs
@@ -11,6 +11,7 @@ use aurora_engine_sdk::env::Env;
 use aurora_engine_sdk::io::{StorageIntermediate, IO};
 use aurora_engine_sdk::promise::{PromiseHandler, PromiseId};
 
+use crate::accounting;
 use crate::parameters::{DeployErc20TokenArgs, NewCallArgs, TransactionStatus};
 use crate::prelude::parameters::RefundCallArgs;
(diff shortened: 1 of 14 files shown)


engine/src/engine.rs
```diff
@@ -1650,13 +1653,43 @@ impl<'env, J: IO + Copy, E: Env> ApplyBackend for Engine<'env, J, E> {
...
+                    })
+                    .ok();
+            }
+            accounting::Net::Zero => (),
+            accounting::Net::Gained(_) => {
+                // It should be impossible to gain ETH using normal EVM operations in production.
+                // In tests we have convenience functions that can poof addresses with ETH out of nowhere.
+                #[cfg(all(not(feature = "integration-test"), feature = "contract"))]
+                {
+                    unreachable!()
```
joshuajbouw, line 1689: I get that this is unreachable but I would still include it as an error so that if someone somehow gets this, then they can at least refer us to an `ERR_INFLATION` or something similar error.

joshuajbouw (changes requested):

engine/src/engine.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/536/commits/de999696015d6fdc8f9a394f15295398bfc1f819

joshuajbouw (approved):

mrLSD (approved):
Awesome, good job!

PR #538, opened 2022-06-23
Feat(tracing): Implement callTracer

birchmd:
This particular kind of tracing was requested by etherscan. With it they will be able to include information about "internal transactions" on aurorascan.

### DIFF
--- a/engine-standalone-tracing/src/types/call_tracer.rs
+++ b/engine-standalone-tracing/src/types/call_tracer.rs
@@ -1,0 +1,288 @@
+//! This module defines data structure to produce traces compatible with geths "callTracer":
+//! https://github.com/ethereum/go-ethereum/blob/ad15050c7fbedd0f05a49e81400de18c2cc2c284/eth/tracers/native/call.go
+
+use aurora_engine_types::{types::Address, U256};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CallFrame {
+    call_type: CallType,
+    from: Address,
+    to: Option<Address>,
+    value: U256,
+    gas: u64,
+    gas_used: u64,
+    input: Vec<u8>,
+    output: Vec<u8>,
+    error: Option<String>,
+    calls: Vec<CallFrame>,
+}
+
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct CallTracer {
+    call_stack: Vec<CallFrame>,
+}
+
+impl CallTracer {
+    fn end(&mut self, output: Vec<u8>, error: Option<&evm::ExitReason>) {
+        let frame = self.call_stack.first_mut().unwrap();
+        match error {
+            None => frame.output = output,
+            Some(error) => {
+                let error_message = format!("{:?}", error);
+                if error_message.to_lowercase().contains("revert") {
+                    frame.output = output;
+                }
+                frame.error = Some(error_message);
+            }
+        }
+    }
+
+    fn enter(
+        &mut self,
+        call_type: CallType,
+        from: Address,
+        to: Address,
+        input: Vec<u8>,
+        gas: u64,
+        value: U256,
+    ) {
+        let frame = CallFrame {
+            call_type,
+            from,
+            to: Some(to),
+            value,
+            gas,
+            gas_used: 0,
+            input,
+            output: Vec::new(),
+            error: None,
+            calls: Vec::new(),
+        };
+        self.call_stack.push(frame);
+    }
+
+    fn exit(&mut self, output: Vec<u8>, error: Option<&evm::ExitReason>) {
+        if self.call_stack.len() <= 1 {
+            return self.end(output, error);
+        }
+
+        let mut frame = self.call_stack.pop().unwrap();
+        match error {
+            None => frame.output = output,
+            Some(error) => {
+                frame.error = Some(format!("{:?}", error));
+                match frame.call_type {
+                    CallType::Create | CallType::Create2 => frame.to = None,
+                    _ => (),
+                }
+            }
+        }
+
+        self.call_stack.last_mut().unwrap().calls.push(frame);
+    }
+
+    fn update_gas_from_snapshot(&mut self, snapshot: Option<evm_gasometer::Snapshot>) {
+        if let Some(snapshot) = snapshot {
+            if let Some(frame) = self.call_stack.last_mut() {
+                frame.gas = snapshot.gas_limit;
+                frame.gas_used = snapshot.used_gas + snapshot.memory_gas;
+            }
+        }
+    }
+
+    #[cfg(feature = "serde")]
+    pub fn serializable(mut self) -> Option<SerializableCallFrame> {
+        if self.call_stack.len() != 1 {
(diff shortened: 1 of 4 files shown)


engine-standalone-tracing/src/types/call_tracer.rs
```diff
@@ -0,0 +1,288 @@
...
+                    evm::ExitReason::Succeed(_) => None,
+                    other => Some(other),
+                };
+                self.exit(return_value.to_vec(), error);
+            }
+
+            // not useful
+            evm::tracing::Event::TransactCall { .. } => (),
+            evm::tracing::Event::TransactCreate { .. } => (),
+            evm::tracing::Event::TransactCreate2 { .. } => (),
```
joshuajbouw, line 237: Out of curiosity, why are these not useful?

birchmd: Because the same information is always duplicated in the normal `Call`/`Create` events (above code).

joshuajbouw: ah. I see. So, when it gets to it, it already would basically be returned.

engine-standalone-tracing/src/types/call_tracer.rs
```diff
@@ -0,0 +1,288 @@
...
+            gas_used: 0,
+            input,
+            output: Vec::new(),
+            error: None,
+            calls: Vec::new(),
+        };
+        self.call_stack.push(frame);
+    }
+
+    fn exit(&mut self, output: Vec<u8>, error: Option<&evm::ExitReason>) {
```
joshuajbouw, line 64: Equivalent to: https://github.com/ethereum/go-ethereum/blob/ad15050c7fbedd0f05a49e81400de18c2cc2c284/eth/tracers/native/call.go#L123

engine-standalone-tracing/src/types/call_tracer.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    fn enter(
+        &mut self,
+        call_type: CallType,
+        from: Address,
+        to: Address,
+        input: Vec<u8>,
+        gas: u64,
+        value: U256,
+    ) {
```
joshuajbouw, line 48: I don't suppose we need to skip tracing if it was interrupted here?

engine-standalone-tracing/src/types/call_tracer.rs
```diff
@@ -0,0 +1,288 @@
...
+                let error_message = format!("{:?}", error);
+                if error_message.to_lowercase().contains("revert") {
+                    frame.output = output;
+                }
+                frame.error = Some(error_message);
+            }
+        }
+    }
+
+    fn enter(
```
joshuajbouw, line 40: Equivalent to [CaptureEnter](https://github.com/ethereum/go-ethereum/blob/ad15050c7fbedd0f05a49e81400de18c2cc2c284/eth/tracers/native/call.go#L103), without the skip tracing if it was interrupted check.

engine-standalone-tracing/src/types/call_tracer.rs
```diff
@@ -0,0 +1,288 @@
...
+    input: String,
+    output: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    error: Option<String>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    calls: Vec<SerializableCallFrame>,
+}
+
+#[cfg(feature = "serde")]
+impl From<CallFrame> for SerializableCallFrame {
```
joshuajbouw, line 264: Could be equivalent to [GetResult](https://github.com/ethereum/go-ethereum/blob/ad15050c7fbedd0f05a49e81400de18c2cc2c284/eth/tracers/native/call.go#L151) if you serialize it as JSON.

engine-standalone-tracing/src/types/call_tracer.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    #[cfg(feature = "serde")]
+    pub fn serializable(mut self) -> Option<SerializableCallFrame> {
+        if self.call_stack.len() != 1 {
+            // If there is more than one element in `call_stack` then it must mean the trace did not complete
+            // because there is only 1 top-level call. Note: additional frames are added as new scopes are entered,
+            // but then the frames are coalesced as those scopes are existed.
+            return None;
+        }
```
joshuajbouw, line 100: I take it that this is the essentially [this](https://github.com/ethereum/go-ethereum/blob/ad15050c7fbedd0f05a49e81400de18c2cc2c284/eth/tracers/native/call.go#L105-L108)?

joshuajbouw:
We want to improve our PR descriptions. It would be important to include that this is required to be based on geth's [call.go](https://github.com/ethereum/go-ethereum/blob/ad15050c7fbedd0f05a49e81400de18c2cc2c284/eth/tracers/native/call.go#L105-L108).

engine-standalone-tracing/src/types/call_tracer.rs
mfornet: I don't really understand what is going on here. I did notice how `call_type` was computed in the `Call` branch, but I'm not sure why is it happening this way.

Is it maybe that StaticCall/etc... is only triggered inside the context of a Call?

mfornet (approved):

engine-standalone-tracing/src/types/call_tracer.rs
birchmd: I think all that interrupt handling is needed because tracing can be done async in the Go code, so they need a way to propagate cancellation from other async tasks. For us, all our tracing code is synchronous.

engine-standalone-tracing/src/types/call_tracer.rs
birchmd: No, it's [this](https://github.com/ethereum/go-ethereum/blob/ad15050c7fbedd0f05a49e81400de18c2cc2c284/eth/tracers/native/call.go#L152-L154).

engine-standalone-tracing/src/types/call_tracer.rs
birchmd: I chose to take this approach because all the information I need is present in the `Call`/`Create` events. It would be possible to do this by looking at the `Step` events in the `evm_runtime` module (presently handled as a no-op), and then the opcode would be directly available. But that approach seemed more complex to me overall because the `Step` events are so fine-grained.

The long story short here is that the SputnikVM tracing is structured differently than geth's tracing, and there are multiple ways of mapping one onto the other.

joshuajbouw, line 237: Just a nit, this could be improved `_ => ()`

joshuajbouw (changes requested):

joshuajbouw (approved):

engine-standalone-tracing/src/types/call_tracer.rs
birchmd: I prefer explicitly listing all variants. This makes the compiler yell at you if in the future there is a new variant added that you did not consider.

PR #540, opened 2022-06-27
Feat(precompiles): Use near host functions for alt_bn128

RomanHodulak:
Adds the ability to use the new bn128 host function on NEAR, through Aurora as the bn128 precompile.


joshuajbouw:
Please provide detailed description of why this is important.

engine-precompiles/Cargo.toml
```diff
@@ -17,7 +17,7 @@ aurora-engine-types = { path = "../engine-types", default-features = false }
 aurora-engine-sdk = { path = "../engine-sdk", default-features = false }
 base64 = { version = "0.13.0", default-features = false, features = ["alloc"] }
 borsh = { version = "0.8.2", default-features = false }
-bn = { package = "aurora-bn", git = "https://github.com/aurora-is-near/aurora-bn.git", default-features = false }
+bn = { package = "aurora-bn", git = "https://github.com/birchmd/aurora-bn.git", rev = "0ae7099d379e53d36f1883cf0d840edf67f71e7d", default-features = false }
```
birchmd, line 20: We would not need to use my fork if https://github.com/aurora-is-near/aurora-bn/pull/1 was reviewed and merged cc @joshuajbouw

engine-tests/src/tests/standard_precompiles.rs
```diff
@@ -20,20 +20,15 @@ fn standard_precompiles() {
...
-    runner.wasm_config.limit_config.max_gas_burnt = u64::MAX;
     let (_result, profile) = runner
         .submit_with_signer_profiled(&mut signer, |nonce| {
             contract.call_method("test_ecpair", nonce)
         })
         .unwrap();
 
-    // Some day this number should be less than 200 Tgas.
-    println!("{:?}", profile.all_gas());
     assert!(profile.all_gas() < 200_000_000_000_000);
```
birchmd, line 32: It would be cool to tighten the gas bound on this call. How much less than 200 Tgas is it now? It would also be nice to know the ratio of gas used by wasm vs host implementations (to get a measure of how inefficient wasm interpretation is relative to native execution).

birchmd (commented):
Thanks for this @RomanHodulak 

Note: even if it's merged, it cannot be deployed until the corresponding Near upgrade is live.

engine-precompiles/Cargo.toml
joshuajbouw: We don't need to use aurora-bn anymore at all with the new changes. Now uses zeropool-bn which has everything that we wanted from it :).

engine-tests/src/tests/standard_precompiles.rs
joshuajbouw: Tightened and made exact.

joshuajbouw:
I need this undrafted please @RomanHodulak

engine-precompiles/src/alt_bn256.rs
```diff
@@ -42,21 +42,144 @@ mod consts {
...
+
+#[cfg(feature = "contract")]
+impl SdkEncode for bn::G1 {
+    type Encoded = [u8; consts::POINT_LEN];
+
+    fn sdk_encode(self) -> Self::Encoded {
+        bn::AffineG1::from_jacobian(self)
+            .map(|p| {
+                // It is possible to use sdk_encode for Fr here but, this is
+                // just a small optimisation to avoid additional allocations.
```
birchmd, line 96: Sounds like a premature optimization to me. I would rather see the `sdk_encode` call here because it avoids code duplication and makes the code here a little easier to read. It's also not clear to me how much of an optimization this is in the first place. Rust arrays are not heap-allocated, they live on the stack, so are much less overhead to create than a `Vec`.

engine-precompiles/src/alt_bn256.rs
```diff
@@ -42,21 +42,144 @@ mod consts {
...
+            })
+            .unwrap_or_else(|| [0u8; consts::POINT_LEN])
+    }
+}
+
+#[cfg(feature = "contract")]
+impl SdkEncode for bn::G2 {
+    type Encoded = [u8; consts::POINT_PAIR_LEN];
+
+    fn sdk_encode(self) -> Self::Encoded {
```
birchmd, line 125: I wonder if there is a way to unify all 3 of these implementations. They have the same basic structure: turn the input into its component 256-bit numbers and then concatenate them all together. Maybe some cute const generic trick could work here where we write a function like `fn concat<N: usize>(xs: [U256; N]) -> [u8; 32 * N]`.

engine-precompiles/src/alt_bn256.rs
```diff
@@ -42,21 +42,144 @@ mod consts {
...
+
+    /// Size of BN pairs.
+    pub(super) const POINT_PAIR_LEN: usize = 128;
+
+    /// Output length.
+    pub(super) const OUTPUT_LEN: usize = 64;
+}
+
+#[cfg(feature = "contract")]
+trait SdkEncode {
```
birchmd, line 65: Maybe this should be named differently. Initially I was confused because we have other traits called `Sdk*` which are used to create output from the engine, so that's what I thought these would be for. Then I noticed that the implementations of this trait use little endian encoding instead of big endian, so I was concerned the implementation might be wrong. But then I got to where it was used and saw it is for passing into the NEAR host functions, not creating output, and NEAR does have a habit of preferring little endian.

Long story short, I think `HostFnEncode` might be a better name.

engine-types/src/types/gas.rs
```diff
@@ -99,6 +99,14 @@ impl Mul<usize> for EthGas {
     }
 }
 
+impl Mul<EthGas> for usize {
+    type Output = EthGas;
+
+    fn mul(self, rhs: EthGas) -> Self::Output {
+        EthGas(self as u64 * rhs.0)
```
birchmd, line 106: Note conflict with PR removing `as` conversions.

birchmd (approved):
Looks good to me. Nothing blocking, but some suggestions.

engine-types/src/types/gas.rs
joshuajbouw: I'm aware, I just am ignoring it for now and will update either or when the time comes.

engine-precompiles/src/alt_bn256.rs
```diff
@@ -42,21 +42,144 @@ mod consts {
...
+
+    fn sdk_encode(self) -> Self::Encoded {
+        let (low, high) = {
+            let [low_int, high_int] = self.into_u256().0;
+            (low_int.to_le_bytes(), high_int.to_le_bytes())
+        };
+
+        let mut bytes = [0u8; consts::SCALAR_LEN];
+        bytes[0..consts::SCALAR_PART_LEN].copy_from_slice(&low);
+        bytes[consts::SCALAR_PART_LEN..consts::SCALAR_PART_LEN * 2].copy_from_slice(&high);
```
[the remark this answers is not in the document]
joshuajbouw: There were a few of these, fixed them all in https://github.com/aurora-is-near/aurora-engine/pull/540/commits/a3111453ae5d475a4cd7f1c35c173f03313b280f

engine-precompiles/src/alt_bn256.rs
joshuajbouw: I was just quite a bit worried about how much were being used, and wasn't sure in a wasm context if this actually would cause an issue.

joshuajbouw: Actually, even then the comment didn't make a whole lot of sense given the actual use. I probably intended to do something else, but got a different result.

engine-precompiles/src/alt_bn256.rs
joshuajbouw: I had thought about doing something along those lines, honestly. I think NEAR does something similar too.

engine-precompiles/src/alt_bn256.rs
joshuajbouw: Done in https://github.com/aurora-is-near/aurora-engine/pull/540/commits/55bdbe6d8b7060e8f373d42ce895ed17aabee3e3

engine-precompiles/src/alt_bn256.rs
joshuajbouw: Done in https://github.com/aurora-is-near/aurora-engine/pull/540/commits/54faa982eeccff75ef468db27f6632f62c1decc7

engine-precompiles/src/alt_bn256.rs
```diff
@@ -42,23 +42,115 @@ mod consts {
...
+
+    /// Output length.
+    pub(super) const OUTPUT_LEN: usize = 64;
+}
+
+#[cfg(feature = "contract")]
+trait HostFnEncode {
+    type Encoded;
+
+    fn sdk_encode(self) -> Self::Encoded;
```
birchmd, line 68: Could rename to `host_fn_encode` to match the trait.

engine-precompiles/src/alt_bn256.rs
```diff
@@ -67,39 +159,52 @@ fn read_point(input: &[u8], pos: usize) -> Result<bn::G1, ExitError> {
...
         Ok(output.to_vec())
     }
+
+    #[cfg(not(feature = "contract"))]
+    fn execute(p1: bn::G1, p2: bn::G1) -> Result<[u8; consts::OUTPUT_LEN], ExitError> {
+        let mut output = [0u8; consts::POINT_LEN];
+        if let Some(sum) = bn::AffineG1::from_jacobian(p1 + p2) {
+            sum.x()
+                .to_big_endian(&mut output[0..consts::SCALAR_LEN])
+                .map_err(|_e| ExitError::Other(Borrowed("ERR_BIG_ENDIAN")))?;
```
birchmd, line 190: Could factor out `ExitError::Other(Borrowed("ERR_BIG_ENDIAN"))` as a constant since it appears multiple times throughout these precompiles.

birchmd (approved):
Nice work with the encoding refactor!

Again nothing blocking, just a couple nits.

engine-precompiles/src/alt_bn256.rs
joshuajbouw: missed!

engine-precompiles/src/alt_bn256.rs
joshuajbouw: Evgeny and I found out that WASM magically only states them once if it notices that there are multiple. Regardless, for sake of keeping up code quality, will do so.

engine-types/src/types/gas.rs
```diff
@@ -99,6 +99,14 @@ impl Mul<u64> for EthGas {
...
 }
 
+// impl Mul<EthGas> for usize {
+//     type Output = EthGas;
+//
+//     fn mul(self, rhs: EthGas) -> Self::Output {
+//         EthGas(u64::from(self) * rhs.0)
+//     }
+// }
+
```
mfornet, line 109: Remove?

engine-tests/src/tests/uniswap.rs
```diff
@@ -28,7 +28,7 @@ fn test_uniswap_input_multihop() {
     let mut context = UniswapTestContext::new("uniswap");
 
     // evm_gas = 970k
-    // near total gas = 163 Tgas
+    // near total gas = 122 Tgas
```
mfornet, line 31: 👀 Is this expected? Is uniswap using alt_bn128?

joshuajbouw: It is actually 122, it just wasn't updated.

engine-tests/Cargo.toml
```diff
@@ -38,10 +38,10 @@ hex = { version = "0.4.3", default-features = false }
...
-near-primitives-core = { git = "https://github.com/birchmd/nearcore.git", rev = "980bc48dc02878fea1e0dbc5812ae7de49f12dda" }
-near-primitives = { git = "https://github.com/birchmd/nearcore.git", rev = "980bc48dc02878fea1e0dbc5812ae7de49f12dda" }
+near-vm-runner = { git = "https://github.com/birchmd/nearcore.git", rev = "980bc48dc02878fea1e0dbc5812ae7de49f12dda", default-features = false, features = [ "wasmer2_vm", "protocol_feature_alt_bn128" ] }
+near-vm-logic = { git = "https://github.com/birchmd/nearcore.git", rev = "980bc48dc02878fea1e0dbc5812ae7de49f12dda", default-features = false, features = [ "protocol_feature_alt_bn128" ] }
+near-primitives-core = { git = "https://github.com/birchmd/nearcore.git", rev = "980bc48dc02878fea1e0dbc5812ae7de49f12dda", features = [ "protocol_feature_alt_bn128" ] }
+near-primitives = { git = "https://github.com/birchmd/nearcore.git", rev = "980bc48dc02878fea1e0dbc5812ae7de49f12dda", default-features = false, features = [ "nightly_protocol" ] }
 libsecp256k1 = { version = "0.7.0", default-features = false }
 rand = "0.8.5"
 criterion = "0.3.4"
```
hskang9, line 47: It is kind of out of the blue, but git2 is now updated to 0.15 https://github.com/rust-lang/git2-rs/releases/tag/0.15.0

hskang9 (commented):
hardhat needs to be upgraded to version `^2.10.2` in `etc/eth-contracts/package.json` to support node 18.

joshuajbouw:
> hardhat needs to be upgraded to version `^2.10.2` in `etc/eth-contracts/package.json` to support node 18.
> 
> This is the error log shown after running `cargo make --profile mainnet build-test`.
> 
> ```
> error hardhat@2.9.9: The engine "node" is incompatible with this module. Expected version "^12.0.0 || ^14.0.0 || ^16.0.0". Got "18.7.0"
> error Found incompatible module.
> ```

That doesn't need to be done this release.

engine-tests/Cargo.toml
joshuajbouw: Not related to this.

joshuajbouw (approved):

PR #549, opened 2022-07-05
Pausable Precompiles

RomanHodulak:
Drafting this PR to present the concept, perhaps we can iterate on that. Not a final state, we need to narrow down the idea conceptually.

These are loosely outlined application scenarios based on our internal discussion.

- Pause ETH Connector (already existed)
    - Is a private call
    - Stores `PauseMask` to ETH Connector contract
- Resume Precompiles
    - Needs to `require_owner_only`
    - Let `S` = `PauseMask` from Call args (0 means no action, 1 means resume)
    - Let `P` = `PauseMask` from Engine state (0 means paused, 1 means resumed)
    - Stores `P | S` to Engine state
- Pause Precompiles
    - Let `acl` = `PermissionSet` from Engine state
    - Continue if `predecessor_account_id` is in `acl` and it has a pause permission bit enabled
    - Let `S` = `PauseMask` from Call args (0 means no action, 1 means pause)
    - Let `P` = `PauseMask` from Engine state (0 means paused, 1 means resumed)
    - Stores `P & !S` to Engine state
- Grant Permissions
    - Needs to `require_owner_only`
    - Let `acl` = `PermissionSet` from Engine state
    - Add given `(account, permissions)` to the `acl` set if not present
    - Store `acl` set to Engine state
- Deny Permissions
    - Needs to `require_owner_only`
    - Let `acl` = `PermissionSet` from Engine state
    - Add given `(account, permissions)` to the `acl` set if not present
    - Store `acl` set to Engine state

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,18 +18,20 @@ jobs:
         run: |
           cache-util restore cargo_git cargo_registry sccache yarn_cache
           cache-util restore aurora-engine-target@generic@${{ hashFiles('**/Cargo.lock') }}:target
-      - name: List directories
-        run: ls -la target/wasm32-unknown-unknown/release && ls -la
       - name: Build contracts
         run: cargo make build-contracts
       - name: Test contracts
         run: cargo make test-contracts
       - name: Build mainnet test WASM
         run: cargo make --profile mainnet build-test
+      - name: List directories
+        run: ls -la target/wasm32-unknown-unknown/release && ls -la
       - name: Test mainnet
         run: cargo make --profile mainnet test-workspace
       - name: Build testnet test WASM
         run: cargo make --profile testnet build-test
+      - name: List directories
+        run: ls -la target/wasm32-unknown-unknown/release && ls -la
       - name: Test testnet
         run: cargo make --profile testnet test-workspace
       - name: Save cache


engine/src/lib.rs
```diff
@@ -196,6 +198,92 @@ mod contract {
...
+            .or_insert_with(|| PermissionMask::MAX);
+        let new_permissions = permissions | args.permission_mask;
+        *permissions = new_permissions;
+
+        engine::set_state(&mut io, state);
+    }
+
+    /// Denies permissions to given account.
+    #[no_mangle]
+    pub extern "C" fn deny_permissions() {
```
sept-en, line 223: It should be rather named `revoke_permissions()`?

sept-en (commented):
IMO if we implement ACL, we should implement something similar to OpenZeppelin's `AccessControl`. Ideally, this could be combined with paused features. `Pausable` functionality was already implemented in [near-plugins](https://github.com/aurora-is-near/near-plugins#pausable), I think it makes sense to utilize that approach which works with Rust attributes. If you plan to implement ACL, please consider contributing to that repository so we can reuse it everywhere else.

birchmd:
@RomanHodulak is this superseded by #588 ?

PR #560, opened 2022-07-25
Feat(engine): Cross contract calls

mfornet:
Enable access to trigger cross contract call from addresses internal to Aurora, to NEAR Account Ids.

AIP link: https://github.com/aurora-is-near/AIPs/pull/2/
Forum discussion: https://forum.aurora.dev/t/cross-contract-calls/304

# Notes on the design
Cross contract calls can be triggered by any aurora address calling a particular precompile. When the precompile is called, a new proxy NEAR contract will be deployed (only the first time) and the contract call will be routed through this account.

- There is a mechanism on the engine that allows upgrading the bytecode of the proxy. The proxy will be upgraded automatically if there is an update before routing the call.


engine/src/lib.rs
```diff
@@ -317,6 +318,55 @@ mod contract {
         );
     }
 
+    /// Propose new contract bytecode for proxy contracts.
+    ///
+    /// Anyone can submit a proposal. The user submitting the proposal should cover the storage
+    /// staking fee. The proposal is used when the admin invokes `factory_accept` using
+    /// the hash (sha256) of this proposal as an argument.
+    #[no_mangle]
+    pub extern "C" fn factory_propose() {
```
birchmd, line 327: I have some questions about this method, namely around spam protection. If this function is open for anyone to call, then how would we deal with someone spamming us with random "proposals"? They would eat up our NEAR storage staking room and thus prevent real EVM transactions from going through.

I think the easiest way to avoid this problem would be to make this function owner-only as well, but then it seems pointless to have a propose/accept cycle because the owner would be performing both actions.

Another possibility would be to require calls to `factory_propose` to have attached enough NEAR to cover the storage for their proposal, but then we need to compute how much NEAR that actually is. We probably would also want there to be some logic to refund the NEAR after the proposal is accepted, and a way to remove your proposal (again to get your NEAR back). This adds complexity to the flow, and I am not sure if that complexity is worthwhile. For example, if we only allow owners to change the factory code (remove propose entirely), but the contract is controlled by a DAO then there could still be a propose/accept process; it would just be in the DAO contract instead. This would have the same functionality from a user's perspective, but I think splits up the complexity a little better because the DAO already needs this kind of propose/accept functionality for voting in the first place.

birchmd: Discussed with @mfornet and we agreed to remove `propose`+`accept` in favour of a single `update` method.

engine/src/lib.rs
```diff
@@ -317,6 +318,29 @@ mod contract {
...
+
+    /// Updates the bytecode version for the given account. This is only called as a callback
+    /// when a new version of the router contract is deployed to an account.
+    #[no_mangle]
+    pub extern "C" fn factory_update_address_version() {
+        let mut io = Runtime;
+        io.assert_private_call().sdk_unwrap();
+        let args: crate::xcc::AddressVersionUpdateArgs = io.read_input_borsh().sdk_unwrap();
+        crate::xcc::set_code_version_of_address(&mut io, &args.address, args.version);
+    }
```
mfornet, line 342: Add this two methods to `TransactionKind` in `engine-standalone-storage`.

Outside of the scope of this PR: It would be great to enforce that every public method from the engine has an entry in the `TransactionKind` enum at compile level. It could be enforced in a test.

XCC.md
```diff
@@ -0,0 +1,27 @@
+# Cross contract call
```
mfornet, line 1: Let's remove this file, and I'll be writing a proper AIP.

engine-precompiles/src/xcc.rs
```diff
@@ -0,0 +1,154 @@
...
+        // It's not allowed to call cross contract call precompile in static or delegate mode
+        if is_static {
+            return Err(ExitError::Other(Cow::from(ERR_STATIC)));
+        } else if context.address != cross_contract_call::ADDRESS.raw() {
+            return Err(ExitError::Other(Cow::from(ERR_DELEGATE)));
+        }
+
+        let sender = context.caller;
+        let target_account_id = create_target_account_id(sender, self.engine_account_id.as_ref());
+        // TODO: Is it ok to use Borsh to read the input? It might not be very friendly to construct the input in Solidity...
```
mfornet, line 88: Yes. The borsh implementation in solidity is very efficient, probably more than using Abi encoding (I don't have numbers to share, but we did these experiments for the bridge previously).

Ultimately users should use `aurora-solidity-sdk`, so this is abstracted away.

engine/src/xcc.rs
```diff
@@ -0,0 +1,172 @@
...
+                name: "initialize".into(),
+                args: Vec::new(),
+                attached_yocto: ZERO_YOCTO,
+                gas: INITIALIZE_GAS,
+            });
+            // After the contract is deployed and initialized, we can call the method requested
+            promise_actions.push(PromiseAction::FunctionCall {
+                name: promise.method,
+                args: promise.args,
+                attached_yocto: promise.attached_balance,
```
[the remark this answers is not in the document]
mfornet: I see now that in the precompile we always use zero as the attached balance. Anyway, I still think we should add here a check that this amount is zero since we are not expecting any value greater than zero now or in the future. At least not without extra logic that makes the user pay for this balance in advance.

engine-precompiles/src/xcc.rs
```diff
@@ -0,0 +1,154 @@
...
+        let args = CrossContractCallArgs::try_from_slice(input)
+            .map_err(|_| ExitError::Other(Cow::from(ERR_INVALID_INPUT)))?;
+        let promise = match args {
+            CrossContractCallArgs::Eager(call) => {
+                let call_gas = match &call {
+                    PromiseArgs::Create(call) => call.attached_gas,
+                    PromiseArgs::Callback(cb) => cb.base.attached_gas + cb.callback.attached_gas,
+                };
```
mfornet, line 96: Make a new function `total_gas`:

```suggestion
                let call_gas = call.total_gas();
```

engine-precompiles/src/xcc.rs
```diff
@@ -0,0 +1,154 @@
...
+                // since the promise is not executed right away.
+                attached_gas: costs::ROUTER_SCHEDULE,
+            },
+        };
+
+        let promise_log = Log {
+            address: cross_contract_call::ADDRESS.raw(),
+            topics: Vec::new(),
+            data: promise
+                .try_to_vec()
```
mfornet, line 124: Thinking out loud here: it is unfortunate that we need to serialize the promise here, and then deserialize it back later. This could be potentially saved, but it is not trivial (and outside of the scope of this PR).

mfornet (commented):
Everything looks good! Already reviewed what is in the code, I'll do a future review about what could be potentially missing from it.

I can't approve / reject this PR as I created it 🤦 .

engine/src/xcc.rs
```diff
@@ -0,0 +1,172 @@
...
+    let value_bytes = version.0.to_le_bytes();
+    io.write_storage(&key, &value_bytes);
+}
+
+/// Private utility method for reading code version from storage.
+fn read_version<I: IO>(io: &I, key: &[u8]) -> Option<CodeVersion> {
+    match io.read_u32(key) {
+        Ok(value) => Some(CodeVersion(value)),
+        Err(ReadU32Error::MissingValue) => None,
+        Err(ReadU32Error::InvalidU32) => panic!("{}", ERR_CORRUPTED_STORAGE),
```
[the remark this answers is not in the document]
birchmd: Yes, I was allowing panics in places where the only way the code reaches that point is due to a critical bug (so we want to fail fast to avoid more incorrect operations). The reason I am not using NEAR panics is because the standalone engine should also panic at this spot, not just the contract.

engine/src/xcc.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/// Sets the latest router contract version. This function is intentionally private because
+/// it should never be set manually. The version is managed automatically by `update_router_code`.
+fn set_latest_code_version<I: IO>(io: &mut I, version: CodeVersion) {
+    let key = storage::bytes_to_key(KeyPrefix::CrossContractCall, VERSION_KEY);
+    let value_bytes = version.0.to_le_bytes();
+    io.write_storage(&key, &value_bytes);
+}
+
```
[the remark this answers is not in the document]
birchmd: Personally I prefer to see immutable and mutable functions grouped together as opposed to having them grouped by  what field is being accessed. What do others think about this?

engine/src/xcc.rs
```diff
@@ -0,0 +1,172 @@
...
+    io: &I,
+    handler: &mut P,
+    promise: PromiseCreateArgs,
+    current_account_id: &AccountId,
+) where
+    P: PromiseHandler,
+    I: IO,
+{
+    let target_account: &str = promise.target_account_id.as_ref();
+    let sender = Address::decode(&target_account[0..40]).expect(ERR_INVALID_ACCOUNT);
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/560/commits/f04d09e6fef7b9b5d5599f4364720a3a56871224

etc/xcc-router/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+
+#[derive(BorshSerialize, BorshStorageKey)]
+enum StorageKey {
+    Version,
+    Parent,
+    Nonce,
+    Map,
+}
+
+const CURRENT_VERSION: u32 = 0;
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/560/commits/4fe43be3fcb1554591659a0852af181bacf8aee9

engine/src/xcc.rs
```diff
@@ -0,0 +1,172 @@
...
+
+pub const ERR_NO_ROUTER_CODE: &str = "ERR_MISSING_XCC_BYTECODE";
+pub const ERR_CORRUPTED_STORAGE: &str = "ERR_CORRUPTED_XCC_STORAGE";
+pub const ERR_INVALID_ACCOUNT: &str = "ERR_INVALID_XCC_ACCOUNT";
+pub const VERSION_KEY: &[u8] = b"version";
+pub const CODE_KEY: &[u8] = b"router_code";
+pub const VERSION_UPDATE_GAS: NearGas = NearGas::new(5_000_000_000_000);
+pub const INITIALIZE_GAS: NearGas = NearGas::new(5_000_000_000_000);
+/// Amount of NEAR needed to cover storage for a router contract.
+pub const STORAGE_AMOUNT: Yocto = Yocto::new(5_000_000_000_000_000_000_000_000);
```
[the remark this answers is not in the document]
birchmd: 2 NEAR is not enough for the tests to pass, so more storage staking must be needed. I reduced from 5 NEAR to 4 in https://github.com/aurora-is-near/aurora-engine/pull/560/commits/7815ee44ab96a51bde1e7cf578a09cbdbb4f423a

engine/src/xcc.rs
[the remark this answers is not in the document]
birchmd: Added a check promise.attached_balance is zero in https://github.com/aurora-is-near/aurora-engine/pull/560/commits/0b55a183fc7fd5f3b8435bf8c09a128b49d5d4cd

engine/src/lib.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/560/commits/1dfe17bde9b10f1223af748b948e42c885881448

XCC.md
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/560/commits/53c220eef327617c4caa8f871b0e633731f36124

engine-precompiles/src/xcc.rs
birchmd: Cool, TODO removed in https://github.com/aurora-is-near/aurora-engine/pull/560/commits/e8d0a2472370c15afb334a79b059d628c78ef373

engine-precompiles/src/xcc.rs
birchmd: It's not here that we re-serialize, it is when we serialize `call` (in the `promise.args` field). This could be avoided by manually matching on the first byte of the input (which selects between `Eager` and `Delayed` variants), but this would be more error prone because we would need to remember to change the parsing here any time `CrossContractCallArgs` was updated. Another disadvantage of not parsing the input entirely is that we then are not validating it is a real `PromiseArgs` argument, so it would not be until the call hit the router contract that it would fail. I think it is a much better developer experience to have it fail in the EVM.

For these reasons, I think it worth spending the extra gas to do the round-trip serialization.

engine/src/xcc.rs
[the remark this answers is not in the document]
mfornet: This is probably because you are compiling the contract without optimizations applied. 

I use this line to compile the final binary:

`env RUSTFLAGS='-C link-arg=-s' cargo build --target wasm32-unknown-unknown --release`

engine/src/xcc.rs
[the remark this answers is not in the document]
mfornet: got it, sounds good.

engine/src/xcc.rs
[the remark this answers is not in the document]
birchmd: Yes, that did it, thanks! Fixed in https://github.com/aurora-is-near/aurora-engine/pull/560/commits/b9810cd9f49e8b40e6db4d2adec8f0e4b4518d5e

engine/src/xcc.rs
```diff
@@ -0,0 +1,172 @@
...
+        Some(_version) => AddressVersionStatus::UpToDate,
+    };
+    let _promise_id = match deploy_needed {
+        AddressVersionStatus::DeployNeeded { create_needed } => {
+            let mut promise_actions = Vec::with_capacity(4);
+            if create_needed {
+                promise_actions.push(PromiseAction::CreateAccount);
+                promise_actions.push(PromiseAction::Transfer {
+                    amount: STORAGE_AMOUNT,
+                });
```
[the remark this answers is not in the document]
birchmd: Discussed with @mfornet and we settled on a modification to the xcc flow where users can pay for their calls using wNEAR in the EVM. I'll work on implementing it now.

joshuajbouw (changes requested):
All in all looks good. I don't have any issues per se with the logic itself. I want to go once again when we reach a final review and try to come up with scenarios on how we could break this. I have not done a security analysis of this yet.

engine/src/xcc.rs
```diff
@@ -0,0 +1,202 @@
...
+    I: IO,
+{
+    let target_account: &str = promise.target_account_id.as_ref();
+    let sender = Address::decode(&target_account[0..40]).expect(ERR_INVALID_ACCOUNT);
+
+    // Confirm target_account is of the form `{address}.{aurora}`
+    // Address prefix parsed above, so only need to check `.{aurora}`
+    assert_eq!(&target_account[40..41], ".", "{}", ERR_INVALID_ACCOUNT);
+    assert_eq!(
+        &target_account[41..],
```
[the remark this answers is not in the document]
birchmd: I think this is exactly the case where we do not want an ending index. The assert is to check that whole rest of the string matches the current account id. It could be done by checking the total length and then including the length bound in this check, but I don't see what security that adds over how it is written currently.

engine-precompiles/src/xcc.rs
```diff
@@ -0,0 +1,150 @@
...
+                return Err(ExitError::OutOfGas);
+            }
+        }
+
+        // It's not allowed to call cross contract call precompile in static or delegate mode
+        if is_static {
+            return Err(ExitError::Other(Cow::from(ERR_STATIC)));
+        } else if context.address != cross_contract_call::ADDRESS.raw() {
+            return Err(ExitError::Other(Cow::from(ERR_DELEGATE)));
+        }
```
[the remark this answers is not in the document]
birchmd: It's possible I'm being overly conservative here, but I can't think of a legitimate use case for using a delegate call on this precompile that wouldn't potentially be exploitable (for example creating a promise on behalf of a user without them knowing).

I think it's safer to start with this limitation and remove it later if there is a legitimate use case.

engine/src/xcc.rs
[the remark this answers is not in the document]
joshuajbouw: Sure, thats fair. I'll reconsider limiting that then.

engine-precompiles/src/xcc.rs
```diff
@@ -0,0 +1,150 @@
...
+use borsh::{BorshDeserialize, BorshSerialize};
+use evm::backend::Log;
+use evm_core::ExitError;
+
+const ERR_INVALID_INPUT: &str = "ERR_INVALID_XCC_INPUT";
+const ERR_SERIALIZE: &str = "ERR_XCC_CALL_SERIALIZE";
+const ERR_STATIC: &str = "ERR_INVALID_IN_STATIC";
+const ERR_DELEGATE: &str = "ERR_INVALID_IN_DELEGATE";
+const ROUTER_EXEC_NAME: &str = "execute";
+const ROUTER_SCHEDULE_NAME: &str = "schedule";
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/560/commits/4b516256bcf30a43e8dff4dcf8d9c3c72681367a

engine-precompiles/src/xcc.rs
```diff
@@ -0,0 +1,150 @@
...
+
+pub mod costs {
+    use crate::prelude::types::{EthGas, NearGas};
+
+    // TODO(#483): Determine the correct amount of gas
+    pub(super) const CROSS_CONTRACT_CALL: EthGas = EthGas::new(0);
```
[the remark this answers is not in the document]
birchmd: https://github.com/aurora-is-near/aurora-engine/pull/564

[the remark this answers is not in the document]
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/560/commits/5fa6694f5c25cd5a1b9836fc4a4b584001f22aa7

engine-precompiles/src/xcc.rs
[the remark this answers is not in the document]
joshuajbouw: Sure, lets keep it as such then.

joshuajbouw:
This is blocked until #565 is in, correct?

engine-sdk/Cargo.toml
```diff
@@ -14,7 +14,7 @@ autobenches = false
 
 [dependencies]
 aurora-engine-types = { path = "../engine-types", default-features = false }
-borsh = { version = "0.8.2", default-features = false }
+borsh = { version = "0.9", default-features = false }
```
joshuajbouw, line 17: To prevent potential issues, can you please peg this to 0.9.3 in all Cargo.toml files?

joshuajbouw (changes requested):

engine-sdk/Cargo.toml
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/560/commits/2d91055d16d54651b1ca143f34ccd6ce1921d856

engine/src/xcc.rs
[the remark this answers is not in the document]
birchmd: https://github.com/aurora-is-near/aurora-engine/pull/572

engine-precompiles/src/native.rs
```diff
@@ -30,12 +30,10 @@ mod costs {
     // TODO(#483): Determine the correct amount of gas
     pub(super) const EXIT_TO_ETHEREUM_GAS: EthGas = EthGas::new(0);
 
-    // TODO(#332): Determine the correct amount of gas
-    pub(super) const FT_TRANSFER_GAS: NearGas = NearGas::new(100_000_000_000_000);
+    pub(super) const FT_TRANSFER_GAS: NearGas = NearGas::new(10_000_000_000_000);
```
joshuajbouw, line 33: I don't want to block this PR, but I do believe that rationale for this value should be contributed. Same for the one blow.

engine-types/src/parameters.rs
```diff
@@ -64,3 +81,18 @@ pub struct RefundCallArgs {
...
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub enum CrossContractCallArgs {
+    /// The promise is to be executed immediately (as part of the same NEAR transaction as the EVM call).
+    Eager(PromiseArgs),
+    /// The promise is to be stored in the router contract, and can be executed in a future transaction.
+    /// The purpose of this is to expand how much NEAR gas can be made available to a cross contract call.
+    /// For example, if an expensive EVM call ends with a NEAR cross contract call, then there may not be
+    /// much gas left to perform it. In this case, the promise could be `Delayed` (stored in the router)
+    /// and executed in a separate transaction with a fresh 300 Tgas available for it.
+    Delayed(PromiseArgs),
```
joshuajbouw, line 97: What if these are never consumed? Could we have some process to free these up if they are old? Or should we?

engine/src/xcc.rs
```diff
@@ -0,0 +1,282 @@
...
+
+pub const ERR_NO_ROUTER_CODE: &str = "ERR_MISSING_XCC_BYTECODE";
+pub const ERR_INVALID_ACCOUNT: &str = "ERR_INVALID_XCC_ACCOUNT";
+pub const ERR_ATTACHED_NEAR: &str = "ERR_ATTACHED_XCC_NEAR";
+pub const CODE_KEY: &[u8] = b"router_code";
+pub const VERSION_UPDATE_GAS: NearGas = NearGas::new(5_000_000_000_000);
+pub const INITIALIZE_GAS: NearGas = NearGas::new(15_000_000_000_000);
+pub const UNWRAP_AND_REFUND_GAS: NearGas = NearGas::new(25_000_000_000_000);
+pub const WITHDRAW_GAS: NearGas = NearGas::new(30_000_000_000_000);
+pub const WITHDRAW_TO_NEAR_SELECTOR: [u8; 4] = [0x6b, 0x35, 0x18, 0x48];
```
joshuajbouw, line 20: This should explain what the 4 bytes are meant to represent.

engine/src/xcc.rs
```diff
@@ -0,0 +1,282 @@
...
+use borsh::{BorshDeserialize, BorshSerialize};
+
+pub const ERR_NO_ROUTER_CODE: &str = "ERR_MISSING_XCC_BYTECODE";
+pub const ERR_INVALID_ACCOUNT: &str = "ERR_INVALID_XCC_ACCOUNT";
+pub const ERR_ATTACHED_NEAR: &str = "ERR_ATTACHED_XCC_NEAR";
+pub const CODE_KEY: &[u8] = b"router_code";
+pub const VERSION_UPDATE_GAS: NearGas = NearGas::new(5_000_000_000_000);
+pub const INITIALIZE_GAS: NearGas = NearGas::new(15_000_000_000_000);
+pub const UNWRAP_AND_REFUND_GAS: NearGas = NearGas::new(25_000_000_000_000);
+pub const WITHDRAW_GAS: NearGas = NearGas::new(30_000_000_000_000);
```
joshuajbouw, line 19: These gas costs should be explained.

etc/xcc-router/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+
+const CURRENT_VERSION: u32 = 1;
+
+const ERR_ILLEGAL_CALLER: &str = "ERR_ILLEGAL_CALLER";
+const WNEAR_WITHDRAW_GAS: Gas = Gas(5_000_000_000_000);
+const WNEAR_REGISTER_GAS: Gas = Gas(5_000_000_000_000);
+const REFUND_GAS: Gas = Gas(5_000_000_000_000);
+const WNEAR_REGISTER_AMOUNT: u128 = 1_250_000_000_000_000_000_000;
+/// Must match arora_engine_precompiles::xcc::state::STORAGE_AMOUNT
+const REFUND_AMOUNT: u128 = 2_000_000_000_000_000_000_000_000;
```
joshuajbouw, line 28: These gas costs should be explained.

joshuajbouw (approved):
My only real complaint is that there is a lot of lacking documentation that will not only help ourselves, but also others and especially the auditors.

engine/src/lib.rs
```diff
@@ -318,6 +319,40 @@ mod contract {
...
+
+    /// Sets the address for the wNEAR ERC-20 contract. This contract will be used by the
+    /// cross-contract calls feature to have users pay for their NEAR transactions.
+    #[no_mangle]
+    pub extern "C" fn factory_set_wnear_address() {
+        let mut io = Runtime;
+        let state = engine::get_state(&io).sdk_unwrap();
+        require_owner_only(&state, &io.predecessor_account_id());
+        let address = io.read_input_arr20().sdk_unwrap();
```
joshuajbouw, line 352: I would recommend these to be feature gated as well.

joshuajbouw (changes requested):
Sorry, forgot to say that we require all the `engine/src/lib.rs` new functions should also be feature gated.

joshuajbouw (approved):
It's fine, @birchmd had informed me that this should not be an issue in the end, and I do agree.

birchmd:
> My only real complaint is that there is a lot of lacking documentation that will not only help ourselves, but also others and especially the auditors.

I'll address requests for more documentation in a follow-up PR.

engine-types/src/parameters.rs
birchmd: Ultimately it is the user that has to cover the storage staking for stored promises they never used. I don't know why they would not use the promises they have scheduled. We could add a cancellation interface, but this adds more complexity to the design, and I do not think it is worth having that complexity for this edge case that may never come up.

mfornet: Also this bring the question: who can destroy it and when? probably the answer is the same user that scheduled it, but is still a bit pointless! Agree on leaving it there until it is consumed!

PR #561, opened 2022-07-28
Chore: Disallow as_conversions

joshuajbouw:
## Description

While we have been verbally careful about `as` conversions, i.e `1234u16 as u64`, it is best to include a clippy deny to ensure that this is actually followed. There were many conversions, and we were not treating `usize` conversions as safe as we should've, but I have not found any of the usages to cause issues. This simply is a preventative change to ensure that we do not accidentally include a possible problem in the future.

There are some extra additions to this code which I found to be too small for their own PR and would've benefited this one anyways. This includes the slight change in the `.env` folder files which the latest version of CLion now warns about, so were changed to reduce the amount of warning noise to nil.

Since there are extra checks in place as we dropped `as`, I anticipate a slight gas increase.

## Testing

There does not need to be any additional testing besides the current tests that currently exist.

Reviewers should be careful to understand why using `as` to downcast integers from a larger type to a smaller one can be damaging. For example`u64` to `u32`. This could cause bits to be thrown away carelessly, causing unintentional problems.

## Additional information

I found a small issue with the Clippy command in the `Makefile.toml` file. It only captures clippy with the contract feature but we should run more clippy checks for other features. A follow up PR should be made.


Makefile.toml
```diff
@@ -69,9 +69,7 @@ category = "Check"
 command = "${CARGO}"
 args = [
     "clippy",
-    "--no-default-features",
-    "--features",
-    "contract",
+    "--all-features",
```
birchmd, line 72: This feels a little disingenuous. In a typical Rust project where you have some code that is only compiled when there is a feature present, and none of those features are mutually exclusive, then this works fine. But in our case we have code that is compiled on the presence _or absence_ of a feature (namely `contract`). This means `--all-features` does not actually check all the code we care about because it will miss what is behind `#[cfg(not(feature = "contract"))]`.

That said, the original `--features contract` version of this command obviously didn't check that code either, but my main point is `--all-features` might give a false sense of security. Maybe it makes sense to add multiple `clippy` commands with different feature sets.

engine-precompiles/src/bn128.rs
```diff
@@ -353,8 +354,12 @@ impl<HF: HardFork> Bn128Pair<HF> {
 
 impl Precompile for Bn128Pair<Byzantium> {
     fn required_gas(input: &[u8]) -> Result<EthGas, ExitError> {
+        let input_len = u64::try_from(input.len())
+            .map_err(|_e| ExitError::Other(Borrowed("ERR_USIZE_CONVERSION")))?;
```
birchmd, line 358: I wonder if it makes sense to define a function like

```rust
const fn err_usize_conv<E: Error>(_e: E) -> ExitError {
  ExitError::Other(Borrowed("ERR_USIZE_CONVERSION"))
}
```

Then all these `try_from` calls can look like

```rust
let input_len = u64::try_from(input.len()).map_err(err_usize_conv)?;
```

I think that's a little nicer because (a) it doesn't repeat the `ERR_USIZE_CONVERSION` string multiple times, and (b) it's easier to read in my opinion.

engine-standalone-tracing/Cargo.toml
```diff
@@ -21,6 +21,7 @@ evm-core = { git = "https://github.com/aurora-is-near/sputnikvm.git", rev = "374
 evm = { git = "https://github.com/aurora-is-near/sputnikvm.git", rev = "37448b6cacd98b06282cff5a559684505c29bd2b", default-features = false, features = ["std", "tracing"] }
 evm-runtime = { git = "https://github.com/aurora-is-near/sputnikvm.git", rev = "37448b6cacd98b06282cff5a559684505c29bd2b", default-features = false, features = ["std", "tracing"] }
 evm-gasometer = { git = "https://github.com/aurora-is-near/sputnikvm.git", rev = "37448b6cacd98b06282cff5a559684505c29bd2b", default-features = false, features = ["std", "tracing"] }
+hex = { version = "0.4", default-features = false, features = ["alloc"] }
```
birchmd, line 24: What is this for?

engine-transactions/src/eip_1559.rs
```diff
@@ -79,7 +80,7 @@ impl SignedTransaction1559 {
             message_hash,
             &super::vrs_to_arr(self.parity, self.r, self.s),
         )
-        .ok()
+        .map_err(|_e| ParseTransactionError::InvalidV)
```
birchmd, line 83: I think this should be the `EcRecover` variant.

engine-transactions/src/eip_2930.rs
```diff
@@ -95,7 +96,7 @@ impl SignedTransaction2930 {
             message_hash,
             &super::vrs_to_arr(self.parity, self.r, self.s),
         )
-        .ok()
+        .map_err(|_e| ParseTransactionError::InvalidV)
```
birchmd, line 99: I think this should be the `EcRecover` variant.

engine-transactions/src/legacy.rs
```diff
@@ -45,20 +46,20 @@ impl TransactionLegacy {
...
-            gas_limit: self.gas_limit,
-            max_priority_fee_per_gas: self.gas_price,
-            max_fee_per_gas: self.gas_price,
-            to: self.to,
-            value: self.value,
-            data: self.data,
-            access_list: Vec::new(),
-        }
-    }
+    // pub fn normalize(self) -> super::NormalizedEthTransaction {
```
birchmd, line 49: Please delete commented out code if it is not needed.

engine-transactions/src/eip_1559.rs
```diff
@@ -70,7 +71,7 @@ pub struct SignedTransaction1559 {
 }
 
 impl SignedTransaction1559 {
-    pub fn sender(&self) -> Option<Address> {
+    pub fn sender(&self) -> Result<Address, ParseTransactionError> {
```
birchmd, line 74: I don't know if expanding `ParseTransactionError` to include variants for signature fails makes sense. The transaction has already been parsed into our data structure by this point. The question is whether the data that we parsed contains a valid signature or not. Maybe it makes sense to have a `SignatureVerificationError` type or something like that.

engine-transactions/src/lib.rs
```diff
@@ -65,7 +66,7 @@ impl<'a> From<&'a EthTransactionKind> for Vec<u8> {
 /// A normalized Ethereum transaction which can be created from older
 /// transactions.
 pub struct NormalizedEthTransaction {
-    pub address: Option<Address>,
+    pub address: Address,
```
birchmd, line 69: As far as I remember, this `NormalizedEthTransaction` struct was inspired by a corresponding definition in geth. If the option is used in geth, does that mean there is a use case where we want to be able to define a transaction without a valid `from` address? I can't think of any myself, but I thought I would pose the question anyway.

engine-transactions/src/lib.rs
```diff
@@ -117,12 +120,12 @@ impl From<EthTransactionKind> for NormalizedEthTransaction {
...
                 access_list: tx.transaction.access_list,
             },
-        }
+        })
     }
 }
 
 impl NormalizedEthTransaction {
-    pub fn intrinsic_gas(&self, config: &evm::Config) -> Option<u64> {
+    pub fn intrinsic_gas(&self, config: &evm::Config) -> Result<u64, ParseTransactionError> {
```
birchmd, line 128: Again I think a separate error enum might make sense here. This function is not really about parsing, it's about another validation step.

engine-transactions/src/lib.rs
```diff
@@ -131,31 +134,47 @@ impl NormalizedEthTransaction {
...
+        let gas_access_list_storage = config
+            .gas_access_list_storage_key
+            .checked_mul(
+                u64::try_from(
+                    self.access_list
+                        .iter()
+                        .map(|a| a.storage_keys.len())
+                        .sum::<usize>(),
+                )
+                .map_err(|_e| ParseTransactionError::GasOverflow)?,
```
birchmd, line 168: Should be `IntegerConversion` variant.

engine-standalone-tracing/Cargo.toml
joshuajbouw: Clippy was failing here, saying that hex didn't exist in this library. Not sure why that suddenly was happening.

engine-transactions/src/legacy.rs
joshuajbouw: Its still in draft, got it like this for now until I fix these tests.

Makefile.toml
joshuajbouw: I had been thinking about this as well as I had it miss a problem just running it normally.

engine-precompiles/src/bn128.rs
joshuajbouw: Was going to be going back over this and cleaning it up, just wanted the barebones in there first so that I knew how much of these I would be dealing with. That is a good suggestion though.

engine-transactions/src/eip_1559.rs
joshuajbouw: Instead, I changed this to `Error` as its a crate Error. That also works.

engine-transactions/src/lib.rs
joshuajbouw: It wasn't an option. I believe an `Option` was lazily used here not to make an error type.

engine-types/src/types/mod.rs
```diff
@@ -76,22 +76,14 @@ pub struct FtResolveTransferResult {
...
-    ArgumentParseError,
-    InvalidMetaTransactionMethodName,
-    InvalidMetaTransactionFunctionArg,
-    InvalidEcRecoverSignature,
-}
-
 const HEX_ALPHABET: &[u8; 16] = b"0123456789abcdef";
 
 #[allow(dead_code)]
 pub fn bytes_to_hex(v: &[u8]) -> String {
```
birchmd, line 82: Should probably just delete this function and replace call to it with `hex::encode`.

engine/src/json.rs
```diff
@@ -90,10 +90,13 @@ impl JsonValue {
     pub fn parse_u8(v: &JsonValue) -> Result<u8, JsonError> {
         match v {
             JsonValue::U64(n) => {
-                if *n > u8::MAX as u64 {
+                if *n > u64::from(u8::MAX) {
                     Err(JsonError::OutOfRange(JsonOutOfRangeError::OutOfRangeU8))
                 } else {
-                    Ok(*n as u8)
+                    match u8::try_from(*n) {
```
birchmd, line 96: This is redundant after the `if` check above. Probably better to simply replace the `if` check with this `match`.

birchmd (commented):
Looks pretty good overall, just a few more comments.

engine-types/src/types/mod.rs
joshuajbouw: Kind of unrelated to this PR though. That, and it'll be changed in the future anyways with the change back to serde.

.env/testnet.env
```diff
@@ -1,9 +1,9 @@
...
-NEAR_EVM_ACCOUNT = "aurora"
-WASM_FILE = "aurora-testnet.wasm"
-WASM_FILE_TEST = "aurora-testnet-test.wasm"
-NEAR_CLI = "near"
-IS_PROD = false
+CARGO_FEATURES_BUILD="testnet"
+CARGO_FEATURES_BUILD_TEST="testnet,integration-test"
+CARGO_FEATURES_TEST="testnet-test"
+RUSTC_FLAGS_BUILD="-C link-arg=-s"
+NEAR_EVM_ACCOUNT="aurora.test.near"
```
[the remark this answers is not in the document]
joshuajbouw: Will fix it.

engine/src/json.rs
joshuajbouw: Good catch.

engine-standalone-tracing/Cargo.toml
```diff
@@ -21,6 +21,7 @@ evm-core = { git = "https://github.com/aurora-is-near/sputnikvm.git", tag = "v0.
 evm = { git = "https://github.com/aurora-is-near/sputnikvm.git", tag = "v0.36.0-aurora", default-features = false, features = ["std", "tracing"] }
 evm-runtime = { git = "https://github.com/aurora-is-near/sputnikvm.git", tag = "v0.36.0-aurora", default-features = false, features = ["std", "tracing"] }
 evm-gasometer = { git = "https://github.com/aurora-is-near/sputnikvm.git", tag = "v0.36.0-aurora", default-features = false, features = ["std", "tracing"] }
+hex = { version = "0.4", default-features = false, features = ["alloc"] }
```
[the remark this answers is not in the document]
joshuajbouw: Sure: https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-standalone-tracing/src/types/call_tracer.rs#L283-L284

This uses hex, but it's not in there as a dependency.

.env/testnet.env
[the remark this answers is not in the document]
joshuajbouw: Fixed here: https://github.com/aurora-is-near/aurora-engine/pull/561/commits/dc5ce3658b12cdf47c52c7583d9774e90354df6b

engine-standalone-tracing/Cargo.toml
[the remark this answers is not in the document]
joshuajbouw: Slight change here: https://github.com/aurora-is-near/aurora-engine/pull/561/commits/ebd0d0ed965237909538621c4f8a2c2ef760ae3c

engine-transactions/src/lib.rs
```diff
@@ -131,40 +134,59 @@ impl NormalizedEthTransaction {
...
+            .ok_or(Error::GasOverflow)?;
+
+        let gas_access_list_storage = config
+            .gas_access_list_storage_key
+            .checked_mul(
+                u64::try_from(
+                    self.access_list
+                        .iter()
+                        .map(|a| a.storage_keys.len())
+                        .sum::<usize>(),
```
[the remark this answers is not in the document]
joshuajbouw: Can't quite easily change `len()` being a size into a `u64` inside of a closure. That and at the time, I was highly doubting that `u32::MAX` would ever be hit with this.

engine/src/json.rs
joshuajbouw: Improved in https://github.com/aurora-is-near/aurora-engine/pull/561/commits/926634beb14066bdfef0e0f7342cb3b5fc2bfcb7

engine-standalone-tracing/Cargo.toml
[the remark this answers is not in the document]
birchmd: Oh, I see. Looks like `hex` was there originally: https://github.com/aurora-is-near/aurora-engine/pull/538/files#diff-f2cd6f1f315135550cb1cf68df11064721e60d1560ea6ed8b86ab0a13f831d57R24 but must have somehow been dropped. Thanks for bringing it back.

birchmd (approved):

PR #562, opened 2022-08-02
User version 4.0.0 of near-sdk-rs

mfornet:
The recent version of near-sdk addresses makes the Router Contract cleaner IMO.

### DIFF
--- a/engine-tests/Cargo.toml
+++ b/engine-tests/Cargo.toml
@@ -20,7 +20,7 @@ aurora-engine-precompiles = { path = "../engine-precompiles", default-features =
 aurora-engine-transactions = { path = "../engine-transactions", default-features = false, features = ["std"] }
 engine-standalone-storage = { path = "../engine-standalone-storage" }
 engine-standalone-tracing = { path = "../engine-standalone-tracing" }
-borsh = { version = "0.8.2", default-features = false }
+borsh = { version = "0.9", default-features = false }
 sha3 = { version = "0.9.1", default-features = false }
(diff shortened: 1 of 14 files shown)


birchmd (approved):

engine-tests/Cargo.toml
```diff
@@ -35,8 +35,8 @@ ethabi = { git = "https://github.com/darwinia-network/ethabi", branch = "xavier-
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 hex = { version = "0.4.3", default-features = false }
-near-sdk = { git = "https://github.com/aurora-is-near/near-sdk-rs.git", rev = "ba2eddbfbf4484ac3e44b4c8119bbac4907d6e07" }
-near-sdk-sim = { git = "https://github.com/aurora-is-near/near-sdk-rs.git", rev = "ba2eddbfbf4484ac3e44b4c8119bbac4907d6e07" }
+near-sdk = { git = "https://github.com/aurora-is-near/near-sdk-rs.git", rev = "7a3fa3fbff84b712050370d840297df38c925d2d" }
+near-sdk-sim = { git = "https://github.com/aurora-is-near/near-sdk-rs.git", rev = "7a3fa3fbff84b712050370d840297df38c925d2d" }
```
sept-en, line 39: why not just `tag = "4.0.0"`?

birchmd: It's not a real release. And we want to get off of `near-sdk-sim` because it is not supported by Near any more. There is an ongoing effort in the engine team to upgrade to `workspaces-rs`, which has superceded `near-sdk-sim`.

PR #564, opened 2022-08-04
Feat(xcc): Estimate EVM gas cost for xcc precompile

birchmd:
Addresses https://github.com/aurora-is-near/aurora-engine/pull/560#discussion_r936952965

### DIFF
--- a/engine-precompiles/src/xcc.rs
+++ b/engine-precompiles/src/xcc.rs
@@ -20,12 +20,25 @@ pub mod costs {
     use crate::prelude::types::{EthGas, NearGas};
 
     /// Base EVM gas cost for calling this precompile.
+    /// Value obtained from the following methodology:
+    /// 1. Estimate the cost of calling this precompile in terms of NEAR gas.
+    ///    This is done by calling the precompile with inputs of different lengths
+    ///    and performing a linear regression to obtain a function
+    ///    `NEAR_gas = CROSS_CONTRACT_CALL_BASE + (input_length) * (CROSS_CONTRACT_CALL_BYTE)`.
+    /// 2. Convert the NEAR gas cost into an EVM gas cost using the conversion ratio below
+    ///    (`CROSS_CONTRACT_CALL_NEAR_GAS`).
+    ///
+    /// This process is done in the `test_xcc_eth_gas_cost` test in
+    /// `engine-tests/src/tests/xcc.rs`.
     pub const CROSS_CONTRACT_CALL_BASE: EthGas = EthGas::new(115_000);
     /// Additional EVM gas cost per bytes of input given.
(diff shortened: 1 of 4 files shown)


engine-precompiles/src/xcc.rs
```diff
@@ -19,8 +19,14 @@ use evm_core::ExitError;
...
     use crate::prelude::types::{EthGas, NearGas};
 
-    // TODO(#483): Determine the correct amount of gas
-    pub(super) const CROSS_CONTRACT_CALL: EthGas = EthGas::new(0);
+    /// Base EVM gas cost for calling this precompile.
+    pub const CROSS_CONTRACT_CALL_BASE: EthGas = EthGas::new(115_000);
+    /// Additional EVM gas cost per bytes of input given.
+    pub const CROSS_CONTRACT_CALL_BYTE: EthGas = EthGas::new(2);
+    /// EVM gas cost per NEAR gas attached to the created promise.
+    /// Derived from the gas report https://hackmd.io/@birchmd/Sy4piXQ29
```
joshuajbouw, line 27: It would be helpful to give quite a bit of context of how the value of 175,000,000 was derived, and mention that it was rounded to keep it consistent.

Same goes for the 115_000 value above.

joshuajbouw (changes requested):
This is quite a bit more clever than I had anticipated.

Just needs a bit more documentation, and I'm pizza.

engine-precompiles/src/xcc.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/564/commits/12f123862d704aa12336e46227918738c854487d

joshuajbouw (approved):

engine-types/src/lib.rs
```diff
@@ -24,8 +24,8 @@ mod v0 {
         vec::Vec,
     };
     pub use core::{
-        cmp::Ordering, fmt::Display, marker::PhantomData, mem, ops::Add, ops::Div, ops::Mul,
-        ops::Sub, ops::SubAssign,
+        cmp::Ordering, fmt::Display, marker::PhantomData, mem, ops::Add, ops::AddAssign, ops::Div,
+        ops::Mul, ops::Sub, ops::SubAssign,
```
mfornet, line 28: Is this used?

mfornet (approved):
Interesting that the cost per byte (2) is almost negligible with respect to the cost base (115k).

engine-types/src/lib.rs
birchmd: The diff looks a little strange because of the `cargo fmt` wrapping. But the important change is the addition of `ops::AddAssign`. That trait is now implemented by `EthGas` so that I can use `+=` on the cost of the xcc precompile.

PR #565, opened 2022-08-05
Chore(engine): Update to latest version of SputnikVM

birchmd:
Updates the engine to use the latest version of SputnikVM. This change is important because it adds new functionality to precompiles to allow them to make sub-calls. This will be useful in the upcoming cross-contract calls (xcc) feature.

In this PR we do not change our internal `Precompile` trait to expose the new functionality, this can be done as part of the xcc work. For now we just update all the dependencies so that we have access to this functionality within SputnikVM itself.

Since SputnikVM uses newer versions of the Ethereum types crates we also had to update a few other dependencies, including a welcome return to the official `ethabi` crate instead of a fork because the official one now has no-std support.

These updates also lead to a small performance improvement, as seen by the slightly lower gas costs in the repro tests.

### DIFF
--- a/engine-precompiles/src/lib.rs
+++ b/engine-precompiles/src/lib.rs
@@ -32,7 +32,7 @@ use aurora_engine_sdk::env::Env;
 use aurora_engine_sdk::io::IO;
 use aurora_engine_types::{account_id::AccountId, types::Address, vec, BTreeMap, Box};
 use evm::backend::Log;
-use evm::executor;
+use evm::executor::{self, stack::PrecompileHandle};
 use evm::{Context, ExitError, ExitSucceed};
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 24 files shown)


engine-precompiles/src/lib.rs
```diff
@@ -118,15 +107,29 @@ pub struct Precompiles<'a, I, E> {
 impl<'a, I: IO + Copy, E: Env> executor::stack::PrecompileSet for Precompiles<'a, I, E> {
     fn execute(
         &self,
-        address: prelude::H160,
-        input: &[u8],
-        gas_limit: Option<u64>,
-        context: &Context,
-        is_static: bool,
+        handle: &mut impl PrecompileHandle,
     ) -> Option<Result<executor::stack::PrecompileOutput, executor::stack::PrecompileFailure>> {
```
joshuajbouw, line 111: Just a nit, but could also import executor::stack::PrecompileFailure as well, and others that may need to be imported that are types.

joshuajbouw (approved):
I got a small nit you can choose to do or not. Generally, looks good and the gas improvements are more than welcomed.

PR #571, opened 2022-08-09
Chore: Cleanup dead dependencies

joshuajbouw:
## Description

While working on the libraries, the project has had many changes. This has resulted in many dependencies not being removed when they should've been. This PR simply removes all those dependencies flagged using the [udeps](https://github.com/est31/cargo-udeps) cargo tool.

Further, a new CI addition was added to ensure that there are no more dead dependencies in the future.

None.

## Testing

None.

It is recommended to be familiar with the [udeps](https://github.com/est31/cargo-udeps) cargo tool.

## Additional information

Command ran:
`cargo +nightly udeps --all-targets`


engine/src/lib.rs
```diff
@@ -13,8 +13,6 @@ extern crate alloc;
 extern crate core;
 
 mod map;
-#[cfg(feature = "meta-call")]
-pub mod meta_parsing;
```
birchmd, line 17: Can we delete all of this? We're never going to use it for anything?

birchmd (changes requested):

sept-en (commented):
This is indeed a good sanity! One comment, should we really remove `meta_call` functionality completely?

mrLSD (approved):

engine/src/lib.rs
joshuajbouw: Weird. I thought that I had caught all of these in my review. Must've not searched the entire library by mistake.

joshuajbouw:
> This is indeed a good sanity! One comment, should we really remove `meta_call` functionality completely?

Yes, we do not need it at all. Even then, it still "exists" in the history of things if we need to bring it back from the dead.

engine/src/lib.rs
joshuajbouw: We had discussed this before, where we had decided that it shouldn't be the engine that handles meta calls, but it should be the relayer itself, if I'm not mistaken.

engine/Cargo.toml
```diff
@@ -55,4 +48,4 @@ error_refund = ["aurora-engine-precompiles/error_refund"]
 integration-test = ["log"]
 mainnet = ["contract", "log", "aurora-engine-sdk/mainnet"]
 testnet = ["contract", "log", "aurora-engine-sdk/testnet"]
-impl-serde = ["aurora-engine-types/impl-serde", "serde", "aurora-engine-transactions/impl-serde", "evm-core/with-serde"]
```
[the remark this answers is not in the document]
joshuajbouw: `evm-core` hadn't been used in engine/Cargo.toml though. Was marked as dead.

birchmd, line 58: I promise you we cannot remove this `evm-core/with-serde` feature. Try it yourself, if you compile with `impl-serde` it will fail

```
$ cargo check -p aurora-engine --features impl-serde

error[E0277]: the trait bound `ExitError: Serialize` is not satisfied
   --> engine/src/engine.rs:93:14
    |
93  |     EvmError(ExitError),
    |              ^^^^^^^^^ the trait `Serialize` is not implemented for `ExitError`
    |
note: required by a bound in `serialize_newtype_variant`
   --> /home/birchmd/.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.137/src/ser/mod.rs:940:12
    |
940 |         T: Serialize;
    |            ^^^^^^^^^ required by this bound in `serialize_newtype_variant`

error[E0277]: the trait bound `ExitFatal: Serialize` is not satisfied
   --> engine/src/engine.rs:95:14
    |
95  |     EvmFatal(ExitFatal),
    |              ^^^^^^^^^ the trait `Serialize` is not implemented for `ExitFatal`
    |
note: required by a bound in `serialize_newtype_variant`
   --> /home/birchmd/.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.137/src/ser/mod.rs:940:12
    |
940 |         T: Serialize;
    |            ^^^^^^^^^ required by this bound in `serialize_newtype_variant`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `aurora-engine` due to 2 previous errors
```

birchmd (changes requested):

engine/Cargo.toml
joshuajbouw: Fixed with 2bc270a668258512ffafaecc924285649a764d16 by using `evm/with-serde` instead.

birchmd (approved):

PR #572, opened 2022-08-10
Feat(xcc): User payment using wNEAR ERC-20

birchmd:
See https://github.com/aurora-is-near/aurora-engine/pull/560#discussion_r936082710 for context.


joshuajbouw:
Yeah, this makes sense to me to charge in WNEAR since these are taking place on NEAR.

engine-precompiles/src/xcc.rs
```diff
@@ -52,6 +53,7 @@ mod consts {
     pub(super) const ERR_DELEGATE: &str = "ERR_INVALID_IN_DELEGATE";
     pub(super) const ROUTER_EXEC_NAME: &str = "execute";
     pub(super) const ROUTER_SCHEDULE_NAME: &str = "schedule";
+    pub(super) const TRANSFER_FROM_SELECTOR: [u8; 4] = [0x23, 0xb8, 0x72, 0xdd];
```
joshuajbouw, line 56: This needs an explanation of what this is.

engine-precompiles/src/xcc.rs
```diff
@@ -160,16 +223,102 @@ impl<I: IO> Precompile for CrossContractCall<I> {
...
         .parse()
         .unwrap()
 }
 
+fn revert_with_message(message: &str) -> PrecompileFailure {
+    PrecompileFailure::Revert {
+        exit_status: evm::ExitRevert::Reverted,
+        output: message.as_bytes().to_vec(),
+    }
+}
```
joshuajbouw, line 314: We should use this for other precompiles. Possibly as a macro, with a default message if no message is set.

engine/src/lib.rs
```diff
@@ -342,6 +342,17 @@ mod contract {
         crate::xcc::set_code_version_of_address(&mut io, &args.address, args.version);
     }
 
+    /// Sets the address for the wNEAR ERC-20 contract. This contract will be used by the
+    /// cross-contract calls feature to have users pay for their NEAR transactions.
+    #[no_mangle]
+    pub extern "C" fn factory_set_wnear_address() {
```
joshuajbouw, line 348: I take it that we would have access to this in order to get the wNEAR back as compensation?

etc/xcc-router/src/lib.rs
```diff
@@ -33,12 +39,14 @@ pub struct Router {
     nonce: LazyOption<u64>,
     /// The storage for the scheduled promises.
     scheduled_promises: LookupMap<u64, PromiseArgs>,
+    /// Account ID for the wNEAR contract.
+    wnear_account: AccountId,
 }
 
 #[near_bindgen]
```
joshuajbouw, line 46: Wouldn't using `near_sdk` require us to use NEAR's preferred license...?

joshuajbouw (changes requested):
Just some things that I noticed. This is going to take me a week to go through the whole thing as a whole and try to poke holes into it I'm noticing :D.

engine-precompiles/src/xcc.rs
```diff
@@ -107,46 +115,101 @@ impl<I: IO> Precompile for CrossContractCall<I> {
...
+            }
+            CrossContractCallArgs::Delayed(call) => {
+                let attached_near = call.total_near();
+                let promise = PromiseCreateArgs {
+                    target_account_id,
+                    method: consts::ROUTER_SCHEDULE_NAME.into(),
+                    args: call
+                        .try_to_vec()
+                        .map_err(|_| ExitError::Other(Cow::from(consts::ERR_SERIALIZE)))?,
+                    attached_balance: ZERO_YOCTO,
```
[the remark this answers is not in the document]
birchmd: It is possible for users to attach NEAR to their promises. They specify that in the input to the precompile. The engine itself does not ever attach any NEAR though because the users must pay for their own transactions from their router account. They can fund their router account using wNEAR inside the EVM (the purpose of this PR).

engine-precompiles/src/xcc.rs
```diff
@@ -107,46 +115,101 @@ impl<I: IO> Precompile for CrossContractCall<I> {
...
         check_cost(cost)?;
 
+        let required_near =
+            match state::get_code_version_of_address(&self.io, &Address::new(sender)) {
+                // If there is no deployed version of the router contract then we need to charge for storage staking
+                None => attached_near + state::STORAGE_AMOUNT,
+                Some(_) => attached_near,
+            };
+        // if some NEAR payment is needed, transfer it from the caller to the engine's implicit address
+        if required_near != ZERO_YOCTO {
```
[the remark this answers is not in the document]
birchmd: `STORAGE_AMOUNT` is intentionally over estimated to allow it to cover both the storage staking and a little execution. The tests `test_xcc_precompile_eager` and `test_xcc_precompile_scheduled` both cover the situation you describe and the user's promise is executed successfully.

engine-precompiles/src/xcc.rs
birchmd: I think this should be out of scope for this PR because it is large enough as is. Please file an issue and the engine team can take on that refactor as a follow up PR after the xcc feature is merged.

engine/src/lib.rs
birchmd: Not exactly. The reason the engine needs to know this address is to transfer the wNEAR the user approved. The flow is as follows: (1) user approves wNEAR ERC-20 to be spent by the precompile, (2) the precompile transfers the required amount of wNEAR ERC-20 to the engine's implicit address, (3) when creating the xcc promises, the engine includes a `call` to itself in which it uses the exit precompile to send that wNEAR to the user's router account, (4) additionally, a promise is included where the router unwraps its wNEAR into actual NEAR to cover its costs.

Note that the engine never had to use any of its own funds at any point in this flow; it is only facilitating the user in spending their own wNEAR. The exception is when the router account is created for the first time. In this case the engine must loan some NEAR to create the router account before it can successfully use the exit precompile to send the user's own wNEAR. In this case the router refunds the engine the amount it loaned.

Also note that if step (2) fails then the user did not approve enough wNEAR to cover their costs and the flow will terminate (revert) there with an error.

etc/xcc-router/src/lib.rs
birchmd: It's a little unclear. Technically, the version of the sdk we are using has a GPL-3 license, but since then they have weakened it to be [MIT/Apache](https://github.com/near/near-sdk-rs/blob/master/near-sdk/Cargo.toml#L6), which I think is fine for our purposes. So we could change our dependency to a later commit (after their last release) to get to the point where they made the license change.

engine-precompiles/src/lib.rs
```diff
@@ -72,6 +72,12 @@ pub trait Precompile {
         is_static: bool,
     ) -> EvmPrecompileResult;
 }
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/572/commits/29953b96dfb0920f4fe9a251e6ec600aed6a3d97

engine-precompiles/src/lib.rs
```diff
@@ -114,6 +120,25 @@ impl<'a, I: IO + Copy, E: Env> executor::stack::PrecompileSet for Precompiles<'a
...
+                    Ok(executor::stack::PrecompileOutput {
+                        exit_status: ExitSucceed::Returned,
+                        output: output.output,
+                    })
+                });
+
+            return Some(result);
+        }
+
         self.precompile_action(Address::new(address), |p| {
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/572/commits/92a596514221b20463237445a32906c42d5c5ee8

etc/xcc-router/src/lib.rs
joshuajbouw: Oh, great. If they have that clause, that makes me happy.

engine-precompiles/src/xcc.rs
joshuajbouw: Do make note in the description that a follow up PR should follow something like that, then on some clean up cycle we can make it the same.

engine-precompiles/src/xcc.rs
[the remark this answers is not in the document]
joshuajbouw: Makes perfect sense.

joshuajbouw (approved):

engine-precompiles/src/xcc.rs
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/572/commits/07e0afc5548a637c48d763fb05dcc49c84b033a8

engine-precompiles/src/xcc.rs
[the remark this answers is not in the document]
joshuajbouw: Makes sense.

engine/src/lib.rs
joshuajbouw: Makes sense.

engine/src/engine.rs
```diff
@@ -1386,8 +1386,18 @@ where
                     Some(log.into())
                 }
             } else if log.address == cross_contract_call::ADDRESS.raw() {
-                if let Ok(promise) = PromiseCreateArgs::try_from_slice(&log.data) {
-                    crate::xcc::handle_precompile_promise(io, handler, promise, current_account_id);
+                if log.topics[0] == cross_contract_call::AMOUNT_TOPIC {
+                    let required_near =
+                        Yocto::new(U256::from_big_endian(log.topics[1].as_bytes()).low_u128());
```
mfornet, line 1391: Using `low_u128` here looks like a red flag. Similar to down-casting using `as`!

Should we check that higher bytes are `0`?

Anyway, from the calling site, I see that higher bytes will always be zero, so not sure this is worth it, but just highlighting it here :)

mfornet (approved):
👍

engine/src/engine.rs
birchmd: Added a check in https://github.com/aurora-is-near/aurora-engine/pull/572/commits/e51aec8fdd4d601e5f7829e0fad50827eb78863f

PR #575, opened 2022-08-12
Feat(engine): Get promise results precompile

birchmd:
With the upcoming cross-contract call feature, it will be important to be able to get the results of past calls (promises) within the EVM.


engine-standalone-storage/src/sync/types.rs
```diff
@@ -126,9 +140,31 @@ struct BorshableTransactionMessage<'a> {
...
+    pub succeeded: bool,
+    /// NEAR account that signed the transaction
+    pub signer: Cow<'a, AccountId>,
+    /// NEAR account that called the Aurora engine contract
+    pub caller: Cow<'a, AccountId>,
+    /// Amount of NEAR token attached to the transaction
+    pub attached_near: u128,
+    /// Details of the transaction that was executed
+    pub transaction: BorshableTransactionKind<'a>,
+    pub promise_data: Cow<'a, Vec<Option<Vec<u8>>>>,
```
mfornet, line 162: Should we copy-paste the same comment on this field on `TransactionMessage`? Or maybe removing all other comments and adding a meta-comment pointing to `TransactionMessage`?

engine-standalone-storage/src/sync/types.rs
```diff
@@ -106,7 +114,13 @@ pub enum TransactionKind {
 }
 
 #[derive(BorshDeserialize, BorshSerialize)]
-struct BorshableTransactionMessage<'a> {
+enum BorshableTransactionMessage<'a> {
+    V1(BorshableTransactionMessageV1<'a>),
+    V2(BorshableTransactionMessageV2<'a>),
+}
```
mfornet, line 120: FYI, this is not compatible with BorshableTransactionMessageV1. If you try to deserialize a legacy tx it will fail, because it is expecting the enum type on the first byte.

An alternative solution is making a custom implementation of Borsh(De)Serialize.

mfornet: If the plan is to drop support to legacy types, then we can drop the enum completely!

engine-precompiles/src/promise_result.rs
```diff
@@ -0,0 +1,90 @@
...
+use aurora_engine_sdk::promise::ReadOnlyPromiseHandler;
+use aurora_engine_types::{Cow, Vec};
+use borsh::BorshSerialize;
+use evm::{Context, ExitError};
+
+/// predecessor_account_id precompile address
+///
+/// Address: `0x0a3540f79be10ef14890e87c1a0040a68cc6af71`
+/// This address is computed as: `&keccak("prepaidGas")[12..]`
+pub const ADDRESS: Address = crate::make_address(0x0a3540f7, 0x9be10ef14890e87c1a0040a68cc6af71);
```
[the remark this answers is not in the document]
mfornet: It is very funny, though: it says `predecessor_account_id,` but the hash function mention `prepaidGas`, and it is intended to say in both places `promiseResult`? 😂 

This problem exists as well on [prepaid gas precompile](https://github.com/aurora-is-near/aurora-engine/blob/master/engine-precompiles/src/prepaid_gas.rs#L8).

mfornet (approved):
LGTM. Left few comments

engine-standalone-storage/src/sync/types.rs
birchmd: The custom deserialization logic is [here](https://github.com/aurora-is-near/aurora-engine/pull/575/files#diff-ad2bf34db0628958ff395ae9b757d564469163a314630a681e8bbb5be2a1b5b7R53). Notice we first try to deserialize the enum, then if it fails we fall back on the V1 serialization. Also note that `BorshableTransactionMessage` is private, so we don't need to worry about others not knowing about this special logic.

I decided to do it this way because I wanted to avoid having to do a DB migration on the engine (and now also refiner), and I did not think it would be very common to look at transaction data, so the slightly inefficient deserialization logic should not be a big performance problem (transaction data is only read from the DB to do debug tracing, which is not a very frequently used endpoint on average).

But let me know if you think it is worth doing the DB migration instead of (or perhaps in addition to) having the custom deserialization logic.

engine-precompiles/src/promise_result.rs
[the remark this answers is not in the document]
birchmd: Done in https://github.com/aurora-is-near/aurora-engine/pull/575/commits/8c6f9f09865a659061ab58542a9af9edcc6570a9

engine-standalone-storage/src/sync/types.rs
birchmd: Removed field comments and made one comment on the `BorshableTransactionMessage` itself in https://github.com/aurora-is-near/aurora-engine/pull/575/commits/43874feb9c8e553e064a400da0edbfcecc39cea2

engine-standalone-storage/src/sync/types.rs
```diff
@@ -105,30 +113,55 @@ pub enum TransactionKind {
...
     pub position: u16,
-    /// True if the transaction executed successfully on the blockchain, false otherwise.
     pub succeeded: bool,
-    /// NEAR account that signed the transaction
     pub signer: Cow<'a, AccountId>,
-    /// NEAR account that called the Aurora engine contract
     pub caller: Cow<'a, AccountId>,
-    /// Amount of NEAR token attached to the transaction
     pub attached_near: u128,
-    /// Details of the transaction that was executed
```
joshuajbouw, line 125: Why were the doc comments dropped here?

engine-types/src/lib.rs
```diff
@@ -25,8 +25,8 @@ mod v0 {
         vec::Vec,
     };
     pub use core::{
-        cmp::Ordering, fmt::Display, marker::PhantomData, mem, ops::Add, ops::Div, ops::Mul,
-        ops::Sub, ops::SubAssign,
+        cmp::Ordering, fmt::Display, marker::PhantomData, mem, ops::Add, ops::AddAssign, ops::Div,
+        ops::Mul, ops::Sub, ops::SubAssign,
```
joshuajbouw, line 29: Seems a bit odd to me that this had changed when none of the logic below had. Just commenting on that.

joshuajbouw (approved):
Just two comments. Nice work. I like the split to read-only.

engine-standalone-storage/src/sync/types.rs
birchmd: See https://github.com/aurora-is-near/aurora-engine/pull/575#discussion_r945710183

engine-types/src/lib.rs
birchmd: The diff looks more complicated than it is because of the formatting causing wrapping. All that happened here is `ops::AddAssign` was added to the imports and then it is used to implement `AddAssign` for the `EthGas` type.

engine-standalone-storage/src/sync/types.rs
mfornet: I missed that fallback part! It seems ok to me this way.

PR #581, opened 2022-08-17
Fix(engine): Return correct value for get_bridge_prover function

birchmd:
The intent of the `get_bridge_prover` function is to return the account ID the engine uses to verify proofs coming from the bridge (used in transferring assets from Ethereum to NEAR). The account ID is stored in the eth-connector. In this PR, we fix the `get_bridge_prover` function to return the value from the eth-connector. I think this means the value in the engine state is unused, but it would be a bit of a pain to remove it (requires state migration to delete a field), so it is probably not worth the hassle.

### DIFF
--- a/engine/src/connector.rs
+++ b/engine/src/connector.rs
@@ -619,6 +619,10 @@ impl<I: IO + Copy> EthConnectorContract<I> {
             .return_output(&self.ft.get_accounts_counter().to_le_bytes());
     }
 
+    pub fn get_bridge_prover(&self) -> &AccountId {
+        &self.contract.prover_account
+    }
+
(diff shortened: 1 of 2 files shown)


pustovalov (approved):

sept-en (approved):
IMO as we store the variable in the storage, it makes sense to update it in the engine during `set_contract_data` call too.

birchmd:
Could you clarify what you mean @sept-en ? The function `set_eth_connector_contract_data` does take the prover account ID as an argument. Are you suggesting this function also change the engine state, not just the connector state? I think that would be redundant; we do not need to store the exact same information in two places.

joshuajbouw (approved):

PR #588, opened 2022-08-19
Pausable precompiles

RomanHodulak:
## Description

Introduces pausable precompiles for stopping the engine when it hits a precompiled function. This is useful for emergency cases.

Increases gas cost by a few Tgas because of additional storage read operation to load the few bytes of paused precompiles flags.

## Testing

Every added module is covered by a set of unit tests covering every conceived test scenario.

Then there is a module in `engine-tests` containing a set of functional tests that checks several test scenarios on the wasm binary. For example, there is a test case hitting a paused precompile during a smart contract execution that expects the error `ERR_PAUSED`. 

You should consider the administrative engine smart contract methods this feature is going to be controlled by. Take a look at test scenarios to get an idea. Separately from that, there is the mechanism of ensuring calling a paused precompile throws an error, which you should observe.

## Additional information

I have some thoughts considering the use-cases/usefulness of the engine smart contract API of this feature:

### Engine smart contract method interface
The current interface is like this:

As an admin I want to pause `modexp`, its associated flag is `0b000000000010000`. To pause it I need to run:

`pause_precompiles(0b000000000010000)`

The cause of the issue that made me pause `modexp` is resolved, now I run:

`resume_precompiles(0b000000000010000)`

Seems pretty inconvenient to need to know the right binary flag. Maybe it would be better to accept a list of stringified versions? So the previous example would run like this:

`pause_precompiles(["MODEXP"])`

`resume_precompiles(["MODEXP"])`

Also, when you hit invalid string, the error message could include a list of all possible precompile strings.

`pause_precompiles(["MODXP"])`

`Invalid precompile "MODXP", pick one or more from this list: "MODEXP", "BLAKE2F", ...`

### Reading the current state from the engine smart contract

The `paused_precompiles` method returns u32, where each 1-bit identifies a paused precompile. Same as above, more human friendly would be to return a list:

`paused_precompiles() -> Vec<String>`

For example, outputs:

`["MODEXP", "BLAKE2F", ...]`

### DIFF
--- a/engine-precompiles/src/alt_bn256.rs
+++ b/engine-precompiles/src/alt_bn256.rs
@@ -167,10 +167,11 @@ fn read_point(input: &[u8], pos: usize) -> Result<bn::G1, ExitError> {
     })
 }
 
-pub(super) struct Bn256Add<HF: HardFork>(PhantomData<HF>);
+#[derive(Default)]
+pub struct Bn256Add<HF: HardFork>(PhantomData<HF>);
 
 impl<HF: HardFork> Bn256Add<HF> {
-    pub(super) const ADDRESS: Address = super::make_address(0, 6);
+    pub const ADDRESS: Address = super::make_address(0, 6);
 
     pub fn new() -> Self {
         Self(Default::default())
@@ -269,10 +270,11 @@ impl Precompile for Bn256Add<Istanbul> {
     }
 }
 
-pub(super) struct Bn256Mul<HF: HardFork>(PhantomData<HF>);
+#[derive(Default)]
+pub struct Bn256Mul<HF: HardFork>(PhantomData<HF>);
 
 impl<HF: HardFork> Bn256Mul<HF> {
-    pub(super) const ADDRESS: Address = super::make_address(0, 7);
+    pub const ADDRESS: Address = super::make_address(0, 7);
 
     pub fn new() -> Self {
         Self(Default::default())
@@ -374,10 +376,11 @@ impl Precompile for Bn256Mul<Istanbul> {
     }
 }
 
-pub(super) struct Bn256Pair<HF: HardFork>(PhantomData<HF>);
+#[derive(Default)]
+pub struct Bn256Pair<HF: HardFork>(PhantomData<HF>);
 
 impl<HF: HardFork> Bn256Pair<HF> {
-    pub(super) const ADDRESS: Address = super::make_address(0, 8);
+    pub const ADDRESS: Address = super::make_address(0, 8);
 
     pub fn new() -> Self {
         Self(Default::default())
--- a/engine-tests/src/tests/pausable_precompiles.rs
+++ b/engine-tests/src/tests/pausable_precompiles.rs
@@ -1,0 +1,160 @@
+use crate::prelude::{Address, U256};
+use crate::test_utils::exit_precompile::{Tester, TesterConstructor};
+use crate::test_utils::{
+    self, origin, AuroraRunner, Signer, PAUSED_PRECOMPILES, PAUSE_PRECOMPILES, RESUME_PRECOMPILES,
+};
+use aurora_engine::parameters::{PausePrecompilesCallArgs, TransactionStatus};
+use aurora_engine_types::types::Wei;
+use borsh::BorshSerialize;
+use near_vm_errors::{FunctionCallError, HostError};
+use near_vm_runner::VMError;
+
+const EXIT_TO_ETHEREUM_FLAG: u32 = 0b10;
+const CALLED_ACCOUNT_ID: &str = "aurora";
+
+#[test]
+fn test_paused_precompile_is_shown_when_viewing() {
+    let mut runner = test_utils::deploy_evm();
+
+    let call_args = PausePrecompilesCallArgs {
+        paused_mask: EXIT_TO_ETHEREUM_FLAG,
+    };
+
+    let mut input: Vec<u8> = Vec::new();
+    call_args.serialize(&mut input).unwrap();
+
+    let _ = runner.call(PAUSE_PRECOMPILES, CALLED_ACCOUNT_ID, input.clone());
+    let (outcome, error) = runner.call(PAUSED_PRECOMPILES, CALLED_ACCOUNT_ID, Vec::new());
+
+    assert!(error.is_none(), "{:?}", error);
+
+    let output: Vec<u8> = outcome
+        .as_ref()
+        .unwrap()
+        .return_data
+        .clone()
+        .as_value()
+        .unwrap();
+
+    let actual_paused_precompiles = u32::from_le_bytes(output.as_slice().try_into().unwrap());
+    let expected_paused_precompiles = EXIT_TO_ETHEREUM_FLAG;
+
+    assert_eq!(expected_paused_precompiles, actual_paused_precompiles);
+}
+
+#[test]
+fn test_executing_paused_precompile_throws_error() {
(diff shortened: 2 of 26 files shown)


joshuajbouw:
Make sure CI is working.

engine/src/lib.rs
```diff
@@ -196,6 +199,40 @@ mod contract {
...
+        let mut io = Runtime;
+        let state = engine::get_state(&io).sdk_unwrap();
+        let predecessor_account_id = io.predecessor_account_id();
+
+        require_owner_only(&state, &predecessor_account_id);
+
+        let args: PausePrecompilesCallArgs = io.read_input_borsh().sdk_unwrap();
+        let mut pauser: pausables::EnginePrecompilesPauser = engine::get_pauser(&io).sdk_unwrap();
+        pauser.resume_precompiles(args.paused_mask);
+        engine::set_pauser(&mut io, pauser);
```
joshuajbouw, line 216: This doesn't work exactly as the logic dictates. It's the same logic as below, the only difference being that ONLY the owner can do this. This is a `set_flags` sort of method. Best would only to combine the two methods, call it `set_paused_precompiles` which uses the `!authorizer.is_authorized(&predecessor_account_id)` logic below.

joshuajbouw (changes requested):

engine-tests/src/tests/pausable_precompiles.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    let (outcome, error) = runner.call(RESUME_PRECOMPILES, called_account_id, input);
+
+    assert!(outcome.is_some(), "{:?}", outcome);
+    assert!(error.is_none(), "{:?}", error);
+}
+
+/// The ecrecover implementation in both the standalone and wasm contract should be the same.
+#[test]
+fn test_pause_precompiles() {
```
birchmd, line 24: It would be nice to test more than just "it didn't have an error during execution". I think the test should go something like 

1. Call pause precompiles
2. Try to execute a precompile -> it fails because they're paused
3. Call resume precompiles
4. Try to execute a precompile again -> it works

birchmd: Thanks for adding tests.

Yes, please increment the gas bounds of the failed tests. The purpose of these tests is to track how our gas usage changes over time. They fail for both increases and decreases to the gas cost. It allows us to make informed decisions about features like this. We need to answer the question of whether the feature we are adding is worth the additional cost in the case of increases. In this case I think it is clear that the pausing functionality it worth a few Tgas.

I think it's ok to share the standard_precompiles contract between tests because I think it is unlikely to change much over time.

engine-tests/src/tests/pausable_precompiles.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    let _ = runner.call(PAUSE_PRECOMPILES, CALLED_ACCOUNT_ID, input.clone());
+    let (result, _) = runner
+        .submit_with_signer_profiled(&mut signer, |nonce| {
+            contract.call_method(TEST_MODEXP, nonce)
+        })
+        .unwrap();
+
+    let number = match result.status {
+        TransactionStatus::Succeed(number) => U256::from(number.as_slice()),
```
birchmd, line 31: I am confused about the outcome of this test. I thought after pausing trying to use the precompile should throw an error.

mrLSD:
@RomanHodulak it looks like Gast cost increased?

engine/src/pausables.rs
```diff
@@ -0,0 +1,229 @@
...
+        const EXIT_TO_ETHEREUM = 0b1000000000000;
+        const EXIT_TO_NEAR = 0b10000000000000;
+        const PREPAID_GAS = 0b100000000000000;
+    }
+}
+
+impl PrecompileFlags {
+    /// Checks if the precompile belonging to the `address` is NOT paused.
+    pub fn is_not_paused_by_address(&self, address: &Address) -> bool {
+        let precompile_flag = if address == &ECRecover::ADDRESS {
```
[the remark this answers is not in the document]
joshuajbouw: Thats unfortunate.

joshuajbouw:
Please fix CI @RomanHodulak

mooori:
> As an admin I want to pause `modexp`, its associated flag is `0b000000000010000`. To pause it I need to run:
> 
> `pause_precompiles(0b000000000010000)`
> 
> The cause of the issue that made me pause `modexp` is resolved, now I run:
> 
> `resume_precompiles(0b000000000010000)`
> 
> Seems pretty inconvenient to need to know the right binary flag.

The [bitflags](https://docs.rs/bitflags/latest/bitflags/index.html) crate could be helpful here.

joshuajbouw:
As noted in call, be sure to update https://github.com/aurora-is-near/aurora-cli as well with the new methods.

engine/src/pausables.rs
```diff
@@ -0,0 +1,229 @@
...
+        } else if address == &CurrentAccount::ADDRESS {
+            PrecompileFlags::CURRENT_ACCOUNT
+        } else if address == &predecessor_account::ADDRESS {
+            PrecompileFlags::PREDECESSOR_ACCOUNT
+        } else if address == &exit_to_ethereum::ADDRESS {
+            PrecompileFlags::EXIT_TO_ETHEREUM
+        } else if address == &exit_to_near::ADDRESS {
+            PrecompileFlags::EXIT_TO_NEAR
+        } else if address == &prepaid_gas::ADDRESS {
+            PrecompileFlags::PREPAID_GAS
```
joshuajbouw, line 70: To be fair, most of these should never be paused. Should be only:

`EXIT_TO_ETHEREUM`
`EXIT_TO_NEAR`

(and upcoming cross contract call precompile)

There shouldn't ever be a reason why we would need to pause anything else but those two.

joshuajbouw (changes requested):

joshuajbouw (approved):
This looks like a well introduced feature which will be helpful. Great job!

birchmd (changes requested):
Nice work overall. I think there are some details in the API design that need a second look though.

engine-precompiles/src/alt_bn256.rs
```diff
@@ -167,10 +167,11 @@ fn read_point(input: &[u8], pos: usize) -> Result<bn::G1, ExitError> {
...
-pub(super) struct Bn256Add<HF: HardFork>(PhantomData<HF>);
+#[derive(Default)]
+pub struct Bn256Add<HF: HardFork>(PhantomData<HF>);
 
 impl<HF: HardFork> Bn256Add<HF> {
-    pub(super) const ADDRESS: Address = super::make_address(0, 6);
+    pub const ADDRESS: Address = super::make_address(0, 6);
 
     pub fn new() -> Self {
         Self(Default::default())
```
mrLSD, line 177: Why it's not just:
`Self::default()`?

engine/src/engine.rs
```diff
@@ -883,6 +847,57 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
...
+            Precompiles::new_london(PrecompileConstructorContext {
+                current_account_id,
+                random_seed,
+                io,
+                env,
+                promise_handler: ro_promise_handler,
+            })
+        };
+
+        Self::apply_pause_flags_to_precompiles(precompiles, pause_flags)
```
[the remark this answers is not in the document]
birchmd: Yes, sorry I just have misread it the first time

birchmd (approved):
Looks good! Thanks for addressing my comments.

mrLSD (approved):

PR #594, opened 2022-08-31
Fix(global_allocator): Switch `wee_alloc` for `dlmalloc`

RomanHodulak:


mrLSD:
Please provide a full description of PR>
Why and how is solved?

mrLSD:
Well, this memory leak bug is quite old. And my question is, how much does it affect us?

Also, by the way, `wee_alloc` is a standard allocator in near_sdk.

Also, I would like to understand - what are the advantages and disadvantages of `dlmalloc` compared to `wee_alloc`. Let me remind you that in WebAssembly, memory works on a different principle than under the control of the operating system. This is a completely different model.

Replacing an allocator without research, just based on a post on Reddit - sounds extremely unconvincing. I repeat - we have been with this allocator for 1.5 years. And if we believe that this is really critical for us, then we need to conduct research, comparisons and show the effectiveness of the decision made.

Where is the proof that the new allocator is better? Looks harmless - simply because it looks better.

It may turn out that the total `wee_alloc` is still more efficient. Plus, keep in mind that memory leaks are not of such a critical nature, since we do not have a common stack and application execution state, due to the very specifics of WebAssembly.

joshuajbouw:
Closing given Pagoda's comments: https://github.com/near/near-sdk-rs/issues/910

PR #599, opened 2022-09-05
Add test benchmarking transaction Emufid2pv2UpxrZae4NyowF2N2ZHvYEPq16LsQc7Uoc6

birchmd:
## Description

This PR adds a new test reproducing [a transaction](https://explorer.testnet.near.org/transactions/Emufid2pv2UpxrZae4NyowF2N2ZHvYEPq16LsQc7Uoc6) from testnet. It currently exceeds the 300 Tgas limit. The majority of the gas usage (approx 85%) is in wasm computation, in terms of host functions there is about 8% of the gas usage in bn128 calls, and about 6% in state access (reading, writing, trie touching, etc).

The user says the transaction should be similar to [one on Polygon](https://mumbai.polygonscan.com/tx/0xd9ab182692c74a873f0c444854ed1045edbb32a252b561677042276143a024b7).

The purpose of this test is to allow us to track our progress as we implement new gas optimizations in the engine. Since most of the computation is in the wasm interpretation, there are no optimizations in the immediate future which will help this case, but there is some promising research.


joshuajbouw:
I would imagine bn256 pair is involved. The rest of the precompiles are quite cheap to execute except for that particular one.

PR #601, opened 2022-09-15
[Docs] Eth Connector - extended documentation

mrLSD:
Added extended documentation for Eth Connector logic.


joshuajbouw (approved):
Looks good. I'll adapt this for our doc.aurora.dev too.

joshuajbouw (approved):
We can improve this later, of course, but this should get us over the edge for now.

PR #602, opened 2022-09-19
Parser for ethereum test json on Rust native EVM 

hskang9:
## Description

This PR focuses on bringing json transaction tests from https://github.com/ethereum/tests to modular rust code so that one can test and tune one's own EVM runtime on code without https://github.com/input-output-hk/mantis/blob/develop/ets/retesteth

This is just a rust typing, so It does not consider gas cost for now.

## Testing

Tests should generally check whether each test typings include `general information about test`, `results with gas amount` and `txbytes` to reproduce transaction

Run each test with aurora vm and check whether results match what the test json file describes.

## Additional information

Feeding to EVM is benchmarking the code [here](https://github.com/rust-blockchain/evm-tests/blob/4f7e566217e3b781be63db7704c84d93ae999be4/jsontests/tests/vm.rs#L1)


joshuajbouw (changes requested):
We can not copy and paste the JSON tests. You need to add it into the `etc` folder as a submodule using the git submodule command.

```
mkdir engine-tests/etc/eth-json-test
git submodule add <source.git> engine-tests/etc/eth-json-test
```

Something along those lines above.

joshuajbouw:
Why does this include Releases? Was it built off a branch on master?

hskang9:
> Why does this include Releases? Was it built off a branch on master?
> 
> Commits must be signed with a PGP key linked to your aurora.dev e-mail tool.

Yes, It was built off a branch from master. I will make a new branch out of develop and send a new PR.

PGP key will be set. I forgot I did not setup the PGP key in my desktop.

hskang9:
> We can not copy and paste the JSON tests. You need to add it into the `etc` folder as a submodule using the git submodule command.
> 
> ```
> mkdir engine-tests/etc/eth-json-test
> git submodule add <source.git> engine-tests/etc/eth-json-test
> ```
> 
> Something along those lines above.

I agree.

hskang9:
I am moving PR to #611 with GPG signed commits with submodules.

PR #609, opened 2022-09-22
chore: get paused precompiles name

joshuajbouw:
## Description

Simply renames `paused_precompiles` to `get_paused_precompiles` to fit in line with the rest of our view naming conventions.

A consideration is that this is only breaking for those using the latest commit of the `develop` branch, but not a breaking change for the deployed binary. With that in mind, it is safe to change it now.

None.

## Testing

A test was updated to use `get_paused_precompiles` instead.

The CI should catch any issues, no special requirements needed to review.

## Additional information

None.


birchmd:
Why didn't CI run on this PR?

mrLSD (approved):

sept-en (approved):
I like this naming convention much more. @joshuajbouw, however, IMO, the even better naming would be `get_paused_precompiles_flags()` because it actually returns flags. But the current naming falsely supposes that the function returns a list of paused precompiles (names, indices, whatever).

@RomanHodulak NB from the same guidelines: `The get naming is used only when there is a single and obvious thing that could reasonably be gotten by a getter. `. I think this is the cause.

joshuajbouw:
@sept-en Yes, I typically use "get" if there actually is a storage get myself. It describes that it actually is getting something and costing some computational resources opposed to grabbing something stored in the type itself.

aleksuss (approved):

PR #611, opened 2022-09-27
Parser for ethereum test json on Rust native EVM

hskang9:
## Description

This PR focuses on bringing json transaction tests from https://github.com/ethereum/tests to modular rust code so that one can test and tune one's own EVM runtime on code without https://github.com/input-output-hk/mantis/blob/develop/ets/retesteth

This is just a rust typing, so It does not consider gas cost for now.

## Testing

Tests should generally check whether each test typings include `general information about test`, `results with gas amount` and `txbytes` to reproduce transaction

Run each test with aurora vm and check whether results match what the test json file describes.

## Additional information

Feeding to EVM is benchmarking the code [here](https://github.com/rust-blockchain/evm-tests/blob/4f7e566217e3b781be63db7704c84d93ae999be4/jsontests/tests/vm.rs#L1)


hskang9:
This task is already included in #619. closing this.

PR #614, opened 2022-09-28
Test(engine): Increase unit test coverage

RomanHodulak:
## Description

Increases unit test coverage of the `aurora-engine`.

None because these changes only affect tests.

## Testing

Done by the tests included in the PR.

No areas to highlight.

## Additional information

One thing to note is that there is a new crate called `aurora-test-doubles`.

This crate exists to be able to reuse mock, fake, dummy, stub, etc. objects (for example, in-memory `IO` trait implementation) in unit tests and the `aurora-tests` crate.


joshuajbouw (approved):

joshuajbouw:
I believe that the `aurora-test-doubles` crate could exist elsewhere later, but we can roll with it for now.

PR #615, opened 2022-09-30
Chore(standalone): Upgrade rocksdb to v0.19.

### LINKED ISSUE #612: Update rocksdb to v0.19.0
Out-of-bounds read when opening multiple column families with TTL.

Affected versions of this crate called the RocksDB C API rocksdb_open_column_families_with_ttl() with a pointer to a single integer TTL value, but one TTL value for each column family is expected. This is only relevant when using rocksdb::DBWithThreadMode::open_cf_descriptors_with_ttl() with multiple column families. This bug has been fixed in v0.19.0.

---

Reported by auditors. This doesn't affect us since we are not using `*_ttl` interface but still is worth updating (if not other breaking changes).

birchmd:
Closes #612 

Upgrading rocksdb also requires updating the nearcore dependency in tests so that there is only one version of librocksdb-sys used. Updating nearcore required us to also update rust-toolchain since nearcore uses newer (v1.64.0) Cargo.toml features.

The newer version of rocksdb also changed the API for iterating over keys in the DB, this is why there is now extra error handling in `engine-standalone-storage`.

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -42,6 +42,8 @@ jobs:
           git checkout -f $(git -c user.name=x -c user.email=x@x commit-tree $(git hash-object -t tree /dev/null) < /dev/null) || :
       - name: Clone the repository
         uses: actions/checkout@v3
+      - name: Update udeps
+        run: cargo install --force cargo-udeps
       - name: Run udeps
         run: cargo make udeps
   contracts:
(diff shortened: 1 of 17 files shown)


joshuajbouw:
Looks like Udep is failing due to some missing commit?

birchmd:
> Looks like Udep is failing due to some missing commit?

No, udep just needs to be updated on the build server. You can verify locally that all checks pass if you reinstall udep

mfornet (approved):

joshuajbouw (approved):

PR #616, opened 2022-10-03
Fix(xcc): Only update the router contract version in storage if the deploy is successful

birchmd:
## Description

In the xcc flow, we store the current version of each deployed sub-account in the Engine's storage. When a router contract is out of date (or the sub-account does not yet exist) then we must deploy the new version and update the entry in storage. The storage update is done as a callback after the deploy. In this PR we add logic to check the deploy was successful before updating the version in storage.

## Testing

Existing xcc tests are used for testing (the deploy flow is tested because it occurs when creating a sub-account for the first time).


mfornet (approved):
Good catch

joshuajbouw (approved):
Good catch indeed

PR #617, opened 2022-10-03
Refactor: Mark functions that create promises on NEAR as unsafe

birchmd:
Creating promises on NEAR from the Engine account is dangerous because of the admin privileges it holds. In this PR we mark the functions which create promises as `unsafe` so that the Rust compiler will not let us use them outside of an `unsafe` block. The purpose of this change is to draw attention to new usages of these functions in the future. These functions must always be accompanied by a comment describing why it is safe to make a promise in that instance, and reviewers of the PR which introduces the new usage must verify the comment is accurate. This will hopefully make it more difficult for accidental security vulnerabilities to be introduced in the future.

This PR is purely a refactor; no functionality is changed.


mfornet (approved):

joshuajbouw (approved):
Excellent! I had intended for this to happen. Its just good practice. Good job.

PR #618, opened 2022-10-03
Test(engine): Increase unit test coverage

RomanHodulak:
## Description

Increases unit test coverage of the `aurora-engine`.

None because these changes only affect tests.

## Testing

Done by the tests included in the PR.

Aside from adding tests, there are some changes in `Engine` code to pay attention to.

## Additional information

The `Engine::view` function signature has been changed slightly for simpler testing. Even though it is public, I expect it to be only called from within `Engine` or the engine smart contract.

The `Engine::receive_erc20_tokens` had a parameter removed along with a bit of code. The bit of code was wrapped in a condition that would be always false. And the removed parameter was only used in this bit of code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,6 +111,7 @@ dependencies = [
  "bitflags",
  "borsh",
  "byte-slice-cast",
+ "digest 0.10.3",
  "ethabi",
  "evm",
  "hex",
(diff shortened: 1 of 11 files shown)


joshuajbouw (approved):
Most excellent!

PR #622, opened 2022-10-14
Fix(xcc): Ensure the xcc router attaches enough gas to the execute function

birchmd:
## Description

While working on the native token connector, I found that I could not eagerly execute promises with callbacks using the XCC mechanism. It turned out this was because the XCC precompile was not attaching enough NEAR gas to the router's `execute` function when there are promises involved. The amount of gas needed scales linearly with the number of callbacks. This PR fixes the issue and the XCC precompile now accounts for the linearly increasing amount of gas needed for promises with callbacks.

## Testing

I've updated `test_xcc_exec_gas` to loop through promises with different numbers of callbacks.

### DIFF
--- a/engine-precompiles/src/xcc.rs
+++ b/engine-precompiles/src/xcc.rs
@@ -9,7 +9,7 @@ use aurora_engine_types::{
     account_id::AccountId,
     format,
     parameters::{CrossContractCallArgs, PromiseCreateArgs},
-    types::{balance::ZERO_YOCTO, Address, EthGas},
+    types::{balance::ZERO_YOCTO, Address, EthGas, NearGas},
     vec, Cow, Vec, H160, H256, U256,
 };
 use borsh::{BorshDeserialize, BorshSerialize};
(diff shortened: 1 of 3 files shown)


engine-precompiles/src/xcc.rs
```diff
@@ -42,7 +42,8 @@ pub mod costs {
     /// `NEAR Gas / EVM Gas`, we simply multiply `0.175 * 10^12 / 10^3 = 175 * 10^6`.
     pub const CROSS_CONTRACT_CALL_NEAR_GAS: u64 = 175_000_000;
 
-    pub const ROUTER_EXEC: NearGas = NearGas::new(7_000_000_000_000);
+    pub const ROUTER_EXEC_BASE: NearGas = NearGas::new(7_000_000_000_000);
+    pub const ROUTER_EXEC_PER_CALLBACK: NearGas = NearGas::new(12_000_000_000_000);
```
mfornet, line 46: How are these numbers determined? As I understand this is just for scheduling new transactions, but the cost seems higher than expected since creating new promises should be on the order of 1TGas.

mfornet (approved):
In general, looks good, so approving.

engine-tests/src/tests/xcc.rs
```diff
@@ -134,9 +134,14 @@ fn test_xcc_eth_gas_cost() {
...
-    let total_gas2 = y2 + baseline.all_gas();
+    // As a sanity check, confirm that the total EVM gas spent aligns with expectations.
+    // The additional gas added is the amount attached to the XCC call (this is "used", but not
+    // "burnt").
+    let total_gas1 = y1 + baseline.all_gas() + costs::ROUTER_EXEC_BASE.as_u64();
+    let total_gas2 = y2
+        + baseline.all_gas()
+        + costs::ROUTER_EXEC_BASE.as_u64()
+        + costs::ROUTER_EXEC_PER_CALLBACK.as_u64();
     assert!(
```
joshuajbouw, line 145: Just noting now that assertions that could lead to failure in the future shouldn't be a thing, only errors.

engine-tests/src/tests/xcc.rs
```diff
@@ -589,6 +605,17 @@ fn test_xcc_schedule_gas() {
 fn test_xcc_exec_gas() {
     let mut router = deploy_router();
 
+    let create_promise_chain =
```
joshuajbouw, line 608: Just a comment, good practice to have named defined functions which are self-descriptive. Good work!

joshuajbouw (approved):
Looks good thanks.

PR #623, opened 2022-10-18
Update README to exclude pre-req of installing specific Rust version

joshuajbouw:
## Description

The README currently displays out of date information about the current Rust version which we use. This simply removes that detail as it is no longer needed as its handled by the toolchain file automatically.

Make sure that there are no other redundant details in the README.


aleksuss:
There is a wrong command for installing `aurora-cli` in the README.md:
`npm install -g aurora-is-near/aurora-cli` -> `npm install -g @auroraisnear/cli`

birchmd (approved):

PR #626, opened 2022-10-20
fix: remove sscache from CI

joshuajbouw:
## Description

The CI is currently broken, and according to @strokovok it fails with [sscache](https://lib.rs/crates/sccache). Removing this should get it running again. Right now it is quite unclear why it is hanging here and may be some issue with our own build servers.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,4 +44,3 @@ jobs:
 env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
-  RUSTC_WRAPPER: sccache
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -44,4 +44,3 @@ jobs:
 env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
-  RUSTC_WRAPPER: sccache
```
aleksuss, line 47: Probably you should delete `sccache` in `cache-util restore/save ...` as well? It seems doesn't make sense anymore.

Issue #627, opened 2022-10-21
Aurora Engine crashes before hitting the EVM call stack limit


birchmd:
The EVM has a call stack limit of 1024. If there are more nested calls than this, then no call is performed and the failure is propagated back up the call chain. However, when executing a contract on Aurora that makes deeply nested calls, then it crashes before hitting this limit. This is an incompatibility with Ethereum and therefore must be addressed.

The core issue actually arises from SputnikVM, the EVM interpreter used by Aurora Engine. The way nested calls are executed is by literally recursively calling the `call_inner` [function](https://github.com/rust-blockchain/evm/blob/6534c1dd8ad77b53d05032f80e8a5f2de4d37fd2/src/executor/stack/executor.rs#L815), thus using Rust's call stack to implement the EVM call stack. The problem with this approach is that the Rust call stack can (and does) run out of space before hitting the EVM limit.

A better approach would be to have a data structure to model the EVM call stack and handle nested EVM calls in a loop rather than with recursive Rust function calls.

PR #628, opened 2022-10-21
Feat(tests): Test for trying to obtain TransactionStatus::CallTooDeep from the Engine

birchmd:
## Description

This PR is related to issue #627 . In this PR I add a test illustrating the problem outlined in that issue. Once this test passes, we know the issue is resolved.

## Testing

A new test includes an infinitely recursive Solidity contract. The point of this test is to create a situation where the EVM call stack limit is exceeded and see how the engine behaves. Presently the test is failing (per the issue described in #627 ), but once it passes then we should merge this PR.

## Additional information

The fix itself will be in SputnikVM, not in the Engine. I'll make a PR in that repo and reference this PR in it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1318,8 +1318,8 @@ dependencies = [
 
 [[package]]
 name = "evm"
-version = "0.35.0"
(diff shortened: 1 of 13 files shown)


joshuajbouw:
On rebase CI should be fixed.

joshuajbouw (approved):
Sucks to have a gas cost regression but this is necessary :(

PR #629, opened 2022-10-31
new_eth_connector: Fix permission checks

jumpsiegel:
new_eth_connector is called during deployment from the client and not from inside the contract itself.


joshuajbouw:
This is a known issue, but can't -quite- be fixed yet. There are numerous other examples of this which will be fixed up before the end of the year.

aleksuss:
No longer relevant

PR #633, opened 2022-11-03
Build: Add reproducible build job.

RomanHodulak:
## Description

Adds a build job that produces identical binaries across different build environments.

This is useful to verify that the binary wasn't changed, preventing malicious code from being deployed.

None.

## Testing

Ran locally with @joshuajbouw, and resulted in the same hash of the WASM binary.

```bash
shasum -a 256 bin/aurora-mainnet.wasm
```

Nothing to highlight.

#### Steps to produce
```
cargo make --profile <profile> build-docker
shasum -a 256 bin/aurora-<profile>.wasm
```

#### Expected results
```
57bb07e64187780ddd9870efbe683634f95a98200c026a950757dc0fd3876fca  bin/aurora-mainnet.wasm
789029331d468155bbee6114f70d05945d45691439a69a70bec14f0e99558630  bin/aurora-testnet.wasm
```


sept-en (commented):
Could you please provide the expected verified binary hash and instructions on reproducing it locally?

joshuajbouw (changes requested):
As what @sept-en said, we need instructions written up in in the README file and also include the binary hash that the process created in the PR so that we can have multiple people building.

PR #634, opened 2022-11-04
chore: replace unnecessary usage of heap allocated strings in the tests

aleksuss:
## Description

The PR replaces the usage of heap allocated strings for `&str` in the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ wget https://github.com/aurora-is-near/aurora-engine/releases/download/latest/ma
 ### Installing the Aurora CLI tool
 
 ```sh
-npm install -g aurora-is-near/aurora-cli
+npm install -g @auroraisnear/cli
 ```
 
 ### Deploying the EVM with the CLI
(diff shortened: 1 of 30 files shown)


README.md
```diff
@@ -145,7 +145,7 @@ wget https://github.com/aurora-is-near/aurora-engine/releases/download/latest/ma
 ### Installing the Aurora CLI tool
 
 ```sh
-npm install -g aurora-is-near/aurora-cli
+npm install -g @auroraisnear/cli
 ```
```
joshuajbouw, line 149: Actually this whole CLI section should be remove as this library is deprecated.

joshuajbouw (approved):

README.md
joshuajbouw: Can we confirm whether "@auroraisnear" is us or not? If it isn't, then we are exposing people to dangerous tech.

aleksuss: It works for now. I've checked it out. We could rework the section when `aurora-cli-rs` will be ready.

joshuajbouw: Because it works, doesn't mean it is ours. Its dangerous to proceed with this section. I would remove all mention of the CLI as it is deprecated.

aleksuss: Ok. We will modify the README after the `aurora-cli-rs` release.

mrLSD (approved):

PR #646, opened 2022-12-05
Fix(tracing): update to SputnikVM `v0.37.3-aurora`

birchmd:
## Description

There was a bug introduced in SputnikVM `v0.37.1-aurora` where some `Exit` tracing events were missing, which caused the tracing feature not to work correctly. This was fixed in `v0.37.3-aurora`. In this PR we update to the fixed version and add new tests to prevent such a regression in the future.

## Testing

New tests for tracing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1318,7 +1318,7 @@ dependencies = [
 [[package]]
 name = "evm"
 version = "0.37.0"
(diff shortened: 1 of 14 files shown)


joshuajbouw (approved):
Thanks for including tests on this level.

PR #645, opened 2022-11-29
Chore: Update to SputnikVM `v0.37.2-aurora`

birchmd:
## Description

Updates to SputnikVM `v0.37.2-aurora`, which includes some performance improvements.

## Performance / NEAR gas cost considerations

Provides about a 5% reduction in Aurora Engine gas usage.

## Testing

Existing tests.

PR #648, opened 2022-12-07
Release 2.8.1

birchmd:
### Changes

Performance improvement (approximately 5% reduction in NEAR gas usage) by @birchmd #645 

### Fixes

Tracing bug fix by @birchmd #646


joshuajbouw (approved):
Looks good and bumps where they needed to be.

engine-tests/src/tests/standalone/call_tracer.rs
```diff
@@ -0,0 +1,350 @@
...
+    assert_eq!(U256::from_big_endian(&subcall.output), U256::one());
+
+    runner.close();
+}
+
+/// A convenience function for pulling out a sub-call from a trace.
+/// The `path` gives the index to pull out of each `calls` array.
+/// For example `path == []` simply returns the given `root`, while
+/// `path == [2, 0]` will return `root.calls[2].calls[0]`.
+fn subcall_lense<'a, 'b>(
```
aleksuss, line 341: Is this function used anywhere else rather than in the tests? If not then should be annotated as `#[cfg(test)]`?

aleksuss (approved):
Overall looks good

PR #652, opened 2022-12-14
function create_legacy_address

Casuso:


engine/Cargo.toml
```diff
@@ -30,6 +30,7 @@ hex = { version = "0.4", default-features = false, features = ["alloc"] }
 rjson = { git = "https://github.com/aurora-is-near/rjson", rev = "cc3da949", default-features = false, features = ["integer"] }
 rlp = { version = "0.5.0", default-features = false }
 serde = { version = "1", features = ["derive"], optional = true }
+sha3 = "0.10.6"
```
birchmd, line 33: I think pulling in the dependency like this is what is causing CI to fail (would need to bring it in with `default-features = false` to play nice with the `no-std` aspects of this crate).

I suggest using `aurora_sdk::types::keccak` instead of pulling in the `sha3` dependency directly (`aurora-sdk` is already a dependency of this crate, so you won't need to change `Cargo.toml` at all).

Casuso:
Closing this PR to target develop branch in another PR.

PR #653, opened 2022-12-14
create legacy address function

Casuso:
## Description
Adds an utility function to create legacy addresses from a caller address and a nonce. We want to use this on the refiner to populate contract addresses.

I founded an almost identical function in the tests on engine/src/engine.rs so I remove it and keep only the new public one. The function uses the aurora_engine_sdk keccak function and not directly the one from sha3 crate, so I remove it from the dev dependencies. 

## Testing
Ran all the tests and added a new one targeting the new function.

I would like you to give me feedback about rust coding. If you see anything that is wrong or if you think there is another way of doing things that is more elegant or common, pls let me know.

### DIFF
--- a/engine/src/engine.rs
+++ b/engine/src/engine.rs
@@ -1277,7 +1277,9 @@ pub fn create_legacy_address(caller: &Address, nonce: &U256) -> Address {
     let mut stream = rlp::RlpStream::new_list(2);
     stream.append(&caller.raw());
     stream.append(nonce);
-    Address::new(aurora_engine_sdk::keccak(&stream.out()).into())
+    let hash = aurora_engine_sdk::keccak(&stream.out());
+    let hash_bytes = hash.as_bytes();
+    Address::try_from_slice(&hash_bytes[12..]).unwrap()
 }
 
 pub fn nep141_erc20_map<I: IO>(io: I) -> BijectionMap<NEP141Account, ERC20Address, I> {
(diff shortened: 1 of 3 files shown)


joshuajbouw:
Can you please follow the PR description format? It helps with context.

Casuso:
General question @joshuajbouw and @birchmd. In the description format it mentioned that the commits should be squash. Is this something that we currently want? It will be done automatically by github or some CI flow (I was used to this option)?

joshuajbouw:
We typically will just squash and merge it, not exactly a requirement now. Its important that you are signing your commits with your work e-mail. Prior, we couldn't squash for someone and choose their work e-mail to use as the commit e-mail, now we can.

engine/src/engine.rs
```diff
@@ -2281,6 +2289,18 @@ mod tests {
...
 
+    #[test]
+    fn test_create_legacy_address() {
+        let caller = Address::decode("3160f7328df59c14d85dfd09addad4ef18ae3e2c").unwrap();
+        let nonce = U256::from_dec_str("109438").unwrap();
+        let created_address = create_legacy_address(&caller, &nonce);
+
+        assert_eq!(
+            created_address.encode(),
+            "140e8a21d08cbb530929b012581a7c7e696145ef"
```
birchmd, line 2300: Might be nice to document where this example came from.

birchmd (approved):
Lgtm

engine/src/engine.rs
Casuso: Transaction data included.

joshuajbouw (approved):
Looks good.

PR #655, opened 2022-12-19
Feat(tracing): include deployed code in trace output for CREATE

birchmd:
## Description

EVM tracing is expected to include the code deployed during a `CREATE`/`CREATE2` operation (see for example the `code` field in https://docs.alchemy.com/reference/what-are-evm-traces#create ). This PR updates to a newer version of SputnikVM where I have added this functionality and includes a test for it.

Note: this PR does not impact the Aurora Engine smart contract, only the tracing functionality which is used by the standalone engine.

## Testing

New test for tracing a contract deploy transaction.

### DIFF
--- a/engine-standalone-tracing/src/types/call_tracer.rs
+++ b/engine-standalone-tracing/src/types/call_tracer.rs
@@ -26,7 +26,13 @@ impl CallTracer {
     fn end(&mut self, output: Vec<u8>, error: Option<&evm::ExitReason>) {
         let frame = self.call_stack.first_mut().unwrap();
         match error {
-            None => frame.output = output,
+            None => {
+                match frame.call_type {
+                    // In CREATE case, output is set by the `CreateOutput` rather than the `Exit` event.
+                    CallType::Create | CallType::Create2 => (),
+                    _ => frame.output = output,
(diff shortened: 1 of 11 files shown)


engine-standalone-tracing/src/types/call_tracer.rs
```diff
@@ -68,14 +74,17 @@ impl CallTracer {
 
         let mut frame = self.call_stack.pop().unwrap();
         match error {
-            None => frame.output = output,
-            Some(error) => {
-                frame.error = Some(format!("{:?}", error));
+            None => {
                 match frame.call_type {
-                    CallType::Create | CallType::Create2 => frame.to = None,
```
birchmd, line 75: I've removed this logic because it was incorrect. Even when there is an error in the deploy execution the `to` address that would have been created is still reported. For example, see this [transaction on Etherscan](https://goerli.etherscan.io/tx/0xdeff004bc889227cd416f39014d352421e8a18ada67123ad295f80269fcc9a2f).

aleksuss (approved):
LGTM

joshuajbouw (approved):
LGTM

PR #661, opened 2023-01-22
chore: remove dynamic memory allocation while creating precompile flags

aleksuss:
## Description

The PR adds tiny optimisation by removing unnecessary dynamic memory allocation while creating precompile flags from storage value.

### DIFF
--- a/engine/src/pausables.rs
+++ b/engine/src/pausables.rs
@@ -117,10 +117,7 @@ impl<I: IO> EnginePrecompilesPauser<I> {
             None => PrecompileFlags::empty(),
             Some(bytes) => {
                 const U32_SIZE: usize = core::mem::size_of::<u32>();
-
-                if bytes.len() < U32_SIZE {
-                    return PrecompileFlags::empty();
-                }
+                assert_eq!(bytes.len(), U32_SIZE, "ERR_CORRUPTED_PAUSE_FLAGS");
 
                 let mut buffer = [0u8; U32_SIZE];
                 bytes.copy_to_slice(&mut buffer);
--- a/engine/src/pausables.rs
+++ b/engine/src/pausables.rs
@@ -116,17 +116,16 @@ impl<I: IO> EnginePrecompilesPauser<I> {
         match self.io.read_storage(&Self::storage_key()) {
             None => PrecompileFlags::empty(),
             Some(bytes) => {
-                let int_length = core::mem::size_of::<u32>();
-                let input = bytes.to_vec();
+                const U32_SIZE: usize = core::mem::size_of::<u32>();
 
-                if input.len() < int_length {
+                if bytes.len() < U32_SIZE {
                     return PrecompileFlags::empty();
                 }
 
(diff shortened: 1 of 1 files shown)


mrLSD (approved):

engine/src/pausables.rs
```diff
@@ -116,17 +116,16 @@ impl<I: IO> EnginePrecompilesPauser<I> {
         match self.io.read_storage(&Self::storage_key()) {
             None => PrecompileFlags::empty(),
             Some(bytes) => {
-                let int_length = core::mem::size_of::<u32>();
-                let input = bytes.to_vec();
+                const U32_SIZE: usize = core::mem::size_of::<u32>();
 
-                if input.len() < int_length {
+                if bytes.len() < U32_SIZE {
```
birchmd, line 121: I wonder if we should change this check to be something like

```rust
if bytes.len() != U32_SIZE {
    panic!("ERR_CORRUPTED_PAUSE_FLAGS");
}
```

The original code included a check for the input being too small, and used `split_at` to handle the case the input was too long, but it seems to me like the value in storage should always be exactly `U32_SIZE` since it is controlled by this struct.

aleksuss: The check for too small input is present: `if bytes.len() < U32_SIZE`, and as for too big input, we just won't read more amount of bytes than size of the `buffer` which is `U32_SIZE`. So, it should be ok. But yes, panic here really makes sense.

engine/src/pausables.rs
```diff
@@ -116,17 +116,13 @@ impl<I: IO> EnginePrecompilesPauser<I> {
         match self.io.read_storage(&Self::storage_key()) {
             None => PrecompileFlags::empty(),
             Some(bytes) => {
-                let int_length = core::mem::size_of::<u32>();
-                let input = bytes.to_vec();
+                const U32_SIZE: usize = core::mem::size_of::<u32>();
+                assert_eq!(bytes.len(), U32_SIZE, "ERR_CORRUPTED_PAUSE_FLAGS");
```
joshuajbouw, line 120: We need get out of the habit of having `ERR_*`, instead using it to be a bit more like actual errors. There should be an overhaul later, given that internal services also use the engine. A bit like this, for now, is ok: https://github.com/aurora-is-near/aurora-engine/blob/feat/erc20-gas-token/engine/src/state.rs#L59-L84.

joshuajbouw (changes requested):

engine/src/pausables.rs
aleksuss: @joshuajbouw, your proposal is overkill, in this case, IMO. 
1. Because it's a panic and not returning an error.
2. In this module, we don't use errors at all. And therefore creating an additional error module is overkill.

joshuajbouw: Ah fair, in this regard. I just see `"ERR_*"` and get triggered.

joshuajbouw (approved):

PR #662, opened 2023-01-23
feat(engine): erc20 gas token

joshuajbouw:
## Description

This PR introduces an exciting feature unique to Aurora where you are able to pay in a gas token other than the base currency, in our case ETH. AURORA and others can then be used instead to pay for gas costs.

WIP

## Testing

WIP

WIP

## Additional information

WIP

### DIFF
--- a/engine/src/engine.rs
+++ b/engine/src/engine.rs
@@ -414,10 +414,20 @@ impl From<NewCallArgs> for EngineState {
     }
 }
 
+/// Used to select which gas token to pay in.
+pub enum GasToken {
+    /// Gas is paid in Ether.
+    ETH,
+    /// Gas is paid in a ERC-20 compatible token.
+    ERC20(Address),
+}
+
+
 pub struct Engine<'env, I: IO, E: Env> {
     state: EngineState,
     origin: Address,
     gas_price: U256,
+    gas_token: GasToken,
     current_account_id: AccountId,
     io: I,
     env: &'env E,
@@ -453,6 +463,7 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
             state,
             origin,
             gas_price: U256::zero(),
+            gas_token: GasToken::ETH,
             current_account_id,
             io,
             env,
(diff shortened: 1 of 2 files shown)


aleksuss (approved):

engine/src/state.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    #[test]
+    fn test_missing_engine_state_is_not_found() {
+        let storage = Storage::default();
+        let storage = RwLock::new(storage);
+        let io = StoragePointer(&storage);
+
+        let actual_error = get_state(&io).unwrap_err();
+        let actual_error = std::str::from_utf8(actual_error.as_ref()).unwrap();
+        let expected_error = std::str::from_utf8(error::ERR_STATE_NOT_FOUND).unwrap();
```
aleksuss, line 153: It seems that creating utf8 strings is redundant. You can just compare `&[u8]` instead.
```rust
assert_eq!(actual_error.as_ref(), error::ERR_STATE_NOT_FOUND);
```

joshuajbouw: Will rework it. Part of code that I refactored, but haven't looked into improving yet. I also don't necessarily like the `RwLock`, seems to be a bit overkill.

engine/src/engine.rs
```diff
@@ -1243,6 +1296,14 @@ pub fn get_balance<I: IO>(io: &I, address: &Address) -> Wei {
     Wei::new(raw)
 }
 
+/// Checks whether there are enough funds for the given address and amount.
+///
+/// This does not take gas into account.
+pub fn can_transfer<I: IO>(io: &I, address: &Address, amount: Wei) -> bool {
+    let current_balance = get_balance(io, address);
+    amount >= current_balance
```
aleksuss, line 1304: Probably should be less or equal?
 
```suggestion
    amount <= current_balance
```

joshuajbouw:
Whatever happened with this because it still is technically possible?

aleksuss:
Hey @joshuajbouw. Yes, it is. But we no longer have a request for this feature, and we have more prioritised tasks to do now.

PR #664, opened 2023-01-24
chore: some charge gas optimisations

aleksuss:
## Description

Charge gas optimisation

### DIFF
--- a/engine-tests/src/tests/sanity.rs
+++ b/engine-tests/src/tests/sanity.rs
@@ -843,8 +843,8 @@ fn test_eth_transfer_charging_gas_not_enough_balance() {
         &runner,
         source_address,
         INITIAL_BALANCE,
-        // nonce is still incremented since the transaction was otherwise valid
-        (INITIAL_NONCE + 1).into(),
+        // we shouldn't increment nonce in case of OutOfFund error
+        (INITIAL_NONCE).into(),
     );
     test_utils::validate_address_balance_and_nonce(&runner, dest_address, Wei::zero(), 0.into());
     test_utils::validate_address_balance_and_nonce(&runner, relayer, Wei::zero(), 0.into());
@@ -1015,7 +1015,7 @@ fn test_eth_transfer_charging_gas_not_enough_balance_sim() {
     // validate post-state
     assert_eq!(
         query_address_sim(&address, "get_nonce", &aurora),
-        U256::from(INITIAL_NONCE + 1),
+        U256::from(INITIAL_NONCE),
     );
     assert_eq!(
         query_address_sim(&address, "get_balance", &aurora),
--- a/engine/src/engine.rs
+++ b/engine/src/engine.rs
@@ -475,7 +475,7 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
 
     // TODO: If we don't actually charge_gas here after changes, this should
     // be renamed.
-    pub fn charge_gas(
+    pub fn get_gas_payment(
         &mut self,
         sender: &Address,
         transaction: &NormalizedEthTransaction,
@@ -494,20 +494,15 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
             .checked_mul(effective_gas_price)
             .map(Wei::new)
             .ok_or(GasPaymentError::EthAmountOverflow)?;
+        let current_balance = get_balance(&self.io, sender);
 
         match gas_token {
             GasToken::ETH => {
-                // TODO: Remove the set_balance as we need to just check if they can spend
-                // Use new `can_transfer` function instead.
-                // Also needs to be verified how it is done on geth.
-                let new_balance = get_balance(&self.io, sender)
-                    .checked_sub(prepaid_amount)
-                    .ok_or(GasPaymentError::OutOfFund)?;
-
-                // This part is questionable.
-                set_balance(&mut self.io, sender, &new_balance);
+                if current_balance < prepaid_amount {
+                    return Err(GasPaymentError::OutOfFund);
+                }
             }
-            GasToken::ERC20(_addr) => {
+            GasToken::ERC20(_address) => {
                 // TODO: Needs SputnikVM balance check
                 todo!()
             }
@@ -974,7 +969,7 @@ pub fn submit<I: IO + Copy, E: Env, P: PromiseHandler>(
 
     let mut engine = Engine::new_with_state(state, sender, current_account_id, io, env);
     // TODO: Have GasToken derived from storage.
-    let prepaid_amount = match engine.charge_gas(&sender, &transaction, GasToken::ETH) {
+    let prepaid_amount = match engine.get_gas_payment(&sender, &transaction, GasToken::ETH) {
         Ok(gas_result) => gas_result,
         Err(GasPaymentError::OutOfFund) => {
             let result = SubmitResult::new(TransactionStatus::OutOfFund, 0, vec![]);
@@ -1050,12 +1045,12 @@ pub fn submit<I: IO + Copy, E: Env, P: PromiseHandler>(
         Ok(submit_result) => submit_result.gas_used,
         Err(engine_err) => engine_err.gas_used,
     };
-    refund_unused_gas(&mut io, &sender, gas_used, prepaid_amount, &relayer_address).map_err(
-        |e| EngineError {
+    charge_gas(&mut io, &sender, gas_used, prepaid_amount, &relayer_address).map_err(|e| {
+        EngineError {
             gas_used,
             kind: EngineErrorKind::GasPayment(e),
-        },
-    )?;
+        }
+    })?;
 
     // return result to user
     result
@@ -1173,7 +1168,7 @@ pub fn get_authorizer() -> EngineAuthorizer {
 }
 
 // TODO: Gas must be charged from here.
-pub fn refund_unused_gas<I: IO>(
+pub fn charge_gas<I: IO>(
     io: &mut I,
     sender: &Address,
     gas_used: u64,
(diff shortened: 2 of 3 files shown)


engine-tests/src/tests/sanity.rs
```diff
@@ -843,8 +843,8 @@ fn test_eth_transfer_charging_gas_not_enough_balance() {
         &runner,
         source_address,
         INITIAL_BALANCE,
-        // nonce is still incremented since the transaction was otherwise valid
-        (INITIAL_NONCE + 1).into(),
+        // we shouldn't increment nonce in case of OutOfFund error
```
joshuajbouw, line 846: Can you remove anyhting that has to do with the OutOfFund issue please?

joshuajbouw (changes requested):

birchmd (commented):
Looks to me like the balance used to cover the gas limit is not deducted from the user's account before executing the transaction.

Could you add a test where the user attempts a double-spend in this way? The set up is that the user has enough ETH to cover their gas limit, and enough ETH to cover a transfer they want to do, but not enough ETH to cover both. The expected behaviour is that the EVM transaction reverts with OutOfFunds when the transfer is attempted (the Near transaction is still a success so the user is charged for the gas they consumed in attempting the transfer; it is only EVM transaction that fails).

engine/src/engine.rs
```diff
@@ -921,19 +978,19 @@ pub fn submit<I: IO + Copy, E: Env, P: PromiseHandler>(
...
 
-    // Give refund
+    // Getting an amount of real gas usage after the execution of the transaction.
     let gas_used = match &result {
         Ok(submit_result) => submit_result.gas_used,
         Err(engine_err) => engine_err.gas_used,
     };
-    refund_unused_gas(&mut io, &sender, gas_used, prepaid_amount, &relayer_address).map_err(
-        |e| EngineError {
+    engine
```
mrLSD, line 986: This solution looks completely non-obvious and potentially dangerous to me. The fact is that `charge_gas` existed (executed) before the transaction was EVM completed, and the user's balance changed before the transaction was completed in the EVM. Those. In a fact, Transfer, and Balance **SHOULD** work differently than before. Because their balance amount will be different. Previously this would have been an EVM error, now it is a NEAR level error. Such a decision seems controversial and unreasonable.

engine/src/engine.rs
```diff
@@ -431,17 +430,15 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
...
             .map(Wei::new)
             .ok_or(GasPaymentError::EthAmountOverflow)?;
+        let balance = self.cached_balance(sender);
 
-        let new_balance = get_balance(&self.io, sender)
-            .checked_sub(prepaid_amount)
-            .ok_or(GasPaymentError::OutOfFund)?;
-
-        set_balance(&mut self.io, sender, &new_balance);
+        if balance < prepaid_amount {
```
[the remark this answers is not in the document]
joshuajbouw: Correct, the gas must be deducted THEN stored in memory. From there Sputnik MUST get the balance from memory, and no longer from storage in the `basic` method of the `Backend` `impl` for `Engine`. That also is missing.

joshuajbouw (changes requested):

engine/src/engine.rs
```diff
@@ -431,13 +431,12 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
             .checked_mul(effective_gas_price)
```
mrLSD, line 431: if `effective_gas_price` is zero, we can do a bit optimize it:
- `prepaid_amount` - in that case always = 0, we don't need `mul` it
- `new_balance` is the same as `balance`, we don't need `sub` it
- we don't need cache it, as the balance was not changed

mrLSD (changes requested):

engine/src/engine.rs
aleksuss: I don't think that checks are much much cheaper than `mul` and `sub`. IMO, it's not worse it.

aleksuss: And the goal of this PR is to reduce the number of operations with storage.

joshuajbouw (changes requested):
I would like to request one last test, just to ensure that Sputnik is getting the correct information:
* Charge gas
* Use method `Engine::basic` to get the balance (nonce is irrelevant for this test)
* Assert gas is deducted from `basic`

engine/src/engine.rs
mrLSD: So, we do not need to do mathematical calculations, but you still need to do these calculations because there is no benefit from this? Given that these mathematical operations are strict with verification (checked)?

What exactly concerns this PR is that there is no need to update the cache. It's definitely a redundant operation.

mrLSD: You just reduce 2 strict math operations, and 1 memory operation. Why it's not cheaper?

aleksuss: > You just reduce 2 strict math operations, and 1 memory operation. Why it's not cheaper?

Because checking values is not free.

mrLSD: 1. Not free. but 4 math instructions are cheaper than 1 JUMP instruction and memory operation?!!
2. When using the Aurora+, zero gas operations are common. In this case, `update_cached_balance` is a guaranteed read of the Storage. And it is not needed in that particular case.

joshuajbouw (changes requested):

engine-tests/src/tests/sanity.rs
```diff
@@ -1025,6 +1025,36 @@ fn test_eth_transfer_charging_gas_not_enough_balance_sim() {
...
+        INITIAL_BALANCE,
+        INITIAL_NONCE.into(),
+    );
+    test_utils::validate_address_balance_and_nonce(&runner, dest_address, Wei::zero(), 0.into());
+
+    // do transfer
+    let result = runner
+        .submit_with_signer(&mut source_account, transaction)
+        .unwrap();
+    // The status is `OutOfFund` because we try to spend 800_000 (transfer) + 300_000 (gas prepaid)
```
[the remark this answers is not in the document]
aleksuss: The idea was provided by @birchmd. The user has enough ETH to cover their gas limit, and enough ETH to cover a transfer they want to do, but not enough ETH to cover both.

engine-tests/src/tests/sanity.rs
```diff
@@ -1053,6 +1053,17 @@ fn test_eth_double_spending_trying_sim() {
...
     assert_eq!(result.status, TransactionStatus::OutOfFund);
+
+    // validate post-state
+    test_utils::validate_address_balance_and_nonce(
+        &runner,
+        source_address,
+        // Cover the EVM gas spent on the transaction before failing. 21_000 is the base cost of
+        // initiating a transaction in the EVM.
+        INITIAL_BALANCE - Wei::new_u64(21_000 * GAS_PRICE),
+        (INITIAL_NONCE + 1).into(),
```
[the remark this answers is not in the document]
aleksuss: I'm not sure. The error occurs in the EVM.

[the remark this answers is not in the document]
joshuajbouw: Interesting, this should be explored. Can you do that please?

engine-tests/src/tests/sanity.rs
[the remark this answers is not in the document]
birchmd: Thia test setup and outcome is exactly what I asked for. No changes needed!

aleksuss:
I should mention before. Upcoming commit brings this changes.

engine-precompiles/src/promise_result.rs
```diff
@@ -16,7 +16,7 @@ pub mod costs {
     use crate::prelude::types::EthGas;
 
     /// This cost is always charged for calling this precompile.
-    pub const PROMISE_RESULT_BASE_COST: EthGas = EthGas::new(105);
+    pub const PROMISE_RESULT_BASE_COST: EthGas = EthGas::new(53);
```
aleksuss, line 19: @joshuajbouw @birchmd this change is a bit confusing to me. Could you check it out?

engine-tests/src/tests/sanity.rs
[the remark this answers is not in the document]
joshuajbouw: Thanks for the clarification, it can be tough to tell in go-ethereum.

engine-precompiles/src/promise_result.rs
birchmd: This gas cost is computed from [this test](https://github.com/aurora-is-near/aurora-engine/blob/master/engine-tests/src/tests/promise_results_precompile.rs#L52). The basic idea of this test is to measure the gas cost by doing a linear regression of the actual cost (measured in Near gas) of calling this precompile for different sizes of promise data. It's actually not much of a linear regression because I just take two points and compute the line through them, but we also don't need this to be super accurate (the gas cost is only to prevent abuse) so I think that's ok.

Anyway, the cost is supposed to exclude overhead that applies to all transactions by subtracting a baseline transaction. So it is surprising to me that this change impacts this value. But the test also allows a 5% wiggle room on the value, so the change might be smaller than it looks if we were on the edge of being outside the 5% range before. Overall, I think this change is ok even if it is unexpected.

aleksuss:
> One question I have about this: it's supposed to be a gas optimization, but none of the benchmark tests (repro, uniswap, etc) needed to be updated after this change. So did this optimization really have an impact on our gas usage?

Yes. As I said, the last commit brings some TGas decreasing. And I think a more significant decrease in the costs we would see in the process of working the aurora because, in the tests, we don't use the cache effectively (the first call gets value from the storage rather than from the cache).

birchmd:
> in the tests, we don't use the cache effectively

Could you elaborate on this point more? I thought the tests should be accurate compared with the production deployment of the contract because they are both running in the Near runtime.

joshuajbouw:
> > in the tests, we don't use the cache effectively
> 
> Could you elaborate on this point more?

I'm curious about this as well, if so, then tests would need to be updated to accurately reflect this.

aleksuss:
> > > in the tests, we don't use the cache effectively
> > 
> > 
> > Could you elaborate on this point more?
> 
> I'm curious about this as well, if so, then tests would need to be updated to accurately reflect this. We should be seeing gas reductions though.

No. It's ok. It was my misunderstanding of how exactly the contract works.

aleksuss:
Some comparisons. First value is a result from repro_* test from develop branch and second one is from current branch.
```
repro_GdASJ3KESs:
122721043002795 - 122596036542627 = 125006460168 
repro_FRcorNv:
181030562726220 - 180905374436976 = 125188289244
repro_D98vwmi:
182732573474634 - 182607493789182 = 125079685452 
repro_8ru7VEA:
223178543337084 - 223056899480688 = 121643856396
repro_Emufid2:
306428381815360 - 306038787160204 = 389594655156
repro_5bEgfRQ:
657955355677707 - 657565031237979 = 390324439728
```

joshuajbouw:
I feel like we should go forward with this, if possible, for the release @aleksuss thoughts?

PR #666, opened 2023-01-25
fix(eth-contracts): slither / "immutable" admin

vzctl:
fixes slither errors


0x3bfc (commented):
Cool PR number 

Admin address can be changed with `adminSStore` which is controlled by admin itself. So I think this is can be considered as a false positive slither check.

aleksuss (approved):

PR #668, opened 2023-01-25
Chore: Relocate Engine parameter type definitions

birchmd:
## Description

We're working on an Aurora SDK for Near contracts. The goal of this project is to make it easier for developers on Near to interact with the Aurora Engine from their own contracts. Part of this effort includes exposing common Engine types used for input and output of Engine functions (for example `CallArgs` and `SubmitResult`). Unfortunately the `aurora-engine` crate itself cannot be included as a library in another Wasm contract, and maybe that's for the best since it's pretty heavy-weight to include the whole Engine logic in something that is only acting as a client. Therefore, in this PR I move these important types to be defined in `aurora-engine-types` instead. I have left them as `pub use` in `aurora-engine` to as to not break any crates that depended on their definition being in the Engine as well.

I also needed to add an addition feature to the `aurora-engine-types` crate to avoid `primitive-types/std` since it pulls in `getrandom` which cannot be compiled to Near Wasm.

None; this is pure refactor only.


engine/src/engine.rs
```diff
@@ -1464,12 +1464,25 @@ where
...
                 None
             } else {
-                Some(log.into())
+                Some(evm_log_to_result_log(log))
             }
         })
         .collect()
 }
 
+fn evm_log_to_result_log(log: Log) -> ResultLog {
```
birchmd, line 1473: I had to remove the `From<Log> for ResultLog` implementation when I moved the type definition to `aurora-engine-types` because I didn't want to pull in the whole `evm` dependency just for that small usage and Rust does not allow trait impls on types defined outside the crate.

aleksuss (approved):
LGTM

mrLSD (approved):
LGTM!

joshuajbouw (approved):
LGTM

PR #669, opened 2023-01-26
Feat: standalone nep141 legacy

mrLSD:
## Description

After splitting logic in #607 - Split NEP-141 from Engine, `engine-standalone` requires changes. Since `engine-standalone` cannot interact with the new `aurora-eth-connectpr` contract, the previous NEP-141-related codebase has been moved separately to `standalone-nep141-legacy`. This will ensure that `engine-standalone` works well with historical data.

Modules were added:
- `legacy_connector`
- `fungible_token`

It affects only standalone. So no side effects.

## Testing

Tests not changed for standalone. Just restored expected behaviour. For some unknown reason failed function `validate_standalone`, but that function don't affect any test.

### DIFF
--- a/engine-standalone-storage/src/sync/mod.rs
+++ b/engine-standalone-storage/src/sync/mod.rs
@@ -1,11 +1,14 @@
-use aurora_engine::admin_controlled::AdminControlled;
-use aurora_engine::pausables::{
-    EnginePrecompilesPauser, PausedPrecompilesManager, PrecompileFlags,
-};
-use aurora_engine::{connector, engine, parameters::SubmitResult, xcc};
+use aurora_engine::pausables::PausedPrecompilesManager;
+use aurora_engine::pausables::PrecompileFlags;
+use aurora_engine::{engine, parameters::SubmitResult, pausables::EnginePrecompilesPauser, xcc};
 use aurora_engine_sdk::env::{self, Env, DEFAULT_PREPAID_GAS};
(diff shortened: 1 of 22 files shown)


engine/src/lib.rs
```diff
@@ -842,6 +842,16 @@ mod contract {
             .set_eth_connector_contract_account(&args.account);
     }
 
+    #[no_mangle]
+    pub extern "C" fn disable_legacy_nep141() {
+        let io = Runtime;
+        io.assert_private_call().sdk_unwrap();
```
birchmd, line 848: Probably better to make it owner-only instead of private. Not that it matters too much since I think for `aurora` the owner is the engine itself.

engine-standalone-storage/src/sync/mod.rs
```diff
@@ -234,6 +237,7 @@ fn non_submit_execute<'db>(
             None
         }
 
+        TransactionKind::FtTransferCall(_) if is_disabled_legacy_nep141 => None,
```
birchmd, line 240: Please also add this for the other NEP-141 methods that become no-op after the split.

birchmd (commented):
Nice idea with the flag in the state to indicate if the legacy code can be disabled! I still think a more general solution involving the engine version would be better because it would generalize to other changes we might want to make, but this gets the job done for this case.

joshuajbouw (approved):
Yeah this can work for now. Thank you.

aleksuss (approved):

PR #671, opened 2023-01-26
fix: incremented nonce on out of fund failure

joshuajbouw:
## Description

A bug was found where we increment the nonce when a user is `OutOfFund`. After looking into it, there were some other areas where the nonce could be incremented incorrectly. As far as I can tell, with these changes, this shouldn't affect internal services, but that needs to be checked ideally by @birchmd.

No gas increases, though a slight reduction should be possible.

## Testing

Tests were updated for this fix.

While the fix should be correct, it should be checked that this is fine for internal services.

### DIFF
--- a/engine-tests/src/tests/sanity.rs
+++ b/engine-tests/src/tests/sanity.rs
@@ -150,7 +150,6 @@ fn test_state_format() {
     assert_eq!(hex::encode(state.try_to_vec().unwrap()), expected_hex);
 }
 
-#[cfg(test)]
 fn generate_code(len: usize) -> Vec<u8> {
     let mut rng = rand::thread_rng();
     let mut buf = vec![0u8; len];
--- a/engine-tests/src/tests/sanity.rs
+++ b/engine-tests/src/tests/sanity.rs
@@ -150,19 +150,20 @@ fn test_state_format() {
     assert_eq!(hex::encode(state.try_to_vec().unwrap()), expected_hex);
 }
 
+#[cfg(test)]
+fn generate_code(len: usize) -> Vec<u8> {
+    let mut rng = rand::thread_rng();
+    let mut buf = vec![0u8; len];
+    rng.fill_bytes(&mut buf);
+    buf
+}
+
 #[test]
 fn test_deploy_contract() {
     let (mut runner, mut signer, _) = initialize_transfer();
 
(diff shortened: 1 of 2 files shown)


engine/src/engine.rs
```diff
@@ -490,13 +483,7 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
         };
 
         let used_gas = executor.used_gas();
-        let status = match exit_reason.into_result(result) {
-            Ok(status) => status,
-            Err(e) => {
```
aleksuss, line 495: Is it correct that we don't increment the nonce in case of any error?

joshuajbouw: I'm double checking.

joshuajbouw (commented):
This looks good to me. Good code quality improvements, I appreciate it.

engine-tests/src/tests/sanity.rs
```diff
@@ -150,19 +150,20 @@ fn test_state_format() {
     assert_eq!(hex::encode(state.try_to_vec().unwrap()), expected_hex);
 }
 
+#[cfg(test)]
```
birchmd, line 153: Not needed, this whole crate is just tests.

engine/src/engine.rs
birchmd: Removing this code is fine because we were not actually incrementing the nonce in this case even though it looks like we were. In `lib.rs` we are unwrapping the `Result` from this function (and similarly for `call` and `submit`), so the contract will panic when we return an `Err` variant, thus reverting any state changes we made.

engine/src/engine.rs
```diff
@@ -1208,6 +1184,7 @@ pub fn get_nonce<I: IO>(io: &I, address: &Address) -> U256 {
         .unwrap_or_else(|_| U256::zero())
 }
 
+#[cfg(test)]
```
birchmd, line 1187: Do we actually use this function in tests or it can be deleted too?

birchmd (approved):

engine/src/engine.rs
aleksuss: This function is used in tests now only. But I decided to leave it just in case.

aleksuss (approved):

PR #673, opened 2023-01-31
chore: update some rust dependencies

aleksuss:
## Description

The PR updates some rust dependencies and extends `aurora-engine-sdk` crate with two functions for base64 en/decoding.


birchmd (approved):

0x3bfc (approved):

mrLSD (approved):
As I understand `borsh 0.10` was not updated because incompatible with test dependencies.

aleksuss:
> As I understand `borsh 0.10` was not updated because incompatible with test dependencies.

Yes. We use `unwrap_borsh` from `near-sdk-sim` which depends on `borsh 0.9.3`.

PR #677, opened 2023-02-02
feat: replace rjson with serde_json

aleksuss:
## Description

The PR replaces `rjson` crate with `serde_json`. 

There are no changes related to gas costs.

Pay attention to JSON API.

## Additional information

Question about `Yocto` struct. Should it have a custom de/serialisation like `Balance` and `NEP141Wei`?


joshuajbouw (changes requested):
We need to remove the `engine/src/json.rs` file too, that was added specifically to help make `rjson` json compliant.

aleksuss:
> We need to remove the `engine/src/json.rs` file too, that was added specifically to help make `rjson` json compliant.

Yes. Good catch. My fault.

joshuajbouw (approved):

joshuajbouw:
Needs to be explored we are unblocked on not being on Nightly anymore finally.

aleksuss:
Yes. But I preferred to do that in another PR.

engine/src/lib.rs
```diff
@@ -909,7 +906,7 @@ mod contract {
     #[no_mangle]
     pub extern "C" fn mint_account() {
         use crate::connector::ZERO_ATTACHED_BALANCE;
-        use crate::prelude::{NEP141Wei, U256};
+        use crate::prelude::{NEP141Wei, ToString, U256};
```
birchmd, line 909: Why is this extra import needed now?

birchmd (approved):

engine/src/lib.rs
aleksuss: Interesting notice. For some reason, the compiler asked to add it. It's because the method `to_string` is used below, I guess. I will check it out once again.

aleksuss: In fact, it's unnecessary 👍🏻

birchmd: Ah, I see. `ToString` was originally imported along with the other `crate::prelude` imports (circa L96), but it was removed from there so it needed to be included here now instead. Thanks for double checking!

mrLSD (approved):
LGTM!

PR #681, opened 2023-02-08
fix: return error if `amount` is greater than `2^128`

0x3bfc:
## Description

The `exitToNear` and `exitToEthereum` precompiles may panic if the input amount is greater than `2^128` when cast from a `U256` to `u128` .

### DIFF
--- a/engine-precompiles/src/native.rs
+++ b/engine-precompiles/src/native.rs
@@ -20,6 +20,7 @@ use evm::backend::Log;
 use evm::{Context, ExitError};
 
 const ERR_TARGET_TOKEN_NOT_FOUND: &str = "Target token not found";
+const MAX_U128: u128 = 340_282_366_920_938_463_463_374_607_431_768_211_455u128;
 
 mod costs {
     use crate::prelude::types::{EthGas, NearGas};
@@ -326,6 +327,10 @@ impl<I: IO> Precompile for ExitToNear<I> {
                 let amount = U256::from_big_endian(&input[..32]);
                 input = &input[32..];
 
+                if amount > MAX_U128.into() {
+                    return Err(ExitError::Other(Cow::from("ERR_INVALID_AMOUNT")));
+                }
+
                 if let Ok(receiver_account_id) = AccountId::try_from(input) {
                     (
                         nep141_address,
@@ -513,6 +518,10 @@ impl<I: IO> Precompile for ExitToEthereum<I> {
 
                 let amount = U256::from_big_endian(&input[..32]);
                 input = &input[32..];
+
+                if amount > MAX_U128.into() {
+                    return Err(ExitError::Other(Cow::from("ERR_INVALID_AMOUNT")));
+                }
 
                 if input.len() == 20 {
                     // Parse ethereum address in hex


engine-precompiles/src/native.rs
```diff
@@ -20,6 +20,7 @@ use evm::backend::Log;
 use evm::{Context, ExitError};
 
 const ERR_TARGET_TOKEN_NOT_FOUND: &str = "Target token not found";
+const MAX_U128: u128 = 340_282_366_920_938_463_463_374_607_431_768_211_455u128;
```
joshuajbouw, line 23: No need for this, you can use `u128::MAX`, which is already a constant.

engine-precompiles/src/native.rs
```diff
@@ -326,6 +327,10 @@ impl<I: IO> Precompile for ExitToNear<I> {
                 let amount = U256::from_big_endian(&input[..32]);
                 input = &input[32..];
 
+                if amount > MAX_U128.into() {
```
joshuajbouw, line 330:
```suggestion
                if amount > U256::from(u128::MAX) {
```
Something along those lines above. Avoid use of `.into`. This will be banned with checks in the future.

engine-precompiles/src/native.rs
```diff
@@ -514,6 +519,10 @@ impl<I: IO> Precompile for ExitToEthereum<I> {
...
```
joshuajbouw, line 522: See above.

joshuajbouw (changes requested):

engine-precompiles/src/native.rs
0x3bfc, line 522:
```suggestion
                if amount > U256::from(u128::MAX) {
```

joshuajbouw (approved):

joshuajbouw (changes requested):
If possible, add test that triggers these errors please inline.

aleksuss:
@0x3bfc for writing tests for this I'd suggest to create helper function like this: 
```rust
fn validate_amount(amount: U256) -> Result<(), ExitError> {
    if amount > U256::from(u128::MAX) {
        return Err(ExitError::Other(Cow::from("ERR_INVALID_AMOUNT")));
    }

    Ok(())
}
```
And then you'll be able to cover the function with tests.

engine-precompiles/src/native.rs
```diff
@@ -580,9 +588,19 @@ impl<I: IO> Precompile for ExitToEthereum<I> {
 
 #[cfg(test)]
 mod tests {
+    use aurora_engine_types::{Cow, U256};
+    use evm::ExitError;
+
     use super::{exit_to_ethereum, exit_to_near};
     use crate::prelude::sdk::types::near_account_to_evm_address;
 
+    fn validate_amount(amount: U256) -> Result<(), ExitError> {
```
aleksuss, line 597: But this function should be out of the test module. And it should replace checks on the 329 and 512 lines.

joshuajbouw (approved):
Nice use of `validate_amount` here. Thanks.

aleksuss (approved):

joshuajbouw:
Before I merge it, you need to sign all the commits with your @aurora.dev e-mail. Really important.

Aurora-ClickUp:
Task linked: [CU-863g0gc9w AUR-01  Reachable Panic Converting Balances to u128](https://app.clickup.com/t/863g0gc9w)

PR #684, opened 2023-02-09
fix: size checks on input array passed to `exitToNear` & `exitToEther`

0x3bfc:
## Description
There are no size checks implemented on input array passed to `exitToNear` and `exitToEthereum` precompiles, which may result in panic if accessing elements outside of array boundries.


engine-precompiles/src/native.rs
```diff
@@ -237,17 +244,26 @@ impl<I: IO> Precompile for ExitToNear<I> {
...
     ) -> EvmPrecompileResult {
+        // ETH transfer input format: (21+ bytes)
+        //  - flag (1 byte)
+        //  - refund_address (20 bytes)
+        //  - 1 byte ??
+        //  - recipient_account_id (N bytes)
+        // ERC20 transfer input format: (54+ bytes)
+        //  - flag (1 byte)
+        //  - refund_address (20 bytes)
+        //  - 1 byte ???
```
0x3bfc, line 255: I didn't get why in the case of `ExitToNear` there is an extra byte ?

cc @joshuajbouw, @birchmd Still @aleksuss

engine-precompiles/src/native.rs
```diff
@@ -216,6 +216,13 @@ impl<I> ExitToNear<I> {
     }
 }
 
+fn is_valid_input_size(input: &[u8], min: usize, max: Option<usize>) -> Result<(), ExitError> {
+    if max.is_some() && (input.len() < min || input.len() > max.unwrap()) {
```
aleksuss, line 220: Probably more rusty: 
```suggestion
    if input.len() < min || map.map_or(false, |max| input.len() > max)) {
```

aleksuss: And in your version you don't check `min` lenght if `max` is None. Add test case: 
```rust
#[test]
     fn test_check_valid_input() {
         let input = [1u8; 1];
         is_valid_input_size(&input, 2, None).unwrap();
     }
```
This test would pass.

aleksuss (changes requested):

engine-precompiles/src/native.rs
```diff
@@ -609,4 +633,26 @@ mod tests {
             super::events::EXIT_TO_ETH_SIGNATURE
         );
     }
+
+    #[test]
+    #[should_panic(expected = "ERR_INVALID_INPUT")]
+    fn test_check_invalid_input_lt_min() {
+        let input = [0u8; 4];
+        is_valid_input_size(&input, 10, Some(20)).unwrap();
+    }
```
aleksuss, line 642: Modify this test, please:
```rust
#[test]
fn test_check_invalid_input_lt_min() {
    let input = [0u8; 4];

    assert!(is_valid_input_size(&input, 10, Some(20)).is_err());
    assert!(is_valid_input_size(&input, 5, None).is_err());
}
```

engine-precompiles/src/native.rs
0x3bfc: it is failing here https://github.com/aurora-is-near/aurora-engine/actions/runs/4145399283/jobs/7169697666#step:9:375

engine-precompiles/src/native.rs
```diff
@@ -237,17 +244,26 @@ impl<I: IO> Precompile for ExitToNear<I> {
...
+        //  - refund_address (20 bytes)
+        //  - 1 byte ??
+        //  - recipient_account_id (N bytes)
+        // ERC20 transfer input format: (54+ bytes)
+        //  - flag (1 byte)
+        //  - refund_address (20 bytes)
+        //  - 1 byte ???
+        //  - amount (32 bytes)
+        //  - recipient_account_id (N bytes)
+        is_valid_input_size(input, 21, None)?;
```
aleksuss, line 258: It seems that this call should be in the body of `parse_input` function?

0x3bfc: Which one? There are 2 functions with the same name.

joshuajbouw:
Fix test please.

engine-precompiles/src/native.rs
```diff
@@ -237,17 +244,24 @@ impl<I: IO> Precompile for ExitToNear<I> {
...
+        //  - recipient_account_id (N bytes)
         #[cfg(feature = "error_refund")]
         fn parse_input(input: &[u8]) -> Result<(Address, &[u8]), ExitError> {
-            if input.len() < 21 {
-                return Err(ExitError::Other(Cow::from("ERR_INVALID_INPUT")));
-            }
-            let refund_address = Address::try_from_slice(&input[1..21]).unwrap();
-            Ok((refund_address, &input[21..]))
+            is_valid_input_size(input, 21, None)?;
+            let refund_address = Address::try_from_slice(&input.get(1..21).unwrap()).unwrap();
```
aleksuss, line 259: I'm just curious what's the motivation to replace indexing access with the `get` method? Anyway, you don't handle an Option the `get` method returns.

aleksuss: I think the `get` method is redundant here, and you can use indexing access safely because you make a validation check by the `validation_input_size` function.

engine-precompiles/src/native.rs
```diff
@@ -323,8 +337,8 @@ impl<I: IO> Precompile for ExitToNear<I> {
                 let erc20_address = context.caller;
                 let nep141_address = get_nep141_from_erc20(erc20_address.as_bytes(), &self.io)?;
 
-                let amount = U256::from_big_endian(&input[..32]);
-                input = &input[32..];
+                let amount = U256::from_big_endian(input.get(..32).unwrap());
+                input = input.get(32..).unwrap();
```
mrLSD, line 341: Do not use `unwrap()`. One of the Aurora Audit proposals was to avoid `unwrap`.

engine-precompiles/src/native.rs
```diff
@@ -462,8 +484,8 @@ impl<I: IO> Precompile for ExitToEthereum<I> {
         //      0x0 -> Eth transfer
         //      0x1 -> Erc20 transfer
         let mut input = input;
-        let flag = input[0];
-        input = &input[1..];
+        let flag = input.first().unwrap();
+        input = input.get(1..).unwrap();
```
mrLSD, line 488: Avoid `unwrap`

engine-precompiles/src/native.rs
joshuajbouw: Just a bit of guidance, if you are 100% sure that it can not possibly fail, `.expect("Some explanation")` is acceptable, otherwise, error return.

engine-precompiles/src/native.rs
joshuajbouw: We shouldn't be using unwrap, at all, anywhere ever. Production code should never have it period.

joshuajbouw (changes requested):

engine-precompiles/src/native.rs
```diff
@@ -444,6 +459,14 @@ impl<I: IO> Precompile for ExitToEthereum<I> {
...
         is_static: bool,
     ) -> EvmPrecompileResult {
+        // ETH transfer input format (min size 21 bytes)
+        //  - flag (1 byte)
+        //  - eth_recipient (20 bytes)
+        // ERC20 transfer input format: max 53 bytes
+        //  - flag (1 byte)
+        //  - amount (32 bytes)
+        //  - eth_recipient (20 bytes)
+        validate_input_size(input, 21, 117)?;
```
aleksuss, line 469: Why max is 117 rather than 53?

0x3bfc: This seems due to autosave wasn't enabled.

joshuajbouw:
Can you update this please?

engine-precompiles/src/native.rs
```diff
@@ -242,17 +249,25 @@ impl<I: IO> Precompile for ExitToNear<I> {
...
+        //  - refund_address (20 bytes)
+        //  - amount (32 bytes)
+        //  - recipient_account_id (max 64 bytes)
         #[cfg(feature = "error_refund")]
         fn parse_input(input: &[u8]) -> Result<(Address, &[u8]), ExitError> {
-            if input.len() < 21 {
-                return Err(ExitError::Other(Cow::from("ERR_INVALID_INPUT")));
-            }
+            validate_input_size(input, 21, 117)?;
             let refund_address = Address::try_from_slice(&input[1..21]).unwrap();
```
aleksuss, line 264:
```suggestion
            let refund_address = Address::try_from_slice(&input[1..21])?;
```

aleksuss: It should be like this in the end.

```rust
        #[cfg(feature = "error_refund")]
        fn parse_input(input: &[u8]) -> Result<(Address, &[u8]), ExitError> {
            validate_input_size(input, 21, 117)?;
            let mut buffer = [0; 20];
            buffer.copy_from_slice(&input[1..21]);
            let refund_address = Address::from_array(buffer);
            Ok((refund_address, &input[21..]))
        }
```

Aurora-ClickUp:
Task linked: [CU-863g0gctk AUR-02 Reachable Panic Due to Lack of Input Array Size Checks](https://app.clickup.com/t/863g0gctk)

Aurora-ClickUp:
Task linked: [CU-85zrqmdcd AUR-25 Lack of Input Array Size Checks in alt_bn256::read_point()](https://app.clickup.com/t/85zrqmdcd)

joshuajbouw (approved):

aleksuss (approved):

PR #686, opened 2023-02-09
Fix: Missing Gas Costs in exitToNear and exitToEthereum Precompiles

lempire123:
Calling the `exitToNear` and `exitToEthereum` precompiles will not add any gas costs as it will use a default value of 0 for the `cost` field.

### DIFF
--- a/engine-precompiles/src/native.rs
+++ b/engine-precompiles/src/native.rs
@@ -404,7 +404,7 @@ impl<I: IO> Precompile for ExitToNear<I> {
         Ok(PrecompileOutput {
             logs: vec![promise_log, exit_event_log],
             cost: gas_cost,
-            output: Default::default()
+            output: Default::default(),
         })
     }
 }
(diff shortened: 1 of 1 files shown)


engine-precompiles/src/native.rs
```diff
@@ -400,9 +400,11 @@ impl<I: IO> Precompile for ExitToNear<I> {
             data: exit_event_log.data,
         };
 
+        let gas_cost = Self::required_gas(input)?;
         Ok(PrecompileOutput {
             logs: vec![promise_log, exit_event_log],
-            ..Default::default()
+            cost: gas_cost,
+            output: Default::default(),
```
aleksuss, line 407: It's better to use a real type of field rather than `Default` here. Also, the clippy suggests doing so.

lempire123: Used `Vec::new()` , is that better?

aleksuss: Yep. Or `Vec::default()` absolutely fine. The point is that using a particular type helps you to understand which type is used when you read the code.

PR #687, opened 2023-02-14
Fix: Missing Gas Costs in exitToNear and exitToEthereum Precompiles

lempire123:
Calling the `exitToNear` and `exitToEthereum` precompiles will not add any gas costs as it will use a default value of 0 for the `cost` field.

### DIFF
--- a/engine-precompiles/src/native.rs
+++ b/engine-precompiles/src/native.rs
@@ -399,10 +399,12 @@ impl<I: IO> Precompile for ExitToNear<I> {
             topics: exit_event_log.topics,
             data: exit_event_log.data,
         };
-
+        
+        let gas_cost = Self::required_gas(input)?;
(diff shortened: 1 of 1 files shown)


joshuajbouw (approved):
This attentively can be approved but, would require 2nd opinion from @birchmd

birchmd (approved):
As pointed out in the description, this does not change anything because `Self::required_gas` returns a constant value equal to 0. But at least the semantics are now correct such that changing the value of the gas constants will have the intended effect.

Aurora-ClickUp:
Task linked: [CU-863g0gjgh AUR-16 Missing Gas Costs in exitToNear and exitToEthereum Precompiles](https://app.clickup.com/t/863g0gjgh)

joshuajbouw:
Fix lints please.

PR #688, opened 2023-02-14
fix: reachable panic due to arithmetic overflow

0x3bfc:
## Description

This PR covers two correlated issues:

- [x] **Issue-1**

It should check if the `base length` and `modulus length` are `zero` to return successfully with `empty data` before handle any error.

- [x] **Issue-2** 

The `modExp` precompile may panic if the exponent length is greater than `usize::MAX - 96` when calling to retrieve the exponent length from the input. To trigger this panic the base length and modulus length must be `zero` due to gas requirements and the exponent length must be greater than `usize::MAX - 96` , where `usize` is `4 bytes` for the `wasm32` platform.

Regarding issue#2, the check is no longer needed because it has to happen when base length and mod length are zeros which is resolved in issue-1.


engine-precompiles/src/modexp.rs
```diff
@@ -22,6 +22,9 @@ impl<HF: HardFork> ModExp<HF> {
     fn calc_iter_count(exp_len: u64, base_len: u64, bytes: &[u8]) -> Result<U256, ExitError> {
         let start = usize::try_from(base_len).map_err(utils::err_usize_conv)?;
         let exp_len = usize::try_from(exp_len).map_err(utils::err_usize_conv)?;
+        if exp_len > usize::MAX - 96 {
```
aleksuss, line 25: It seems it would be good to introduce a new constant `const EXP_LEN: usize = usize::MAX - 96;`

aleksuss: And leave a comment about what a magic 96 is.

0x3bfc: This PR will be merged into #689 in one PR.

engine-precompiles/src/modexp.rs
```diff
@@ -507,4 +513,52 @@ mod tests {
...
+    fn test_modexp_not_overflow() {
+        let base_len = U256::from(0);
+        let exp_len = U256::from(4294967200usize); // `usize::MAX` - 95
+        let mod_len = U256::from(0);
+        let base = U256::MAX;
+        let exp = U256::MAX;
+        let modulus = U256::MAX;
+
+        let mut input: Vec<u8> = Vec::new();
+        input.extend_from_slice(&u256_to_arr(&base_len));
```
aleksuss, line 527: You can use here `.extend` method.
```suggestion
        input.extend(u256_to_arr(&base_len));
```

engine-precompiles/src/modexp.rs
```diff
@@ -142,13 +145,16 @@ impl ModExp<Berlin> {
 impl Precompile for ModExp<Berlin> {
     fn required_gas(input: &[u8]) -> Result<EthGas, ExitError> {
         let (base_len, exp_len, mod_len) = parse_lengths(input);
+        if base_len == 0 && mod_len == 0 {
+            Ok(EthGas::new(200u64)) // min gas 200
```
aleksuss, line 149: It's better to use constants in such cases rather than magic numbers. 
```rust
const MIN_GAS: u64 = 200;
```

aleksuss: And to do it in every `impl` block for `ModExp<_>` struct. E.g.
```rust
impl ModExp<Byzantium> {
    const MIN_GAS: u64 = 0;

    // output of this function is bounded by 2^128
    fn mul_complexity(x: u64) -> U256 {
        if x <= 64 {
            U256::from(x * x)
        } else if x <= 1_024 {
            U256::from(x * x / 4 + 96 * x - 3_072)
        } else {
            // up-cast to avoid overflow
            let x = U256::from(x);
            let x_sq = x * x; // x < 2^64 => x*x < 2^128 < 2^256 (no overflow)
            x_sq / U256::from(16) + U256::from(480) * x - U256::from(199_680)
        }
    }
}

impl ModExp<Berlin> {
    const MIN_GAS: u64 = 200;

    // output bounded by 2^122
    fn mul_complexity(base_len: u64, mod_len: u64) -> U256 {
        let max_len = core::cmp::max(mod_len, base_len);
        let words = U256::from(Integer::div_ceil(&max_len, &8));
        words * words
    }
}
```

0x3bfc: But it will not be found in the scope of the `Precompile` impl

aleksuss: It will be
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5dce524e9c06b126e455296afbfba024

aleksuss:
```rust
Ok(EthGas::new(Self::MIN_GAS))
```

aleksuss: The constant could be `EthGas` since the `new` method of `EthGas` is const.
```rust
const MIN_GAS: EthGas = EthGas::new(200);
```

birchmd (approved):

engine-precompiles/src/modexp.rs
```diff
@@ -507,4 +515,52 @@ mod tests {
         );
         assert_eq!(Err(ExitError::OutOfGas), res);
     }
+
+    #[test]
+    fn test_modexp_not_overflow() {
+        let base_len = U256::from(0);
+        let exp_len = U256::from(4294967200usize); // `usize::MAX` - 95
```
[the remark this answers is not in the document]
0x3bfc: Just found that using `usize::MAX - 95` in the tests it leads `attempt to add with overflow` error in this [line](https://github.com/aurora-is-near/aurora-engine/blob/f7bd2b9ea9641990c1bd737b53c8a8bbbe235b3a/engine-precompiles/src/modexp.rs#L186) and which is WIP in another PR.

aleksuss (approved):

joshuajbouw (approved):

Aurora-ClickUp:
Task linked: [CU-863g0ge4a AUR-06 Reachable Panic Due to Arithmetic Overflow](https://app.clickup.com/t/863g0ge4a)

Aurora-ClickUp:
Task linked: [CU-863g0gfmh AUR-13 Inconsistencies With Other EVM Clients](https://app.clickup.com/t/863g0gfmh)

PR #689, opened 2023-02-15
fix: reachable panic due to out-of-memory in `modexp` precompile

0x3bfc:
## Description

The precompile may panic if the exponent length is greater than the allocable space when calling `parse_bytes()` to retrieve the exponent length from the input. To trigger this panic the base length and modulus length must be `zero` due to gas requirements. 
- If the exponent length is greater than `isize::MAX` (2^31−1 for the wasm32 platform), the `modexp` precompile will panic with `capacity overflow`.  
- If the exponent length is less or equal `isize::MAX` and there is not enough allocable space, the `modexp` precompile will trigger a panic.


Aurora-ClickUp:
Task linked: [CU-863g0ge8z AUR-07 Reachable Panic Due to Out-of-Memory](https://app.clickup.com/t/863g0ge8z)

engine-precompiles/src/modexp.rs
```diff
@@ -1,11 +1,29 @@
...
+        let len = $input.len();
+        let from_zero = min($from, len);
+        let from = min(from_zero + SPLIT, len);
+        let to = min($to, len);
+        let overflow_bytes = &$input[from_zero..from];
+
+        let mut len_bytes = [0u8; size_of::<u64>()];
+        len_bytes[..to - from].copy_from_slice(&$input[from..to]);
+        #[allow(clippy::as_conversions)]
+        let out = u64::from_be_bytes(len_bytes) as usize;
```
birchmd, line 21: This conversion is not safe in Wasm because `usize` there is 32 bit. I also don't think we should be taking macros from other codebases.

birchmd (changes requested):
This PR currently does not address the issue. The problem is that we eagerly allocate memory based on the lengths in the input (this happens at `engine-precompiles/src/modexp.rs:70`, but it's outside the diff so I can't put the comment there). I think our parsing of the lengths was fine, it's what we do with those lengths is currently a problem.

engine-precompiles/src/modexp.rs
0x3bfc: Thanks for pointing out to the actual issue. I will try to reproduce it with a fix shortly.

engine-precompiles/src/modexp.rs
```diff
@@ -42,22 +44,27 @@ impl<HF: HardFork> ModExp<HF> {
...
-        let (base_len, exp_len, mod_len) = parse_lengths(input);
+        let (base_len, exp_len, mod_len) = parse_lengths(input)?;
         let base_len = usize::try_from(base_len).map_err(utils::err_usize_conv)?;
         let exp_len = usize::try_from(exp_len).map_err(utils::err_usize_conv)?;
         let mod_len = usize::try_from(mod_len).map_err(utils::err_usize_conv)?;
 
         let base_start = 96;
+
+        if base_len == 0 && mod_len == 0 && exp_len > usize::MAX.saturating_sub(base_start) {
+            return Err(ExitError::Other(Cow::Borrowed("INVALID_MODEXP_LENGTHS")));
```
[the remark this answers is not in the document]
0x3bfc: Interesting I didn't know that. I will clean it up.

engine-precompiles/src/modexp.rs
```diff
@@ -179,12 +186,21 @@ impl Precompile for ModExp<Berlin> {
...
     let len = input.len();
     if start >= len {
-        return f(&[]);
+        return Ok(f(&[]));
     }
     let end = start + size;
+
+    if end > usize::MAX.saturating_sub(96) {
+        return Err(ExitError::Other(Cow::Borrowed("INVALID_LENGTH")));
+    }
```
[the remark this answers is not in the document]
0x3bfc: Yes, that's make sense. What I understood that the main cause for the capacity overflow and the out-of-memory issues was the padding with zeros in parse_bytes, that's why I added this check. Given that we shouldn't return errors through the precompiler as you mentioned, that function (parse_bytes) won't be safe to use without an extra check in the calling function. Anyway I will update it to follow the example you mentioned

engine-precompiles/src/modexp.rs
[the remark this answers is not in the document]
birchmd: Thanks! Please be sure to double check behaviour against geth (go-ethereum client). You can run against their EVM implementation directly by compiling the `evm` binary from their source code. For example you can see me using it to verify the behaviour of the ecrecover precompile on [a previous PR](https://github.com/aurora-is-near/aurora-engine/pull/717#pullrequestreview-1328633657).

birchmd (commented):
Nice! I like how simple this fix is, and I do think that this fixes the problem 🚀 

Just some minor comments on the tests.

engine-precompiles/src/modexp.rs
```diff
@@ -519,20 +529,14 @@ mod tests {
...
-        let mut input: Vec<u8> = Vec::new();
-        input.extend(u256_to_arr(&base_len));
-        input.extend(u256_to_arr(&exp_len));
-        input.extend(u256_to_arr(&mod_len));
-        input.extend(u256_to_arr(&base));
-        input.extend(u256_to_arr(&exp));
-        input.extend(u256_to_arr(&modulus));
+        let input = generate_modexp_test_input(
+            U256::from(0),
+            U256::from(4294967200usize), // `usize::MAX` - 95
```
birchmd, line 534:
```suggestion
            U256::from(usize::MAX - 95),
```

engine-precompiles/src/modexp.rs
```diff
@@ -459,38 +479,28 @@ mod tests {
...
-        let mut input: Vec<u8> = Vec::new();
-        input.extend(u256_to_arr(&base_len));
-        input.extend(u256_to_arr(&exp_len));
-        input.extend(u256_to_arr(&mod_len));
-        input.extend(base.to_be_bytes());
-        input.extend(u256_to_arr(&exp));
-
+        let input = generate_modexp_test_input(
+            U256::from(4),
+            U256::from(u64::MAX), // `usize::MAX` - 95
```
birchmd, line 484:
```suggestion
            U256::from(u64::MAX),
```

engine-precompiles/src/modexp.rs
```diff
@@ -459,38 +479,28 @@ mod tests {
...
-        let mut input: Vec<u8> = Vec::new();
-        input.extend(u256_to_arr(&base_len));
-        input.extend(u256_to_arr(&exp_len));
-        input.extend(u256_to_arr(&mod_len));
-        input.extend(base.to_be_bytes());
-        input.extend(u256_to_arr(&exp));
-
+        let input = generate_modexp_test_input(
+            U256::MAX,
+            U256::MAX, // `usize::MAX` - 95
```
birchmd, line 498:
```suggestion
            U256::MAX,
```

engine-precompiles/src/modexp.rs
```diff
@@ -557,20 +561,14 @@ mod tests {
...
-        let mut input: Vec<u8> = Vec::new();
-        input.extend(u256_to_arr(&base_len));
-        input.extend(u256_to_arr(&exp_len));
-        input.extend(u256_to_arr(&mod_len));
-        input.extend(u256_to_arr(&base));
-        input.extend(u256_to_arr(&exp));
-        input.extend(u256_to_arr(&modulus));
+        let input = generate_modexp_test_input(
+            U256::from(0),
+            U256::from(1), // `isize::MAX`
```
birchmd, line 566:
```suggestion
            U256::from(1),
```

engine-precompiles/src/modexp.rs
```diff
@@ -592,4 +590,124 @@ mod tests {
...
+            ModExp::<Berlin>::new().run(&input, Some(EthGas::new(100_000)), &new_context(), false);
+        assert!(res.is_ok());
+    }
+
+    #[test]
+    fn test_modexp_add_to_overflow() {
+        // This test should not panicing with capacity overflow
+        let input = generate_modexp_test_input(
+            U256::from(0),
+            U256::from(usize::MAX), // `isize::MAX`
```
birchmd, line 703:
```suggestion
            U256::from(usize::MAX),
```

engine-tests/src/benches/res/StandardPrecompiles.sol
```diff
@@ -150,12 +175,19 @@ contract StandardPrecompiles {
     return output;
   }
 
-  function modexp(uint256 base, uint256 exponent, uint256 modulus) private view returns (uint256 output) {
+  function modexp(
```
birchmd, line 178: I don't think this function has been modified quite right. I think probably the second argument to `add` in the various `mstore` calls also needs to be modified to account for the changed length.

But rather than mess around with pointer arithmetic in Solidity, I think it would be better to add the new integration tests as separate Rust tests. These tests could call the precompile directly with some constructed input similarly to how the [ecrecover tests](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-tests/src/tests/ecrecover.rs) are done.

birchmd (changes requested):

birchmd (approved):
LGTM, nice work @0x3bfc

engine-precompiles/src/modexp.rs
```diff
@@ -229,6 +231,24 @@ mod tests {
...
 
+    fn generate_modexp_test_input(
+        base_len: U256,
+        exp_len: U256,
+        mod_len: U256,
+        base: U256,
+        exp: U256,
+        modulus: U256,
+    ) -> Vec<u8> {
+        let mut input: Vec<u8> = Vec::new();
```
aleksuss, line 242: Could be written shorter: 
```rust
[
     u256_to_arr(&base_len),
     u256_to_arr(&exp_len),
     u256_to_arr(&mod_len),
     u256_to_arr(&base),
     u256_to_arr(&exp),
     u256_to_arr(&modulus)
]
.concat()
```

aleksuss (approved):
Overall looks good 👍🏻

hskang9 (commented):
Have you though concatenating `base` with specific endian encoding method?

engine-precompiles/src/modexp.rs
```diff
@@ -229,6 +231,25 @@ mod tests {
...
+        mod_len: U256,
+        base: U256,
+        exp: U256,
+        modulus: U256,
+    ) -> Vec<u8> {
+        [
+            u256_to_arr(&base_len),
+            u256_to_arr(&exp_len),
+            u256_to_arr(&mod_len),
+            u256_to_arr(&base),
```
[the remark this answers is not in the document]
birchmd: [u256_to_arr does use big endian](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-types/src/types/wei.rs#L214).

hskang9 (approved):

engine-precompiles/src/modexp.rs
```diff
@@ -56,13 +56,15 @@ impl<HF: HardFork> ModExp<HF> {
         let mod_start = exp_end;
 
         let base = parse_bytes(input, base_start, base_len, BigUint::from_bytes_be);
```
mandreyel, line 58: I probably don't fully understand the vulnerability but can the same issue occur if only the base is provided but not the exponent and modulus? Should parsing the base be moved into the else branch as well?

engine-precompiles/src/modexp.rs
```diff
@@ -459,38 +480,28 @@ mod tests {
...
-        let exp = U256::MAX;
-
-        let mut input: Vec<u8> = Vec::new();
-        input.extend(u256_to_arr(&base_len));
-        input.extend(u256_to_arr(&exp_len));
-        input.extend(u256_to_arr(&mod_len));
-        input.extend(base.to_be_bytes());
-        input.extend(u256_to_arr(&exp));
-
+        let input = generate_modexp_test_input(
```
mandreyel, line 483: I think passing these argus as a struct would make it cleaner, or maybe even declarative macro where the arguments are key-prefixed. E.g.:
```rust
let input = generate_modexp_test_input(ModexpInput {
    base_len: U256::from(0x80),
    exp_len: U256::from(0x80),
    mod_len: U256::from(0x80),
    base: U256::from(0x80),
    exp: U256::from(0x80),
    modulus: U256::from(0x80),
});
```
or:
```rust
let input = modexp_test_input {
    base_len: U256::from(0x80),
    exp_len: U256::from(0x80),
    mod_len: U256::from(0x80),
    base: U256::from(0x80),
    exp: U256::from(0x80),
    modulus: U256::from(0x80),
};
```

engine-precompiles/src/modexp.rs
```diff
@@ -592,4 +591,124 @@ mod tests {
...
+            &input.unwrap(),
+            Some(EthGas::new(100_000)),
+            &new_context(),
+            false,
+        );
+        assert_eq!(Err(ExitError::OutOfGas), res);
+    }
+
+    #[test]
+    fn test_max_exp_zero_base_non_zero_mod() {
```
mandreyel, line 656: If my above question is valid regarding max base, no exp, no mod, then I think that would need a test case as well.

mandreyel (changes requested):
My understanding of this part of the code is very limited so please take my questions more as my trying to learn.

engine-precompiles/src/modexp.rs
0x3bfc: I think It is nice to have, however it is all about test inputs not sure why we would need to have a macro for it.

engine-precompiles/src/modexp.rs
0x3bfc: The scope of this PR is to solve the out-of-memory issue in case of passing a very big exponent. For the base and modulus are not the case as it will revert with `out-of-gas` in the `required_gas()` before executing `runn_inner()`

birchmd: Thanks for you vigilance on this @mandreyel ! I think it is ok to eagerly parse the `base` value because it is accounted for in the gas cost of the precompile. I'll try to explain this in more detail below.

First of all, the basic principle public blockchains that have smart contracts operate on is the following: users can do arbitrary computations (thus consuming arbitrary resources on the nodes of the blockchain network) so long as the user pays for them. The metric used to measure how much the user has to pay is called "gas". The idea is that operations which would crash a node (e.g. allocating terabytes of memory) are prohibitively expensive in terms of gas and therefore no user can cause those operations to happen because they will lack the funds to do it. [As an aside, our particular project is extra complicated in this regard because we are running the EVM, which has its own gas system, inside the Near runtime, which also has a gas system; so we need to make sure that operations are adequately charged in both systems, but I won't go into details on that complexity here]

Given this background, take a look at the [required_gas function](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-precompiles/src/modexp.rs#L149) for this precompile. The key point is that it has a special case for when the base and modulus are both zero, independent of the value of the exponent. So input with no base or modulus is very cheap in terms of gas, but our implementation did not have the same short-circuit, thus allowing malicious users to perform operations they were not adequately charged for. This PR fixes the issue by only parsing the exponent if it is needed for a calculation.

To directly address your question about the base, notice that the [gas cost for the base is quadratic in its size](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-precompiles/src/modexp.rs#L141-L146). Therefore, users can input very large base values, but they will also be charged very large amounts or else such input will be rejected.

engine-precompiles/src/modexp.rs
birchmd: +1 for struct as input.

I personally try to avoid macros, except where they save lots of code duplication, because the IDE support for them is not as good as normal rust syntax.

mandreyel: Fair enough, struct inputs it is then.

@0x3bfc, well, tests are first class citizens so the same code hygiene should apply as with normal code. In this case the reasoning is that given all six arguments are of the same type, there is nothing stopping one from accidentally switching up some of the values--future contributors could make that mistake. With the struct, however, it's very clear what purpose a value corresponds to. As a rule of thumb, in Rust one should rely on the type system as much as possible.

The struct based approach also makes it easier to read what's going on for those who aren't intimately familiar with the code base and to get an idea what the tests are doing (which is important because tests are a form of live documentation).

engine-precompiles/src/modexp.rs
mandreyel: That makes sense, thank you for the in-depth explanation @birchmd. Resolving.

PR #690, opened 2023-02-15
Add set_owner method for Aurora DAO

hskang9:
## Description

Aurora DAO requires an owner shift from the contract creator to Aurora DAO contract to execute its admin functions in the future. To achieve this, previous owner have to set the owner address from before to a new address that the previous owner sets. 

`set_owner` method checks whether the sender is marked as the owner of the engine contract's state.

Then converts string input into AccountId. 

It changes `owner` property in Aurora engine's contract state. 

## Testing

Tests were done in https://github.com/hskang9/aurora-silo-tests/blob/main/src/test.rs#L156. 

I am concerned if converting a random string into NEAR account id is still safe.


engine/src/lib.rs
```diff
@@ -142,6 +144,20 @@ mod contract {
...
+    pub extern "C" fn set_owner() {
+        let mut io = Runtime;
+        let mut state = state::get_state(&io).sdk_unwrap();
+        require_owner_only(&state, &io.predecessor_account_id());
+        let args = io.read_input().to_vec();
+        let sr: String = String::from_utf8(args).unwrap();
+        let id = AccountId::from_str(&sr).unwrap();
+        state.owner_id = id.clone();
+        state::set_state(&mut io, state).sdk_unwrap();
+        io.return_output(id.as_bytes());
```
birchmd, line 158: Rather than returning the input again, I think a simple `bool` would be fine (`true` for updating the value, `false` otherwise). This would avoid the clone when setting `state.owner_id`. In this scheme it also makes sense to check whether the new value differs from the old value so that `false` is actually returned sometimes.

engine/src/lib.rs
```diff
@@ -142,6 +144,20 @@ mod contract {
         io.return_output(state.owner_id.as_bytes());
     }
 
+    /// Set owner account id for this contract.
+    #[no_mangle]
+    pub extern "C" fn set_owner() {
```
birchmd, line 149: This is a new public method, so it requires changes to the Borealis Engine code as well. Please add a new transaction kind [here](https://github.com/aurora-is-near/aurora-engine/blob/master/engine-standalone-storage/src/sync/types.rs#L76) and follow the compilation errors that arise (unexhaustive match) to finish the implementation.

engine/src/lib.rs
```diff
@@ -142,6 +144,20 @@ mod contract {
...
+        let mut io = Runtime;
+        let mut state = state::get_state(&io).sdk_unwrap();
+        require_owner_only(&state, &io.predecessor_account_id());
+        let args = io.read_input().to_vec();
+        let sr: String = String::from_utf8(args).unwrap();
+        let id = AccountId::from_str(&sr).unwrap();
```
birchmd, line 155: Please avoid using the Rust unwrap. It is more efficient to use Near's host functions to abort the computation (see https://docs.near.org/sdk/rust/best-practices#stdpanic-vs-envpanic ). I would implement this as something like

```rust
let account_str = String::from_utf8(args).unwrap_or_else(|_| aurora_engine_sdk::panic_utf8(b"ERR_INVALID_STRING"));
let id: AccountId = account_str.try_into().sdk_unwrap();
```

birchmd (changes requested):
I think it would also be nice to add a test for calling this method (one which runs both through the Wasm contract and the standalone engine).

engine/src/lib.rs
hskang9: TransactionKind is now added, borsh inputs are added and tested with workspace. https://github.com/hskang9/aurora-silo-tests/blob/05f05505d9c362c3b2648bfa7647474689014271/src/test.rs#L154

engine/src/lib.rs
hskang9: changing bool into bytes with `as_bytes` methods was not successful, so I made the method to return output as `&[1]`. I am not sure this will make sense though as `return_output` may act different than I expected. when I looked at the `SuccessValue` at workspace [test](https://github.com/hskang9/aurora-silo-tests/blob/11b5bb038d0c6c9422887ad546f355a8fa1033ea/src/test.rs#L156), it was returning nothing in SuccessValue.

```
Set Owner result: ExecutionFinalResult { total_gas_burnt: 5571972969272, transaction: ExecutionOutcome { block_hash: `ASrDKAigw1xeo4YmZCBuoP6M9zAgX11dNoEvR43VBS4j`, logs: [], receipt_ids: [`9ABq9UjKNjEgCAs1om4dHUsc86ivQMCvkzZ6U4JVEX51`], gas_burnt: 2427990313954, tokens_burnt: 242799031395400000000, executor_id: AccountId("owner.test.near"), status: SuccessReceiptId(9ABq9UjKNjEgCAs1om4dHUsc86ivQMCvkzZ6U4JVEX51) }, receipts: [ExecutionOutcome { block_hash: `ASrDKAigw1xeo4YmZCBuoP6M9zAgX11dNoEvR43VBS4j`, logs: [], receipt_ids: [`GRi9jNEWu9CU8kow4zKBLJJR1v1nRgcR5irhcr2gDa18`], gas_burnt: 2920800092818, tokens_burnt: 292080009281800000000, executor_id: AccountId("owner.test.near"), status: SuccessValue(``) }, ExecutionOutcome { block_hash: `GtjXTQbjfqdM7Lfhm2eaF8bz3tngBkNsf6kawW9pHHhJ`, logs: [], receipt_ids: [], gas_burnt: 223182562500, tokens_burnt: 0, executor_id: AccountId("owner.test.near"), status: SuccessValue(``) }], status: SuccessValue(``) }
```

hskang9: It would be helpful if I could find a way to display true in the return, should I put ascii bytes in "true" ?

aleksuss: let data = true.try_to_vec().unwrap();
io.return_output(&data[..]);

hskang9: Ok thank you.

hskang9: Will indexing logic handle "true" as boolean value from result? I am not sure how this will go as I do not know the codec encode/decode of this.

hskang9: If indexing this tx result is unnecessary, I would just pass "true" as bytes. I need to confirm whether if this is ok.

engine-standalone-storage/src/sync/types.rs
```diff
@@ -493,6 +496,7 @@ enum BorshableTransactionKind<'a> {
     StorageUnregister(Option<bool>),
     StorageWithdraw(Cow<'a, parameters::StorageWithdrawCallArgs>),
     SetPausedFlags(Cow<'a, parameters::PauseEthConnectorCallArgs>),
+    SetOwner(Cow<'a, parameters::SetOwnerArgs>),
```
birchmd, line 499: This new variant needs to be put at the bottom of the enum (after `Unknown`). The reason is because the borsh serialization is sensitive to the order of the variants as listed in the source code. By adding a new variant in the middle you will shift the variant selection byte in the serialization causing it to fail to deserialize old data.

The order of variants in `TransactionKind` do not matter because it does not implement borsh serialization, but it is very important to put new transaction kinds at the bottom of `BorshableTransactionKind`.

engine/src/lib.rs
```diff
@@ -142,6 +143,19 @@ mod contract {
         io.return_output(state.owner_id.as_bytes());
     }
 
+    /// Set owner account id for this contract.
+    #[no_mangle]
+    pub extern "C" fn set_owner() {
+        let mut io = Runtime;
+        let mut state = state::get_state(&io).sdk_unwrap();
+        require_owner_only(&state, &io.predecessor_account_id());
+        let args: SetOwnerArgs = io.read_input_borsh().sdk_unwrap();
```
birchmd, line 152: It is worth thinking about how we want the input and output to be encoding. In my opinion, the main factor to consider is what other tooling will be used with this `set_owner` function. For example, JSON encoding is standard in the Near ecosystem, so if this method will be called from Near CLI (even if that might be indirect by passing arguments to sputnik DAO) then it might be easier for users if the input and output were encoded with JSON instead of borsh.

engine/src/lib.rs
```diff
@@ -142,6 +143,19 @@ mod contract {
...
+        state.owner_id = args.new_owner;
+        state::set_state(&mut io, state).sdk_unwrap();
```
birchmd, line 154: I still think we should check if the owner is really new or not.

```rust
if state.owner_id == args.new_owner {
    // Would be a no-op to set, do nothing and return false
    io.return_output(b"false");
} else {
    state.owner_id = args.new_owner;
    state::set_state(&mut io, state).sdk_unwrap();
    io.return_output(b"true");
}
```

birchmd (changes requested):

engine/src/lib.rs
birchmd: Added a comment about IO encoding: https://github.com/aurora-is-near/aurora-engine/pull/690#discussion_r1108502063

engine-standalone-storage/src/lib.rs
```diff
@@ -37,6 +37,21 @@ pub enum StoragePrefix {
     EngineAccountId = 0x07,
 }
 
+impl From<StoragePrefix> for u8 {
```
aleksuss, line 40: Could you let me know why you added this implementation? The annotation `#[repr(u8)]` allows you to convert any variant of `StoragePrefix` into `u8` type. 

```rust
let block_hash: u8 = StoragePrefix::BlockHash as u8;
```

engine-standalone-storage/src/lib.rs
```diff
@@ -463,7 +478,7 @@ impl BlockMetadata {
 }
 
 fn construct_storage_key(prefix: StoragePrefix, key: &[u8]) -> Vec<u8> {
-    [&[VERSION], &[prefix as u8], key].concat()
+    [&[VERSION], &[u8::from(prefix)], key].concat()
```
aleksuss, line 481: And again, why?

engine-standalone-storage/src/relayer_db/types.rs
```diff
@@ -146,12 +186,41 @@ pub struct TransactionRow {
     pub output: Vec<u8>,
 }
 
+struct TransactionRowBlock(pub i64);
+
+impl From<TransactionRowBlock> for u64 {
+    fn from(value: TransactionRowBlock) -> Self {
+        // set negative values to 0
+        let res: u64 = value.0.max(0).try_into().unwrap();
```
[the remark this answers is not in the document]
aleksuss: In the case of `value.0` is negative, you get an error.

engine/src/lib.rs
```diff
@@ -142,6 +143,25 @@ mod contract {
...
+        if state.owner_id == args.new_owner {
+            // Would be a no-op to set, do nothing and return false
+            io.return_output(b"false");
+        } else {
+            state.owner_id = args.new_owner;
+            state::set_state(&mut io, state).sdk_unwrap();
+            io.return_output(b"true");
+        }
+        // return true as bytes
+        io.return_output(b"true");
```
aleksuss, line 162: Why? With this output, you will get `falsetrue` or `truetrue`.

joshuajbouw: That, and we normally don't return on `set_` unless there is an error. So on false, return error. On true, return nothing. There doesn't need to be a return in case of true.

engine-standalone-storage/src/relayer_db/types.rs
```diff
@@ -68,14 +68,54 @@ pub struct BlockRow {
     pub receipts_root: H256,
 }
 
+struct BlockRowSize(i32);
+
+impl From<BlockRowSize> for u32 {
+    fn from(value: BlockRowSize) -> Self {
+        // set negative values to 0
+        let res: u32 = value.0.max(0).try_into().unwrap();
```
joshuajbouw, line 76: These should be `try_from`

engine-standalone-storage/src/relayer_db/types.rs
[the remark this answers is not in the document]
joshuajbouw: `unwrap_or_default` would not be ideal. Can lead to unexpected errors. Implementing `TryFrom` is better. `unwrap` anywhere in production code is unacceptable.

joshuajbouw (changes requested):

engine-standalone-storage/src/relayer_db/types.rs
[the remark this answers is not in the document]
aleksuss: Of course, it depends on the business logic. If a negative value is unacceptable, we should return an error here. In this case, you need to implement the `TryFrom` trait rather than just `From`.

hskang9:
I tried to remove `unwrap()` as far as possible by aggregating both `postres::Error` and `TryFromIntError`, but they happen to collide on `RowIter` type's map function.

hskang9:
@birchmd to add standalone tests, I think we need to first publish new parameter type with `SetAdminArgs` as the test code imports parameter from published `aurora_engine` crate? https://github.com/aurora-is-near/aurora-engine/blob/13eb7089c3cb1facec94ee6acfb6114aad83ef87/engine-tests/src/tests/state_migration.rs#L3 or is it just fine to import from crate?

engine-standalone-storage/src/relayer_db/types.rs
```diff
@@ -68,35 +69,81 @@ pub struct BlockRow {
...
         let parent_hash = get_hash(&row, "parent_hash");
         let transactions_root = get_hash(&row, "transactions_root");
         let state_root = get_hash(&row, "state_root");
         let receipts_root = get_hash(&row, "receipts_root");
 
-        Self {
-            chain: chain as u64,
-            id: id as u64,
+        Ok(Self {
+            chain: u64::try_from(chain).unwrap(),
```
aleksuss, line 134: You can change the type of Error for `TryFromIntError` and then use error propagation (`?`) here.

engine-standalone-storage/src/relayer_db/types.rs
```diff
@@ -68,35 +69,81 @@ pub struct BlockRow {
...
+    type Error = TryFromIntError;
+
+    fn try_from(value: BlockRowBlock) -> Result<Self, Self::Error> {
+        // set negative values to 0
+        value.0.max(0).try_into()
+    }
+}
+
+impl TryFrom<postgres::Row> for BlockRow {
+    type Error = postgres::Error;
```
aleksuss, line 117:
```suggestion
    type Error = TryFromIntError;
```

engine-standalone-storage/src/relayer_db/types.rs
```diff
@@ -146,12 +193,47 @@ pub struct TransactionRow {
...
-    fn from(row: postgres::Row) -> Self {
-        let block: i64 = row.get("block");
+struct TransactionRowBlock(pub i64);
+
+impl TryFrom<TransactionRowBlock> for u64 {
+    type Error = TryFromIntError;
+
+    fn try_from(value: TransactionRowBlock) -> Result<Self, Self::Error> {
+        // set negative values to 0
+        value.0.max(0).try_into()
```
aleksuss, line 203: Are you sure it is ok to set the negative value to 0? Is it not better to return an error in case of a negative value?
```suggestion
        value.0.try_into()
```

engine-standalone-storage/src/relayer_db/types.rs
```diff
@@ -146,12 +193,47 @@ pub struct TransactionRow {
...
+}
+
+struct TransactionRowIndex(pub i32);
+
+impl TryFrom<TransactionRowIndex> for u16 {
+    type Error = TryFromIntError;
+
+    fn try_from(value: TransactionRowIndex) -> Result<Self, Self::Error> {
+        // set Maximum value to u16::MAX
+        value.0.min(u16::MAX.into()).try_into()
```
aleksuss, line 214: Same here.

engine-standalone-storage/src/relayer_db/types.rs
```diff
@@ -146,12 +193,47 @@ pub struct TransactionRow {
...
+}
+
+struct TransactionRowId(pub i64);
+
+impl TryFrom<TransactionRowId> for u64 {
+    type Error = TryFromIntError;
+
+    fn try_from(value: TransactionRowId) -> Result<Self, Self::Error> {
+        // set negative values to 0
+        value.0.max(0).try_into()
```
aleksuss, line 225: And here.

engine-standalone-storage/src/relayer_db/types.rs
```diff
@@ -146,12 +193,47 @@ pub struct TransactionRow {
...
+
+impl TryFrom<postgres::Row> for TransactionRow {
+    type Error = postgres::Error;
+
+    fn try_from(row: postgres::Row) -> Result<Self, Self::Error> {
+        let block: TransactionRowBlock = TransactionRowBlock(row.get("block"));
         let block_hash = get_hash(&row, "block_hash");
-        let index: i32 = row.get("index");
-        let id: i64 = row.get("id");
+        let index: TransactionRowIndex = TransactionRowIndex(row.get("index"));
```
aleksuss, line 235: It's needless to define explicit type here because you use it on the right side.
```suggestion
        let index = TransactionRowIndex(row.get("index"));
```

engine-standalone-storage/src/relayer_db/types.rs
```diff
@@ -229,11 +311,25 @@ fn get_address(row: &postgres::Row, field: &str) -> Address {
...
+
+impl TryFrom<TransactionDuration> for u64 {
+    type Error = &'static str;
+
+    fn try_from(value: TransactionDuration) -> Result<Self, Self::Error> {
+        if value.0 > u64::MAX.into() {
+            Err("Value is too large to fit in u64")
+        } else {
+            value.try_into()
+        }
```
aleksuss, line 324: Why? `u64::try_from` will drop an error if `value.0` won't fit to `64` bits.

engine-standalone-storage/src/relayer_db/types.rs
hskang9: yeah that too.

engine-standalone-storage/src/relayer_db/types.rs
hskang9: that too.

engine-standalone-storage/src/relayer_db/types.rs
aleksuss: 100% agree

engine-standalone-storage/src/relayer_db/types.rs
hskang9: But then trait bounds are not met for the type as super::initialize_blocks function in mod.rs requires postgres::Error to return. I tried to return both Error (TryFromIntError and postgres::Error) with an enum type, but iterating RowIter type always requires only postgres::Error to return. Hence, I had to include the integer conversion error into postgres::Error type.

engine-standalone-storage/src/relayer_db/types.rs
hskang9: I could not find a specific error type called `TryFromUnsignedIntError`, so I had to make one.

hskang9: if it was i32 -> {unsigned type}, I would do that, but it seems there is no support for Unsigned int conversion error type.

engine-standalone-storage/src/relayer_db/types.rs
hskang9: After trials and errors, If one makes a new type that wraps `RowIter` type then add error handler for the crate::Error it would work. I need to see how much codes will be effected if I add the change.

engine-standalone-storage/src/relayer_db/types.rs
aleksuss: `type Error = TryFromIntError;`
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=02565c9a8436769fcee41c905acf7c7a

hskang9: oh it works, nice.

hskang9: Ah the documentation says the error is for "integral" types, including unsigned int.

engine/src/lib.rs
aleksuss: I also have an opinion that we should use JSON encoding here. 
cc @joshuajbouw

engine/src/lib.rs
aleksuss: @hskang9 `io.return_output(b"true");` is redundant.

hskang9: The output return in line 159 should have been already removed.

engine/src/lib.rs
hskang9: I agree.

engine/src/lib.rs
aleksuss: I mean on line 161.

aleksuss: > There doesn't need to be a return in case of true.
⬆️

engine/src/lib.rs
aleksuss: Not yet. Because it isn't finished yet. But I have plans to add it this week for the SILO contract.

hskang9: I think I just need to change borsh input to json.

aleksuss: Yes. You're right.
```rust
let args: SetOwnerArgs = serde_json::from_slice(&io.read_input().to_vec())
            .map_err(Into::<ParseTypeFromJsonError>::into)
            .sdk_unwrap();
```

engine/src/lib.rs
```diff
@@ -142,6 +142,24 @@ mod contract {
...
+    #[no_mangle]
+    pub extern "C" fn set_owner() {
+        let mut io = Runtime;
+        let mut state = state::get_state(&io).sdk_unwrap();
+        require_owner_only(&state, &io.predecessor_account_id());
+        let args: parameters::SetOwnerArgs = serde_json::from_slice(&io.read_input().to_vec())
+            .map_err(Into::<ParseTypeFromJsonError>::into)
+            .sdk_unwrap();
+        if state.owner_id == args.new_owner {
+            // Would be a no-op to set, do nothing and return false
```
birchmd, line 155: Outdated comment.

birchmd (approved):

hskang9:
So I was told by @joshuajbouw to overhaul the json encode/decode after adding this method as borsh, because it is inconsistent for admin methods to have it while all other methods use borsh encode/decode for now.

engine-tests/src/tests/sanity.rs
```diff
@@ -1023,6 +1023,64 @@ fn test_eth_transfer_charging_gas_not_enough_balance_sim() {
...
+
+    // set owner args
+    let set_owner_args = SetOwnerArgs {
+        new_owner: str_to_account_id("new_owner.near"),
+    };
+
+    let (outcome, error) = runner.call(
+        "set_owner",
+        &aurora_account_id,
+        set_owner_args.try_to_vec().unwrap(),
```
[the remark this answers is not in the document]
joshuajbouw: We shouldn't flip-flop between JSON and Borsh, all other admin functions are Borsh. Eventually, we should support both, just for backwards compatibility, and improved experience with modern NEAR tools like RAEN and nearblocks.io.

hskang9:
@joshuajbouw it is now used with borsh.

aleksuss:
Resolve the conflicts.

engine/src/lib.rs
```diff
@@ -142,6 +143,24 @@ mod contract {
...
 
+    /// Set owner account id for this contract.
+    #[no_mangle]
+    pub extern "C" fn set_owner() {
+        let mut io = Runtime;
+        let mut state = state::get_state(&io).sdk_unwrap();
+        require_owner_only(&state, &io.predecessor_account_id());
+        let args: SetOwnerArgs = io.read_input_borsh().sdk_unwrap();
+        if state.owner_id == args.new_owner {
+            // Would be a no-op to set, do nothing and return false
```
aleksuss, line 154: Remove this comment please.

hskang9: ok

engine/src/lib.rs
aleksuss: Remove 160 and 161 lines.

aleksuss (approved):
LGTM

joshuajbouw (approved):

PR #693, opened 2023-02-15
Fix: check promise results before executing XCC callbacks

birchmd:
## Description

On Near every promise in a chain of callbacks will execute regardless of the success/failure of previous promises in the chain. We need to explicitly check for success of past promises before continuing.

## Testing

This is pretty tricky to test because we already had many checks in place to ensure that promises would not be failing before executing the chain in the first place. These checks are mostly a safety net for if something totally unexpected occurs. My standpoint is that if this change doesn't break any of the existing xcc tests then it should be ok. The code is simple enough that a careful review should reveal any problems with it.


joshuajbouw (approved):
Looks solid.

PR #695, opened 2023-02-20
replace wee_alloc to lol_alloc

hskang9:
## Description

This PR replaces `wee_alloc` cargo that is not currently being maintained from [dependabot](https://github.com/aurora-is-near/aurora-engine/security/dependabot/101).

Changing allocator **increased** the near gas cost by 1~5

## Testing

All existing tests should pass on the new global allocator

See if there are other options on setting global allocator through `lol_alloc` for optimization on allocating memory and decreasing near gas cost.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,11 +150,11 @@ dependencies = [
  "ethabi",
  "evm",
  "hex",
+ "lol_alloc",
  "rlp",
  "serde",
  "serde_json",
  "test-case",
- "wee_alloc",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


aleksuss:
As I can see, gas cost increases after replacing `wee_alloc` with `lol_alloc`. It's not good. Does it make sense to replace `wee_alloc`?

hskang9:
I agree, `wee_alloc` is not maintained though. Maybe there will be an improvement in `lol_alloc`?

joshuajbouw:
`wee_alloc` was deprecated and no longer maintained. To ensure highest quality, we should strive to seek better alternatives.

aleksuss:
And one more. `wee_alloc` has more than a million downloads, and `lol_alloc` has more than one thousand.

joshuajbouw:
Closing, as this kind of issue isn't one we are concerned about and will remain with whatever NEAR SDK uses.

PR #696, opened 2023-02-21
feat: add a new submit method with additional arguments

aleksuss:
## Description

The PR introduces a new method, `submit_with_args`, which accepts additional arguments: `max_gas_price` and `gas_token_address`. 

There should be no changes in performance and gas costs.

## Testing

Added a test for testing the new method.

## Additional information

The question about adding support for the new method in standalone storage. @birchmd


engine-types/src/types/wei.rs
```diff
@@ -189,6 +189,12 @@ impl From<NEP141Wei> for Wei {
     }
 }
 
+impl From<u128> for Wei {
```
birchmd, line 192: I personally don't like adding convenient conversion functions to a newtype like `Wei`. The reason is because I think we should force developers to think about what their `u128` value represents (is it really Wei? Or maybe it is actually ETH or GWei, etc). If we allow a simple `.into()` anywhere we have a `u128` I worry it will not trigger the same thought process as being forced to write `Wei::new(...)`.

birchmd (changes requested):
To add a new public function you also need to add a variant to the [TransactionKind enum](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-standalone-storage/src/sync/types.rs#L76) (and follow the complier messages from there to finish the implementation; there will be many unexhaustive matches to complete).

engine/src/parameters.rs
```diff
@@ -34,6 +34,17 @@ pub struct AccountBalance {
     pub balance: RawU256,
 }
 
+/// Borsh-encoded submit arguments used by the `submit_with_args` function.
+#[derive(Default, Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct SubmitArgs {
+    /// Bytes of the transaction.
+    pub tx_data: Vec<u8>,
+    /// Max gas price the user is ready to pay for the transaction.
+    pub max_gas_price: Option<u128>,
```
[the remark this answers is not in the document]
aleksuss: As I remember, `U256` doesn't support borsh deserialize.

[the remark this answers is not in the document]
birchmd: Sure, seems reasonable enough. Again, I think `u128` should be big enough to represent gas prices.

engine-standalone-storage/src/sync/types.rs
```diff
@@ -476,6 +479,7 @@ impl<'a> TryFrom<BorshableTransactionMessage<'a>> for TransactionMessage {
 #[derive(BorshDeserialize, BorshSerialize, Clone)]
 enum BorshableTransactionKind<'a> {
     Submit(Cow<'a, Vec<u8>>),
+    SubmitWithArgs(Cow<'a, parameters::SubmitArgs>),
```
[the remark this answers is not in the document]
aleksuss: > I think it would be nice to expand the standalone helpers used in the tests to also handle this new function. See [standalone validation logic](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-tests/src/test_utils/mod.rs#L225-L230) in AuroraRunner which uses the [submit_raw function](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-tests/src/test_utils/standalone/mod.rs#L186). https://github.com/aurora-is-near/aurora-engine/pull/690 also did a similar change.

But it seems already [done](https://github.com/aurora-is-near/aurora-engine/pull/696/files#diff-f65b2de46a5b104b420fca770804b31790e281f48b886d7fe687c8fd9e43fe43R213) No?

birchmd (approved):
LGTM

engine-standalone-storage/src/sync/types.rs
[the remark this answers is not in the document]
birchmd: > But it seems already done

Yes, my apologies. I somehow missed it!

engine/src/lib.rs
```diff
@@ -304,14 +304,42 @@ mod contract {
...
+
+        result
+            .map(|res| res.try_to_vec().sdk_expect(errors::ERR_SERIALIZE))
+            .sdk_process();
+    }
+
+    /// Analog of the `submit` function, but waits for the `SubmitArgs` structure rather than
+    /// the array of bytes representing the transaction.
+    #[no_mangle]
+    pub extern "C" fn submit_with_args() {
```
joshuajbouw, line 333: Its possible to add a variant to this by deserializing first to the most common used one, then to the not-common one.

joshuajbouw (changes requested):

engine/src/lib.rs
aleksuss: Are you sure it's a good idea? The `submit` method is invoked pretty frequently. And on every call, firstly, we should try to deserialize bytes into `SubmitArgs`. It's an additional overhead. Is it really worth it?

aleksuss: We can't deserialize to the most commonly used one because the commonly used one is just an array of bytes (serialized Ethereum transaction).

joshuajbouw: Its quite negligible honestly.

joshuajbouw (approved):

PR #698, opened 2023-02-22
chore: remove unused `etc/state-migration-test/Cargo.lock` file

hskang9:
## Description

Fixes dependabot issue [94](https://github.com/aurora-is-near/aurora-engine/security/dependabot/94) and [98](https://github.com/aurora-is-near/aurora-engine/security/dependabot/98) all happening at path `/etc/state-migration-test/Cargo.lock`. 

This is dependabot fixes, so I don't think gas cost is currently related to this commit. 

## Testing

Run dependabot on this branch and confirm dependabot issue 94 and 98 is gone.

Clarify the use of `Cargo.lock` in `/etc/state-migration-test/` when there is another `state-migration-test` cargo in the `tests` directory.


hskang9:
@RomanHodulak do we still need directory `/etc/state-migration-test/` ? I noticed there is a new directory created in `/etc/tests/state-migration-test/`.

aleksuss (approved):

joshuajbouw (changes requested):
Same issues as: https://github.com/aurora-is-near/aurora-engine/pull/699#pullrequestreview-1309087237

hskang9:
I just removed the `Cargo.lock` file in `/etc/state-migration-test/` as it seems it moved to `/etc/tests/state-migration-test`, so the current dependabot issue in the outdated `Cargo.lock` file should disappear.

birchmd (approved):

aleksuss (approved):

hskang9:
@joshuajbouw I just deleted the outdated Cargo.lock file which creates an issue.

joshuajbouw (changes requested):
If we are deleted this `Cargo.lock` it should be added to `.gitignore` file. Else, every future work will ask for it to be committed.

aleksuss:
> If we are deleted this Cargo.lock it should be added to .gitignore file. Else, every future work will ask for it to be committed.

No. It won't. Probably you mean locally. But if you delete a file by IDE, like CLion, the file will be removed from git index as well. But if you delete it in another way, then you should  delete the file from the git index manually:

```shell
git rm --cached <file>
```

So in this case, you shouldn't change the `.gitignore` file.

hskang9:
Thank you @aleksuss, now the file is removed in cache, and `.gitignore` file is back before edit.

hskang9:
@joshuajbouw I think this is ready to be merged.

Issue #110, opened 2021-05-28
Integration test for upgrading from `master` to `develop`


birchmd:
It would be good to have an integration test that proves we can safely upgrade from current `master` to `develop`, keeping in tact any existing state. If this test fails then we know we need to write a state migration in `develop`. The flow for this test would be something along the lines of

1. Checkout and build `master`
2. Deploy the contract in `near-sdk-sim` and do some transactions (make some accounts, transfer some eth, etc)
3. Checkout and build `develop`
4. Go through the upgrade process (`stage_upgrade` then `deploy_upgrade`)
5. Ensure state is left in tact and do some more transactions to see it works

mrLSD:
@birchmd also for that we should accept `migration flow`, that was started with migration discussion: #140

PR #699, opened 2023-02-22
Fix dependabot issue #110

hskang9:
## Description

Fixes dependabot issue [110](https://github.com/aurora-is-near/aurora-engine/security/dependabot/110) and  all happening at path `/etc/xcc-router/Cargo.lock`. 

If near gas cost changes from changing dependancies, it will appear in the CI.

## Testing

Run dependabot on this branch and issue 110 should be removed.

Run dependabot on this branch and issue 110 should be removed.

### DIFF
--- a/etc/xcc-router/Cargo.lock
+++ b/etc/xcc-router/Cargo.lock
@@ -14,9 +14,18 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
- "getrandom 0.2.7",
+ "getrandom 0.2.8",
  "once_cell",
  "version_check",
+]
+
+[[package]]
+name = "android_system_properties"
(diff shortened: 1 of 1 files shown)


aleksuss (approved):

joshuajbouw (changes requested):
This isn't the right way to approach this. You should never edit the Cargo.lock file directly. Whatever dependency needs to be updated, update only that dependency.

Likewise, change the title to something actually the description of the problem like `fix: <whatever the fix is>`, it doesn't have to mention Dependabot.

aleksuss:
`cargo update` modifies `Cargo.lock` as well.

hskang9:
Ok so after update, it still has time dependancy set to vulnerable version of `chrono` crate. Hopefully, the crate will have 0.5 soon as they just solved this security issue last week. https://github.com/chronotope/chrono/pull/969 if this happens `near_primitives` will need to be update its chrono version.

hskang9:
Wow Near does not keep security on `chrono` package

mrLSD (approved):

aleksuss:
No longer actual

PR #703, opened 2023-02-23
fix: validate account Ids from args

0x3bfc:
# Description

AccountId is a string representing a NEAR account. This string should be between 2 and 64 characters. There are
additional requirements such as limited use of alphanumeric characters and separator characters.

There are a number of external smart contract functions in aurora-engine which BorshDeserialize an AccountId without validating it.

- [x] implement the missing account validations
- [x] Adding extra tests to cover some edge cases.


engine-types/src/account_id.rs
```diff
@@ -93,6 +92,20 @@ impl AccountId {
...
 }
 
+impl BorshDeserialize for AccountId {
+    fn deserialize(buf: &mut &[u8]) -> io::Result<Self> {
+        let account: String = BorshDeserialize::deserialize(buf)?;
+
+        // It's for saving backward compatibility.
+        if account.is_empty() {
+            return Ok(Self::default());
+        }
```
aleksuss, line 102: Pay attention to this approach, please. It seems should be ok because we just skip validating empty strings only... which is the default one in our case. All non-empty strings will be validated. 
cc @joshuajbouw @birchmd

engine-tests/src/tests/contract_call.rs
```diff
@@ -30,6 +30,14 @@ fn setup_test() -> (AuroraRunner, Signer, Address, Tester) {
     (runner, signer, token, tester)
 }
 
+#[test]
+#[should_panic]
+fn test_deploy_erc20_token_with_invalid_account_id() {
+    let mut runner = AuroraRunner::new();
+    let invalid_nep141 = "";
```
aleksuss, line 37: Now it's a default and valid `AccountId`.

0x3bfc: Exactly, I will update it with an invalid accountId.

0x3bfc, line 37:
```suggestion
    let invalid_nep141 = "_";
```

engine-types/src/account_id.rs
```diff
@@ -93,6 +92,20 @@ impl AccountId {
     }
 }
 
+impl BorshDeserialize for AccountId {
+    fn deserialize(buf: &mut &[u8]) -> io::Result<Self> {
+        let account: String = BorshDeserialize::deserialize(buf)?;
+
+        // It's for saving backward compatibility.
```
birchmd, line 99: Are there examples where we needed to deserialize an AccountId that was empty? If not then I think we should not allow empty values.

aleksuss: For example, `bridge_prover_id` is empty when restoring a snapshot from `engine-tests/src/tests/res/contract.aurora.block51077328.json`. The same situation with `engine-tests/src/tests/res/aurora_state_*.json` files.

birchmd: I see. That's pretty unfortunate. Alright, we must accept empty values as valid then. Thanks for pointing out the examples.

engine-types/src/account_id.rs
aleksuss: > we are kind of stuck now

Yes. Exactly.

aleksuss (approved):

engine-types/src/account_id.rs
joshuajbouw: We don't need to persist it, we can't reply all transactions without exceptions already. We've fixed all those issues. Imo, safe to remove it because we do plan on going to NEAR's AccountId crate at some point. I do understand that we would need a new snapshot however though as those as all default Account Id's?

birchmd: It's not just about old transactions, I think it is still true that `bridge_prover_id` is an empty string on mainnet. It's not used, so this isn't a problem for anything, but this does meant that some kind of migration would be needed if we wanted to make the empty string an invalid value.

I also think that if we do decide to do that migration we should still maintain backwards compatibility because it would not be acceptable to be unable to replay such a large portion of the Aurora mainnet history. I think this can be achieved by being careful in how we define the legacy version of the engine state when we do the migration.

joshuajbouw (approved):

engine-types/src/account_id.rs
aleksuss: But near's account_id doesn't have the `Default` implementation.

PR #704, opened 2023-02-23
Fix: re-enable deriving serde traits

birchmd:
## Description

In PR #677 we made `serde` required instead of optional in `aurora-engine` and `aurora-engine-types`. This change means the `serde` feature no longer exists and this prevents the serde traits from being derived on several types. In this PR I fix this regression by using the `impl-serde` feature to trigger deriving the traits instead. I also add new tests which will catch this problem in the future.

### DIFF
--- a/engine-tests/Cargo.toml
+++ b/engine-tests/Cargo.toml
@@ -15,12 +15,12 @@ autobenches = false
 [dependencies]
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


aleksuss (approved):
LGTM

joshuajbouw (approved):

PR #705, opened 2023-02-24
Aurora block hashchain

Casuso:
## Description
Aurora Block Hashchain

This implementation correlates to AIP 8.

Hashchain is turned off by default. We need to call `start_hashchain` on the contract to turned it on. More on this in the additional information below.

We are using a Merkle tree to compute the transactions hash of a block. The tree is actually never constructed; we dynamically maintain only the growing branch, storing hashes of the fully completed subtrees.

The hierarchy of usage is:
-contract (engine/src/lib.rs): Uses BlockchainHashchain to send txs and get the block hashchain.
-standalone: Uses BlockchainHashchain to send txs. 
-BlockchainHashchain: Keeps track of block hashchain and other info, and uses BlockHashchainComputer to add txs and get hashes.
-BlockHashchainComputer: Keeps track of the StreamCompactMerkleTree and computes hashes.
-StreamCompactMerkleTree: Maintains the growing subtrees and compact subtrees applying Merkle tree rules.

To measure the impact on gas consumption of the block hashchain computation, we ran four tests against a base branch and the hashchain branch. After getting the raw gas consumption results from both, we found the delta (difference) between them and used some statistics to show the impact.

Tests 1 and 2 are focused on executing transactions on the same block, which is the most common case as only one transaction triggers the change in block height. For these the average increase observed (delta ave) is less than 0.46  Tgas, and the maximums are less than 0.64 Tgas.

Tests 3 and 4 are focused on executing a transaction on a block height change, that triggers the block hashchain computation, which is the most extensive one. Both show average and max increases of less than 0.62 Tgas.

An increase of 0.64 or 0.62 Tgas is a small amount so we should be fine. For reference, the transaction gas limit is 300 Tgas.

## Testing
Several tests to cover the described structures.

Existing tests are sufficient to correlated the state between contract and standalone.

Specifically, I would like to get feedback on the contract entry points methods on engine/src/lib.rs.

I want to make sure that we are reading correctly the input and output in all the methods that include the hashchain update. This is critical since a small difference on the input or output would imply a different hashchain value for all the blockchain.

Also, I want to make sure that we are covering all and only all the methods in the contract that should include the hashchain. The list of all the contract methods with their inclusion or exclusion of hashchain will be added as a comment on this PR so you can review it.

## Additional information
The discussed mechanism to start the hashchain was:
1. We should merge this PR with the hashchain off.
2. A hashchain seed value should be separately precalculated (and continually updated) starting from Aurora Genesis block.
3. Aurora DAO should pause the contract. There is a separate [PR](https://github.com/aurora-is-near/aurora-engine/pull/779) for that. The hashchain seed value should include that last tx.
4. Aurora Labs should call `start_hashchain` passing the hashchain seed with its corresponding block height. This method adds its own data to the hashchain since it changes the state.
It also resumes the contract, we don't need another call to resume.
6. We can remove the methods _pause, resume or start hashchain_  if decided.

TODO:
1. Add an account requirement (from Aurora Labs) in the `start_hashchain` method.
2. Merge the pause contract PR and add hashchain update to pause and resume methods since they change the state.


engine/src/lib.rs
```diff
@@ -120,7 +124,18 @@ mod contract {
...
+        let blockchain_hashchain = BlockchainHashchain::new(
+            &state.chain_id,
+            b"aurora",
+            io.block_height(),
+            [0; 32],
+            [0; 32],
+        );
+
+        state::set_state(&mut io, state).sdk_unwrap();
+        hashchain::set_state(&mut io, blockchain_hashchain).sdk_unwrap();
```
birchmd, line 138: Adding an extra entry to the state in the contract only will make many tests fail because the standalone engine will not produce the same state. It's fine for now while this PR is in draft and we are mostly just trying to measure the gas cost of this change, but we will eventually need to fix it so that the standalone engine is capturing these state changes too.

engine/src/lib.rs
```diff
@@ -120,7 +124,18 @@ mod contract {
         }
 
         let args: NewCallArgs = io.read_input_borsh().sdk_unwrap();
-        state::set_state(&mut io, args.into()).sdk_unwrap();
+        let state: EngineState = args.into();
+
+        let blockchain_hashchain = BlockchainHashchain::new(
+            &state.chain_id,
+            b"aurora",
```
birchmd, line 131: Should use `io.current_account_id()` instead.

engine/src/lib.rs
```diff
@@ -204,33 +219,41 @@ mod contract {
...
         require_owner_only(&state, &predecessor_account_id);
 
-        let args: PausePrecompilesCallArgs = io.read_input_borsh().sdk_unwrap();
+        let input = io.read_input();
+        let args: PausePrecompilesCallArgs = input.to_value().sdk_unwrap();
         let flags = PrecompileFlags::from_bits_truncate(args.paused_mask);
         let mut pauser = EnginePrecompilesPauser::from_io(io);
         pauser.resume_precompiles(flags);
+
+        update_hashchain(&mut io, function_name!(), &input.to_vec(), &[0; 0]);
```
Casuso, line 236: In many methods, I'm leaving the output empty (`&[0; 0]`) because I don't see a call to `sdk_process`. Please check this list of methods and let me know if I'm missing the output for some:

1. resume_precompiles
2. pause_precompiles
3. register_relayer
4. factory_update
5. factory_update_address_version
6. factory_set_wnear_address
7. ft_on_transfer
8. refund_on_error
9. new_eth_connector
10. set_eth_connector_contract_data
11. deposit
12. finish_deposit
13. ft_transfer
14. ft_resolve_transfer
15. ft_transfer_call
16. storage_deposit
17. storage_unregister
18. storage_withdraw
19. get_paused_flags

engine/src/lib.rs
```diff
@@ -481,6 +555,30 @@ mod contract {
         io.return_output(block_hash.as_bytes())
     }
 
+    #[no_mangle]
+    pub extern "C" fn get_previous_block_hashchain() {
```
Casuso, line 559: In this method, I'm returning the previous block hashchain, which is the hashchain of the last completed block. Do you think it will be better to return a tuple` <block_height, block_hashchain>`, so it is clear to which block belongs?

As a note for your consideration about this, near skip blocks and near blocks without aurora txs, are accounted on the hashchain computation. This means that if we have a block H with some aurora txs, then a block H + 1 as a skip block or a block without aurora txs, and then at block H + 2 we call `get_previous_block_hashchain`, we will get the hashchain of H + 1 and not the one of block H.

engine/src/lib.rs
```diff
@@ -481,6 +555,30 @@ mod contract {
...
+                .move_to_block(block_height)
+                .sdk_unwrap();
+        }
+
+        let previous_block_hashchain = blockchain_hashchain.get_previous_block_hashchain();
+        io.return_output(&previous_block_hashchain);
+    }
+
+    #[no_mangle]
+    pub extern "C" fn get_genesis_block_hashchain() {
```
Casuso, line 575: Similar comment here about returning a tuple `<block_height, block_hashchain>` so it is clear where the genesis hashchain started.

engine/src/hashchain.rs
```diff
@@ -0,0 +1,889 @@
...
+    pub fn compute_block_hashchain(
+        &self,
+        chain_id_hash: RawH256,
+        contract_account_id_hash: RawH256,
+        current_block_height_hash: RawH256,
+        previous_block_hashchain: RawH256,
+    ) -> RawH256 {
+        let txs_hash = self.txs_merkle_tree.compute_hash();
+
+        keccak(
```
Casuso, line 208: This is the current hashing formula that I'm applying for a block. Do you have any comments regarding the parameters or order that I'm using?

engine/src/lib.rs
```diff
@@ -991,6 +1139,32 @@ mod contract {
         near_account_to_evm_address(predecessor_account_id.as_bytes())
     }
 
+    fn update_hashchain(io: &mut Runtime, method_name: &str, input: &[u8], output: &[u8]) {
+        let block_height = io.block_height();
+
+        let mut blockchain_hashchain = hashchain::get_state(io).unwrap_or_else(|_| {
```
Casuso, line 1145: This reads the state of the hashchain or create a new one if the state does not exists.

I don't think this behave should go to production. We should ensure that the upgrade, or new etc. methods, create the state so it exists at this point.

birchmd: Yes, I agree. When we have a proper migration story in place then we should assume the state is present.

engine/src/lib.rs
birchmd: Genesis will start at 0 (this is the standard for Ethereum-like chains), so I do not think the block height is needed here.

engine/src/lib.rs
birchmd: Yes, good idea. I agree it should return the block height associated with the hash, for clarity.

engine/src/lib.rs
birchmd: Methods that include `promise_return` are tricky. The "output" from this function on the Near level is a receipt ID because the function is making a cross-contract call. Unfortunately for us the receipt ID cannot be known inside our contract so we cannot capture it for the hashchain. This means it will still be possible for someone to lie about what receipts corresponded with some executions, but I do not see a good way to resolve this problem.

This applies to functions such as `deposit`, `finish_deposit` and `ft_transfer_call`.

For functions where there is legitimately no return value I think `&[]` is fine. The `&[0; 0]` I think is equivalent, but harder to read in my opinion.

> 19. get_paused_flags

I think you means `set_paused_flags`.

engine/src/hashchain.rs
birchmd: No comments, looks good to me.

engine/src/lib.rs
Casuso: Regarding `promise_return`, what would happen then on the Refiner for the corresponding txs? We would have there the receipt ID I suppose, so we will need to ignore it to build the same hash as in the Engine.

Sure, `&[] `is better.

And yes, I mean `set_paused_flags`.

birchmd: Yes that's right, the Refiner would have access to the receipt ID, but would have to ignore it to get the same result as on-chain.

engine/src/lib.rs
Casuso: This is done. The standalone engine is now on sync with the contract.

engine/src/lib.rs
Casuso: Current behave is that Hashchain is off by default. It is needed to call `start_hashchain` method to start the hashchain.

engine/Cargo.toml
```diff
@@ -20,9 +20,13 @@ aurora-engine-transactions.workspace = true
 aurora-engine-types.workspace = true
 aurora-engine-sdk.workspace = true
 bitflags.workspace = true
+borsh.workspace = true
```
aleksuss, line 23: Use `borsh` from `aurora-engine-types`.

engine/Cargo.toml
```diff
@@ -48,4 +55,4 @@ integration-test = ["log"]
 all-promise-actions = ["aurora-engine-sdk/all-promise-actions"]
 mainnet = ["contract", "log", "aurora-engine-sdk/mainnet"]
 testnet = ["contract", "log", "aurora-engine-sdk/testnet"]
-impl-serde = ["aurora-engine-types/impl-serde", "aurora-engine-transactions/impl-serde", "evm/with-serde"]
+impl-serde = ["aurora-engine-types/impl-serde", "aurora-engine-transactions/impl-serde", "evm/with-serde"]
```
aleksuss, line 58: No new line.

engine/src/bloom.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub fn accrue(&mut self, input: &[u8]) {
+        let m = self.0.len();
+        let bloom_bits = m * 8;
+        let mask = bloom_bits - 1;
+        let bloom_bytes = (log2(bloom_bits) + 7) / 8;
+        let hash = keccak(input);
+        let mut ptr = 0;
+
+        for i in 0..BLOOM_BITS {
+            let _ = i;
```
aleksuss, line 51: Why? Could you please explain this trick?

engine/src/hashchain.rs
```diff
@@ -0,0 +1,973 @@
...
+use borsh::{BorshDeserialize, BorshSerialize};
+#[cfg(feature = "borsh-compat")]
+use borsh_compat::{self as borsh, BorshDeserialize, BorshSerialize};
+
+/// Blockchain Hashchain.
+/// Continually keeps track of the previous block hashchain through the blocks heights.
+#[derive(BorshSerialize, BorshDeserialize, Debug)]
+pub struct BlockchainHashchain {
+    chain_id: [u8; 32],
+    contract_account_id: Vec<u8>,
```
aleksuss, line 16: Why not `AccountId`?

engine/src/hashchain.rs
```diff
@@ -0,0 +1,973 @@
+use self::blockchain_hashchain_error::BlockchainHashchainError;
+use crate::{bloom::Bloom, prelude::Vec};
+use aurora_engine_sdk::keccak;
+use aurora_engine_types::types::RawH256;
+
+#[cfg(not(feature = "borsh-compat"))]
+use borsh::{BorshDeserialize, BorshSerialize};
+#[cfg(feature = "borsh-compat")]
+use borsh_compat::{self as borsh, BorshDeserialize, BorshSerialize};
```
aleksuss, line 9: borsh from `aurora-engine-types`

engine/src/bloom.rs
Casuso: Sorry for not clarifying this before. This file is coming from the Refiner, [here](https://github.com/aurora-is-near/borealis-engine-lib/blob/main/refiner-types/src/bloom.rs). I plan to remove it from Refiner later and consume it from the engine.

engine/src/hashchain.rs
Casuso: For the hashes, things are used in their bytes representation. We can use AccountId here, but I wanted to have as little dependencies as possible so hashchain code is easily replicated on other code bases or languages. For example, right now we are thinking on doing it on Go for the hashchain seed.

In any case, there is a good chance that we remove this field from the hashchain. It is nice to have and it should be easy to obtain, but it is not explicitly part of the Aurora blocks.

Casuso: We finally removed this one.

joshuajbouw:
Adding do not merge until it is ready for release.

engine/src/lib.rs
```diff
@@ -335,19 +421,26 @@ mod contract {
             &io,
         )
         .sdk_unwrap();
-        Engine::deploy_code_with_input(&mut engine, input, &mut Runtime)
-            .map(|res| res.try_to_vec().sdk_expect(errors::ERR_SERIALIZE))
+
+        Engine::deploy_code_with_input(&mut engine, input.clone(), &mut Runtime)
+            .map(|res| {
+                let output = res.try_to_vec().sdk_expect(errors::ERR_SERIALIZE);
```
birchmd, line 427: It's suboptimal that we end up serializing the output twice (once here and once in `sdk_process`. Maybe worth looking into removing that duplication of work, though the gas cost of serialization is probably not that high in most cases.

engine/src/lib.rs
```diff
@@ -547,33 +688,43 @@ mod contract {
                 &mut Runtime,
             );
         }
+
+        update_hashchain(&mut io, function_name!(), &input, &[], &Bloom::default());
```
birchmd, line 692: We need to be careful about the outputs. The output is not empty in the case of this function. You can see the `io` object is used [inside receive_erc20_tokens](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine/src/engine.rs#L781). I think I mentioned this before, but I still wonder if it makes sense to try and connect the hashchain with the `IO` implementation of `NearRuntime` so that we never miss outputs.

I think the issue we ran into before is how to pass in the additional information (input, function name, bloom) without breaking the `IO` API. I'm wondering if there is a way to do a sort of builder pattern so that each call to the `NearRuntime` object updates the piece of the hashchain it knows about, but maybe that will be awkward and not work out well either.

In any case, if we are manually passing outputs then we will need to be very careful that all outputs are properly captured.

engine/src/lib.rs
```diff
@@ -946,15 +1161,18 @@ mod contract {
             // that they over paid for their deposit.
             unsafe { io.promise_create_batch(&promise) };
         }
+        update_hashchain(&mut io, function_name!(), &input, &[], &Bloom::default());
```
birchmd, line 1164: This function also has a [non-empty output](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine/src/connector.rs#L539). And similarly for the other `storage_*` functions.

birchmd (changes requested):
Requesting changes because there are a few outputs that are not correctly captured by the hashchain. Please also resolve conflicts to be up to date with the latest develop branch.

engine-standalone-storage/src/sync/mod.rs
```diff
@@ -117,7 +124,7 @@ fn execute_transaction<'db, M: ModExpAlgorithm + 'static>(
     let relayer_address =
         aurora_engine_sdk::types::near_account_to_evm_address(predecessor_account_id.as_bytes());
     let near_receipt_id = transaction_message.near_receipt_id;
-    let current_account_id = engine_account_id;
+    let current_account_id = engine_account_id.clone();
```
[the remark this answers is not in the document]
Casuso: Fixed. I probably had to do this during some debugging and then forgot to undo it.

engine/src/lib.rs
Casuso: I'm honestly not sure about how this is different from the base code:

```Engine::deploy_code_with_input(&mut engine, input, &mut Runtime)
            .map(|res| res.try_to_vec().sdk_expect(errors::ERR_SERIALIZE))
            .sdk_process();```

File: engine-standalone-storage/src/sync/mod.rs (hunk 16 above)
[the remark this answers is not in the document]
Comment (birchmd) [reply engine-standalone-storage/src/sync/mod.rs:127]: Not fixed yet, maybe you forgot to push the commit?

File: engine/src/lib.rs (hunk 14 above)
Comment (Casuso) [reply engine/src/lib.rs:696]: As per our conversation, capturing the output here and in others would require an investigation of how they look on the standalone side (`get_output_and_log_bloom` method).

File: engine/src/lib.rs (hunk 15 above)
Comment (Casuso) [reply engine/src/lib.rs:1256]: As per our conversation, capturing the output here and in others would require an investigation of how they look on the standalone side (get_output_and_log_bloom method).

File: engine/src/lib.rs [hunk 17]
```diff
@@ -295,30 +325,85 @@ mod contract {
...
+    /// Assumes that no tx has been accepted after the last tx included on the indicated block hashchain.
+    /// This self tx is added to the started hashchain as it changes the state of the contract.
+    #[no_mangle]
+    #[named]
+    pub extern "C" fn start_hashchain() {
+        let mut io = Runtime;
+        let mut state = state::get_state(&io).sdk_unwrap();
+
+        // *** TODO requires some Aurora Labs account
+        // require_account(some_AuroraLabs_account);
```
Casuso, line 342: Please add a `require_account` statement here for an account owned by Aurora Labs. For security reasons, it's better that I don't create the account.

birchmd (approved):
The hashchain implementation still needs a little more iteration as discussed in the comments above regarding how outputs are captured. However, maintaining a feature branch like this with significant code changes requires a lot of work. Given that this PR will have no effect on future Engine deployments until the hashchain feature is enabled in the state by the DAO pausing the contract and us calling `start_hashchain`, I propose that we merge this PR and iterate on the implementation in smaller follow-up PRs.

To reiterate, my argument is that is safe to merge this PR (all functionality is runtime-gated in a way that only the DAO can enable) and that not merging this PR creates greater maintenance overhead for finishing the work. Therefore merging it now seems better to me than continuing development on the feature branch.

Let me know what you think @joshuajbouw @aleksuss @mandreyel

birchmd:
Closing in favour of #831

PR #711, opened 2023-03-02
Fix: Use of Panicking Macros and Functions in aurora-engine Code

lempire123:
Avoid using `unwrap()` or `expect()` functions in the production code as these may lead to unforeseen panics. Instead,
gracefully handle error conditions and returned `Err` values by propagating them with `?` operator and handling them further up the execution chain.

### DIFF
--- a/engine-sdk/src/near_runtime.rs
+++ b/engine-sdk/src/near_runtime.rs
@@ -76,7 +76,7 @@ impl Runtime {
         match AccountId::try_from(bytes) {
             Ok(account_id) => account_id,
             // the environment must give us a valid Account ID.
-            Err(_) => unreachable!(),
+            Err(_) => unreachable!("Invalid Account ID"),
         }
     }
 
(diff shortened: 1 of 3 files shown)


joshuajbouw (changes requested):
This doesn't appear to do what the description details.

lempire123:
> This doesn't appear to do what the description details.

Right, of the code lines changed none were using `unwrap()` or `expect()`, I think more generally the report meant to say that there should be some error message detailing why the code panicked. Also, those were the changes suggested by the audit report (except for the last change, I thought `match` was a good way of handling the error condition, again, let me know if there's a better way to do it).

Aurora-ClickUp:
Task linked: [CU-863g0gja0 AUR-15 Use of Panicking Macros and Functions in aurora-engine Code](https://app.clickup.com/t/863g0gja0)

aleksuss (approved):

joshuajbouw (approved):

PR #714, opened 2023-03-02
fix: avoid attaching values to precompiles

0x3bfc:
## Description

The following precompiles allow attached tokens but will never use them:

- predecessorAccountId
- currentAccountId
- prepaidGas
- getPromiseResults
- randomSeed
- nearCrossContractCall


engine-precompiles/src/random.rs
```diff
@@ -39,9 +39,10 @@ impl Precompile for RandomSeed {
         &self,
         input: &[u8],
         target_gas: Option<EthGas>,
-        _context: &Context,
+        context: &Context,
         _is_static: bool,
     ) -> EvmPrecompileResult {
+        utils::validate_no_value_attached_to_precompile(context.apparent_value.as_u128())?;
```
aleksuss, line 45: Why not to change the `validate_no_value_attached_to_precompile` function like this:
```rust
pub fn validate_no_value_attached_to_precompile(value: U256) -> Result<(), ExitError> {
     if value > U256::zero() {
         // don't attach native token value to that precompile
         return Err(ExitError::Other(Borrowed("ATTACHED_VALUE_ERROR")));
     }
     Ok(())
 }
```
In such way you don't need to call `.as_u128()`.

0x3bfc: Yes, I was going to change it. But I am still waiting a confirmation from @joshuajbouw about whether it is the expected fix or not.

birchmd (commented):
These changes look fine to me. Though I wonder if this is really needed. The normal EVM precompiles accept ETH  (for example [ecrecover](https://etherscan.io/address/0x0000000000000000000000000000000000000001) has 53 ETH on mainnet), so why shouldn't our custom precompiles? I know that this is setting up some guardrails for users to not do something stupid, so it's probably a good thing overall, but it is a little inconsistent.

Anyway, assuming we do want to keep this change then I think it would be good to see integration tests for it as well. Something else to consider: do we want these errors to be `Other` or do we want them to be `Revert`? This has implications for how the errors get handled by the EVM and by our own infrastructure, which is another reason why I think writing integration tests is important (we will know how the error will manifest in the wild).

joshuajbouw:
> Though I wonder if this is really needed.

I was also quite confused as to why this was flagged in the audit. But, they seemed to make sure that its important enough to add.

Aurora-ClickUp:
Task linked: [CU-863g0gk0p AUR-19 Precompiles Allow Attached Value](https://app.clickup.com/t/863g0gk0p)

joshuajbouw (approved):

PR #715, opened 2023-03-03
Fix: deploy_upgrade() Does Not Clear Upgrade Index

lempire123:
There is an issue in `deploy_upgrade()` that the upgrade index is not cleared after each call.

Therefore, `deploy_upgrade()` will pass the check` io.block_height() <= index + state.upgrade_delay_blocks` since
`index` remains unchanged from previous calls.

### DIFF
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -75,7 +75,7 @@ pub unsafe fn on_alloc_error(_: core::alloc::Layout) -> ! {
 #[cfg(feature = "contract")]
 mod contract {
     use borsh::{BorshDeserialize, BorshSerialize};
-    use parameters::SetOwnerArgs;
+    use parameters::{SetOwnerArgs, SetUpgradeDelayBlocksArgs};
 
     use crate::connector::{self, EthConnectorContract};
     use crate::engine::{self, Engine};
@@ -122,8 +122,8 @@ mod contract {
     #[no_mangle]
     pub extern "C" fn new() {
         let mut io = Runtime;
-        
-        if state::get_state(&io).is_some() {
+
+        if state::get_state(&io).is_ok() {
             sdk::panic_utf8(b"ERR_ALREADY_INITIALIZED");
         }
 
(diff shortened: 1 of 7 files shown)


engine/src/lib.rs
```diff
@@ -206,6 +206,10 @@ mod contract {
             sdk::panic_utf8(errors::ERR_NOT_ALLOWED_TOO_EARLY);
         }
         Runtime::self_deploy(&bytes_to_key(KeyPrefix::Config, CODE_KEY));
+        io.write_storage(
```
aleksuss, line 209: Why not remove the element?

```rust
io.remove_storate(&bytes_to_key(KeyPrefix::Config, CODE_STAGE_KEY));
```

lempire123: Yeah this would be better

lempire123:
> #716 duplicate?

I think it's a different issue.

engine/src/lib.rs
aleksuss: It seems that the approach from #716 fixes the same issue but without touching the storage which is good.

aleksuss:
> I think it's a different issue.

It seems that the approach from https://github.com/aurora-is-near/aurora-engine/pull/716 fixes the same issue but without touching the storage which is good.

aleksuss:
@lempire123 sorry. It's my fault. I thought once again and yes, we should remove the deploy index after successful deployment to prevent repeated deployment of the same code.  I re-open the PR.

aleksuss:
@joshuajbouw, the change prevents to re-deploy upgrade more than once. Is it expected behaviour?

@lempire123 I think this is an excellent point to cover with some tests of this functionality.

joshuajbouw (changes requested):
I don't see it as a duplicate. However, yes it is a good idea to ensure that we have tests covering this.

Aurora-ClickUp:
Task linked: [CU-85zrrm83g AUR-18 deploy_upgrade() Does Not Clear Upgrade Index](https://app.clickup.com/t/85zrrm83g)

mandreyel:
Does this need to be rebased?

PR #716, opened 2023-03-03
fix: setting the upgrade index prior calling `stage_upgrade`

0x3bfc:
### Description

Modifying the `stage_upgrade()` to store `block_height + upgrade_delay_blocks`  and avoid changing the delay period between `stage_upgrade` and `deploy_upgrade`.

### DIFF
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -178,7 +178,7 @@ mod contract {
         let mut io = Runtime;
         let state = state::get_state(&io).sdk_unwrap();
         let index = internal_get_upgrade_index();
-        io.return_output(&(index + state.upgrade_delay_blocks).to_le_bytes())
+        io.return_output(&(index).to_le_bytes())
     }
 
     /// Stage new code for deployment.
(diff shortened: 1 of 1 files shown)


engine/src/lib.rs
```diff
@@ -178,20 +178,20 @@ mod contract {
         let mut io = Runtime;
         let state = state::get_state(&io).sdk_unwrap();
```
hskang9, line 179: I see this line is unnecessary as there is an internal function `internal_get_upgrade_index()`. can you remove `state` variable?

hskang9 (changes requested):
I see line 179 is unnecessary as there is an internal function internal_get_upgrade_index(). can you remove state variable?

engine/src/lib.rs
```diff
@@ -178,20 +178,20 @@ mod contract {
...
         let index = internal_get_upgrade_index();
-        io.return_output(&(index + state.upgrade_delay_blocks).to_le_bytes())
+        io.return_output(&(index).to_le_bytes())
```
aleksuss, line 181: The parentheses around `index` are redundant. 
```suggestion
        io.return_output(&index.to_le_bytes());
```

birchmd (approved):

mrLSD (approved):

aleksuss (approved):

joshuajbouw (approved):

PR #717, opened 2023-03-06
fix: return error if signature len does not eq `65`

0x3bfc:
### Description

the `ecrecover` precompile panics if provided signature length is not exactly 65.

### DIFF
--- a/engine-precompiles/src/secp256k1.rs
+++ b/engine-precompiles/src/secp256k1.rs
@@ -19,7 +19,9 @@ mod consts {
 // Quite a few library methods rely on this and that should be changed. This
 // should only be for precompiles.
 pub fn ecrecover(hash: H256, signature: &[u8]) -> Result<Address, ExitError> {
-    assert_eq!(signature.len(), 65);
+    if signature.len() != 65 {
+        return Err(ExitError::Other(Borrowed("INVALID_SIGNATURE_LENGTH")));
+    }
 
     #[cfg(feature = "contract")]
     return sdk::ecrecover(hash, signature).map_err(|e| ExitError::Other(Borrowed(e.as_str())));
@@ -246,5 +248,16 @@ mod tests {
             .unwrap()
             .output;
         assert_eq!(res, expected);
+    }
+
+    #[test]
+    pub fn test_invalid_signature_length() {
+        let hash = H256::from_slice(
+            &hex::decode("1111111111111111111111111111111111111111111111111111111111111111")
+                .unwrap(),
+        );
+        let signature = hex::decode("123").unwrap();
+        let res = ecrecover(hash, &signature);
+        assert!(res.is_err());
     }
 }


engine-precompiles/src/secp256k1.rs
```diff
@@ -19,7 +19,9 @@ mod consts {
 // Quite a few library methods rely on this and that should be changed. This
 // should only be for precompiles.
 pub fn ecrecover(hash: H256, signature: &[u8]) -> Result<Address, ExitError> {
-    assert_eq!(signature.len(), 65);
+    if signature.len() != 65 {
```
aleksuss, line 22: Magic numbers? 😄 Add please constant `SIGNATURE_LEN` near with `INPUT_LEN` into the `consts` module.

engine-precompiles/src/secp256k1.rs
```diff
@@ -247,4 +250,15 @@ mod tests {
...
         assert_eq!(res, expected);
     }
+
+    #[test]
+    pub fn test_invalid_signature_length() {
+        let hash = H256::from_slice(
+            &hex::decode("1111111111111111111111111111111111111111111111111111111111111111")
+                .unwrap(),
+        );
+        let signature = hex::decode("123").unwrap();
```
aleksuss, line 260: Hex string couldn't be odd.

engine-precompiles/src/secp256k1.rs
```diff
@@ -19,7 +20,9 @@ mod consts {
 // Quite a few library methods rely on this and that should be changed. This
 // should only be for precompiles.
 pub fn ecrecover(hash: H256, signature: &[u8]) -> Result<Address, ExitError> {
-    assert_eq!(signature.len(), 65);
+    if signature.len() != consts::SIGNATURE_LENGTH {
+        return Err(ExitError::Other(Borrowed("INVALID_SIGNATURE_LENGTH")));
+    }
```
birchmd, line 25: Instead of returning an error, what if we changed the type signature to `signature: &[u8; consts::SIGNATURE_LENGTH]`? All our internal usage of this function is already doing that, so no other changes should be needed.

Yes, this would technically be a breaking change to a public API, but I think it's fine because I doubt anyone is using this function in their project and if they are then it should be easy enough for them to fix it at their call-site.

birchmd (commented):
While we're looking at ecrecover. It might be good to expand our [integration tests](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-tests/src/tests/ecrecover.rs) for it to cover this case as well. I took a look at geth and extra bytes after a valid signature are simply ignored:

Valid signature:
```
go-ethereum$ ./build/bin/evm \
    --input 0x18c547e4f7b0f325ad1e56f57e26c745b09a3e503d86e00e5255ff7f715d3d1c000000000000000000000000000000000000000000000000000000000000001c73b1693892219d736caba55bdb67216e485557ea6b6af75f37096c9aa6a5a75feeb940b1d03b21e36b0e47e79769f095fe2ab855bd91e3a38756b7d75a9c4549 \
    --receiver 0x0000000000000000000000000000000000000001 \
    run
0x000000000000000000000000a94f5374fce5edbc8e2a8697c15331677e6ebf0b
```

Extra bytes:
```
go-ethereum$ ./build/bin/evm \
    --input 0x18c547e4f7b0f325ad1e56f57e26c745b09a3e503d86e00e5255ff7f715d3d1c000000000000000000000000000000000000000000000000000000000000001c73b1693892219d736caba55bdb67216e485557ea6b6af75f37096c9aa6a5a75feeb940b1d03b21e36b0e47e79769f095fe2ab855bd91e3a38756b7d75a9c4549aabbccddeeff \
    --receiver 0x0000000000000000000000000000000000000001 \
    run
0x000000000000000000000000a94f5374fce5edbc8e2a8697c15331677e6ebf0b
```

I think our implementation does this properly too, but's good to include a test for it just in case.

engine-precompiles/src/secp256k1.rs
```diff
@@ -11,15 +11,21 @@ mod costs {
...
 /// See: `https://etherscan.io/address/0000000000000000000000000000000000000001`
 // Quite a few library methods rely on this and that should be changed. This
 // should only be for precompiles.
-pub fn ecrecover(hash: H256, signature: &[u8]) -> Result<Address, ExitError> {
-    assert_eq!(signature.len(), 65);
+pub fn ecrecover(
+    hash: H256,
+    signature: &[u8; consts::SIGNATURE_LENGTH],
+) -> Result<Address, ExitError> {
+    // if signature.len() != consts::SIGNATURE_LENGTH {
```
birchmd, line 26: Please remove outdated code.

joshuajbouw (changes requested):
As @birchmd noted above, this shouldn't be added until there is test. Unit test is fine enough.

Aurora-ClickUp:
Task linked: [CU-861mfek4f AUR-23 Precompile panics on signatures of invalid length](https://app.clickup.com/t/861mfek4f)

Aurora-ClickUp:
Task linked: [CU-85zrqmcza AUR-23 secp256k1::ecrecover() Precompile Panics on Signatures of Invalid Length](https://app.clickup.com/t/85zrqmcza)

0x3bfc:
> > > As @birchmd noted above, this shouldn't be added until there is test. Unit test is fine enough.
> > 
> > 
> > Unit test won't work as it will lead to a compile time error.
> 
> Why so?



> > > As @birchmd noted above, this shouldn't be added until there is test. Unit test is fine enough.
> > 
> > 
> > Unit test won't work as it will lead to a compile time error.
> 
> Why so?

Testing [ecrecover](https://github.com/aurora-is-near/aurora-engine/blob/c5079625fa62242035e8f570528f3ae1b5f3607f/engine-precompiles/src/secp256k1.rs#L24) with invalid signature length will lead to a compiler time error. Testing it using `run()` will also won't work due to the [input resize here](https://github.com/aurora-is-near/aurora-engine/blob/c5079625fa62242035e8f570528f3ae1b5f3607f/engine-precompiles/src/secp256k1.rs#L84).

engine-precompiles/src/secp256k1.rs
```diff
@@ -118,7 +120,7 @@ mod tests {
     use crate::utils::new_context;
 
     fn ecverify(hash: H256, signature: &[u8], signer: Address) -> bool {
-        matches!(ecrecover(hash, signature), Ok(s) if s == signer)
+        matches!(ecrecover(hash, signature[0..65].try_into().unwrap()), Ok(s) if s == signer)
```
aleksuss, line 123:
```suggestion
        matches!(ecrecover(hash, signature[0..super::consts::SIGNATURE_LENGTH].try_into().unwrap()), Ok(s) if s == signer)
```

aleksuss (approved):

mrLSD (approved):

joshuajbouw (approved):

PR #718, opened 2023-03-07
fix: Lack of Division by `0` Checks In `EthGas::Div()`

0x3bfc:
## Description

In engine-types, the `EthGas` struct does not perform checks on division by `0` in its `Div` implementation.

### DIFF
--- a/engine-types/src/types/gas.rs
+++ b/engine-types/src/types/gas.rs
@@ -1,6 +1,7 @@
 use crate::fmt::Formatter;
 use crate::{Add, AddAssign, Display, Div, Mul, Sub};
 use borsh::{BorshDeserialize, BorshSerialize};
+use core::num::NonZeroU64;
 use serde::{Deserialize, Serialize};
 
 #[derive(
@@ -83,11 +84,10 @@ impl AddAssign for EthGas {
     }
 }
 
-impl Div<u64> for EthGas {
+impl Div<NonZeroU64> for EthGas {
     type Output = Self;
 
-    fn div(self, rhs: u64) -> Self::Output {
-        assert!(rhs != 0, "ZERO_IS_AN_INVALID_DENOMINATOR");
+    fn div(self, rhs: NonZeroU64) -> Self::Output {
         Self(self.0 / rhs)
     }
 }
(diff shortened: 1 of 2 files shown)


engine-types/src/types/gas.rs
```diff
@@ -87,6 +87,7 @@ impl Div<u64> for EthGas {
     type Output = Self;
 
     fn div(self, rhs: u64) -> Self::Output {
+        assert!(rhs != 0, "ZERO_IS_AN_INVALID_DENOMINATOR");
         Self(self.0 / rhs)
```
aleksuss, line 91: I would replace it with [`u64::checked_div`](https://doc.rust-lang.org/std/primitive.u64.html#method.checked_div) method.
```rust
self.0.checked_div(res).map(Self).expect("ZERO_IS_AN_INVALID_DENOMINATOR")
```

engine-types/src/types/gas.rs
```diff
@@ -87,6 +87,7 @@ impl Div<u64> for EthGas {
     type Output = Self;
 
     fn div(self, rhs: u64) -> Self::Output {
+        assert!(rhs != 0, "ZERO_IS_AN_INVALID_DENOMINATOR");
```
birchmd, line 90: Instead of panicking, we could enforce this at the type level by using `rhs: NonZeroU64`. This requires changes in other places as well to map into the non-zero type, but it's not that bad:

```rust
// engine-precompiles/src/alt_bn256.rs

let pair_element_len = NonZeroUsize::try_from(consts::PAIR_ELEMENT_LEN)
    .and_then(NonZeroU64::try_from)
    .map_err(utils::err_usize_conv)?;
```

Changing the type would technically be a public API change, but again I don't think that is a big deal because I doubt anyone is using the library and if they are then changing the call-site is not that hard.

birchmd: Also, I know were are avoiding unsafe conversions (which is why the code in `https://crates.io/crates/const_guards` uses `try_from` conversions), but it looks a little silly when it is a compile-time known constant we are converting.

aleksuss: You mean replace `impl Div<u64> for EthGas` with `impl Div<NonZeroU64> for EthGas` ?
cc @0x3bfc

birchmd (approved):

joshuajbouw (approved):

Aurora-ClickUp:
Task linked: [CU-861mfenq7 AUR-24 Lack of Division by 0 Checks In EthGas::Div()](https://app.clickup.com/t/861mfenq7)

PR #719, opened 2023-03-07
fix: check inputs in `alt_bn256`

0x3bfc:
## Description

- [x] There are no size checks implemented on input array passed to `read_point()` , which may result in panic if accessing elements outside of array boundaries.
- [x] In `concat_low_high`, there are no size checks implemented to ensure provided size of output array `S` is at least twice the size of input arrays sizes `P`.

### DIFF
--- a/engine-precompiles/src/alt_bn256.rs
+++ b/engine-precompiles/src/alt_bn256.rs
@@ -67,6 +67,15 @@ mod consts {
 }
 
 #[cfg(feature = "contract")]
+mod type_arith {
+    pub struct Double<const P: usize>;
+    pub trait Is<const S: usize> {}
+    impl Is<128> for Double<64> {}
+    impl Is<64> for Double<32> {}
+    impl Is<32> for Double<16> {}
+}
+
+#[cfg(feature = "contract")]
 trait HostFnEncode {
     type Encoded;
 
@@ -74,8 +83,10 @@ trait HostFnEncode {
 }
 
 #[cfg(feature = "contract")]
-fn concat_low_high<const P: usize, const S: usize>(low: [u8; P], high: [u8; P]) -> [u8; S] {
-    assert!(S >= (P * 2), "INSUFFICIENT_OUTPUT_ARRAY_SIZE");
+fn concat_low_high<const P: usize, const S: usize>(low: [u8; P], high: [u8; P]) -> [u8; S]
+where
+    type_arith::Double<P>: type_arith::Is<S>,
+{
     let mut bytes = [0u8; S];
     bytes[0..P].copy_from_slice(&low);
     bytes[P..P * 2].copy_from_slice(&high);
--- a/engine-precompiles/src/alt_bn256.rs
+++ b/engine-precompiles/src/alt_bn256.rs
@@ -151,6 +151,9 @@ impl HostFnEncode for bn::G2 {
 /// Reads the `x` and `y` points from an input at a given position.
 fn read_point(input: &[u8], pos: usize) -> Result<bn::G1, ExitError> {
     use bn::{AffineG1, Fq, G1};
+    if input.len() < (pos + consts::SCALAR_LEN * 2) {
+        return Err(ExitError::Other(Borrowed("INVALID_INPUT_LENGTH")));
+    }
 
     let px = Fq::from_slice(&input[pos..(pos + consts::SCALAR_LEN)])
         .map_err(|_e| ExitError::Other(Borrowed("ERR_FQ_INCORRECT")))?;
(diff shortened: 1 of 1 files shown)


engine-precompiles/src/alt_bn256.rs
```diff
@@ -75,6 +75,7 @@ trait HostFnEncode {
 
 #[cfg(feature = "contract")]
 fn concat_low_high<const P: usize, const S: usize>(low: [u8; P], high: [u8; P]) -> [u8; S] {
+    assert!(S >= (P * 2), "INSUFFICIENT_OUTPUT_ARRAY_SIZE");
```
birchmd, line 78: For what it's worth, this could be done as a compile-time check instead since `P` and `S` are const generics:

```rust
#[cfg(feature = "contract")]
mod type_arith {
    pub struct Double<const P: usize>;
    pub trait Is<const S: usize> {}
    impl Is<128> for Double<64> {}
    impl Is<64> for Double<32> {}
    impl Is<32> for Double<16> {}
}

#[cfg(feature = "contract")]
fn concat_low_high<const P: usize, const S: usize>(low: [u8; P], high: [u8; P]) -> [u8; S]
where
    type_arith::Double<P>: type_arith::Is<S>,
{
    let mut bytes = [0u8; S];
    bytes[0..P].copy_from_slice(&low);
    bytes[P..P * 2].copy_from_slice(&high);
    bytes
}
```

I think the compile-time approach has two advantages over the runtime one: (a) no panics, (b) no performance impact. The disadvantage is that the code is a little harder to read. Personally I think it is worth the slightly arcane looking code in case.

birchmd: There is [a more elegant way](https://crates.io/crates/const_guards), but it requires an unstable language feature (and I know we are trying to get off of nightly).

aleksuss: Yes. The coming soon `1.68.0` stabilises `alloc_error_handler`, but `PanicInfo::message(&self)`  is still under the `panic_info_message` feature, unfortunately.

birchmd (approved):

aleksuss (approved):

Aurora-ClickUp:
Task linked: [CU-85zrqmdcd AUR-25 Lack of Input Array Size Checks in alt_bn256::read_point()](https://app.clickup.com/t/85zrqmdcd)

Aurora-ClickUp:
Task linked: [CU-85zrqmdh6 AUR-26 alt_bn256::concat_low_high() May Panic on Output Array of Insufficient Size](https://app.clickup.com/t/85zrqmdh6)

joshuajbouw (approved):

engine-precompiles/src/alt_bn256.rs
```diff
@@ -67,6 +67,15 @@ mod consts {
...
 }
 
+#[cfg(feature = "contract")]
+mod type_arith {
+    pub struct Double<const P: usize>;
+    pub trait Is<const S: usize> {}
+    impl Is<128> for Double<64> {}
+    impl Is<64> for Double<32> {}
+    impl Is<32> for Double<16> {}
+}
```
joshuajbouw, line 77: Whats the motivation?

birchmd: See https://github.com/aurora-is-near/aurora-engine/pull/719#discussion_r1128073168

engine-precompiles/src/alt_bn256.rs
```diff
@@ -67,6 +67,15 @@ mod consts {
     pub(super) const ERR_BIG_ENDIAN: ExitError = ExitError::Other(Borrowed("ERR_BIG_ENDIAN"));
 }
 
+#[cfg(feature = "contract")]
+mod type_arith {
+    pub struct Double<const P: usize>;
+    pub trait Is<const S: usize> {}
+    impl Is<128> for Double<64> {}
+    impl Is<64> for Double<32> {}
+    impl Is<32> for Double<16> {}
```
mandreyel, line 76: Based on the PR description and the previous impl it looks like `S` is allowed to be more than double the size of `P`, but this impl only allows it to be exactly double the size of it.

Is that right? I guess based on the `concat_low_high` function it doesn't really make sense to leave a gap between low and high, but just making sure this is what we want.

birchmd: `contcat_low_high` is a private function, so we only need to enforce constraints based on our own usage. We only use it when `S` is exactly double `P`, so this is the constraint we enforce. If in the future we want to use it for `S` which is greater than `2P` then these compile-time checks will need to made more general (this is still doable).

mandreyel (approved):

PR #722, opened 2023-03-08
Fix: Lack of Minimum Size Checks When Instantiating New EthGas Object

lempire123:
In `engine-standalone-tracing`, when instantiating new `EthGas` objects there are no checks implemented to ensure the passed value is `>= 0` .

Panics will stop execution of the executable, hence any panicking code should be avoided in the release code.


engine-standalone-tracing/src/sputnik.rs
```diff
@@ -51,8 +51,10 @@ impl evm_gasometer::tracing::EventListener for TransactionTraceBuilder {
             Event::RecordCost { cost, snapshot } => {
                 self.current.gas_cost = EthGas::new(cost);
                 if let Some(snapshot) = snapshot {
-                    self.current.gas =
-                        EthGas::new(snapshot.gas_limit - snapshot.used_gas - snapshot.memory_gas);
+                    if let gas = snapshot.gas_limit - snapshot.used_gas - snapshot.memory_gas >= 0 {
```
aleksuss: No. It's wrong. The `gas` var is boolean here. Probably it should be written like this:
```rust
let gas = snapshot.gas_limit - snapshot.used_gas - snapshot.memory_gas;

if gas >= 0 {
    self.current.gas = EthGas::new(gas);
}
```

lempire123: Still getting clippy errors after the fix, says the boolean statement will always resolve to true.

aleksuss: Yes. It's because type `u64` can't be negative 😄

aleksuss:
```rust
if let Some(gas) = snapshot.gas_limit.checked_sub(snapshot.used_gas)
    .and_then(|gas| gas.checked_sub(snapshot.memory_gas)) {
    self.current.gas = EthGas::new(gas);
}
```

lempire123: Yes, thank you, I have been able to get that far in the rust book :) Can I maybe skip storing `snapshot.gas_limit - snapshot.used_gas - snapshot.memory_gas;` in a variable, and instead evaluate if it is `>= 0` directly, or will it not make a difference?

lempire123: Oh, that's nice, thanks for the help @aleksuss!

aleksuss: But the question is, should we skip assigning value in case of overflow, or should we assign `0` to rewrite the previous value of `self.current.gas`?
cc @birchmd

birchmd (changes requested):
This really should never be an issue. The EVM implementation should guarantee that `snapshot.used_gas + snapshot.memory_gas` is always less than or equal to `snapshot.gas_limit`. If we are really concerned about this then I suggest using `saturating_sub` instead of `checked_sub`. Then if some how there would be an overflow the `current_gas` will simply be set to 0.

birchmd (approved):

aleksuss (approved):

joshuajbouw (approved):

PR #724, opened 2023-03-10
fix: check for `0` value as a part of `Depth::decrement()`

0x3bfc:
Given `Depth` is of `u32` type, decrementing it into negative will trigger panic, stopping execution of the executable.

### DIFF
--- a/engine-standalone-tracing/src/types/mod.rs
+++ b/engine-standalone-tracing/src/types/mod.rs
@@ -24,7 +24,9 @@ impl Depth {
     }
 
     pub fn decrement(&mut self) {
-        self.0 -= 1;
+        if !self.is_zero() {
+            self.0 -= 1;
+        }
     }
 
     #[must_use]


engine-standalone-tracing/src/types/mod.rs
```diff
@@ -24,7 +24,9 @@ impl Depth {
     }
 
     pub fn decrement(&mut self) {
-        self.0 -= 1;
+        if !self.is_zero() {
```
birchmd, line 27: Could use `saturating_sub` instead. This is logically equivalent to the current implementation but without the `if` so it is a little less code and a little more efficient at runtime.

0x3bfc: Actually I was thinking the same way. Thanks

joshuajbouw (approved):

aleksuss (approved):

PR #728, opened 2023-03-15
Fix: new() May Be Called Multiple Times

lempire123:
The engine initialisation function `new() `is not limited in the number of times it may be called by the owner.

Recommendation:

Only allow `new() `to be called once at deployment.


Aurora-ClickUp:
Task linked: [CU-8669r8g1y Split functionality into two parts: simple and advanced](https://app.clickup.com/t/8669r8g1y)

PR #729, opened 2023-03-15
Fix: Use the transfer-accept pattern for ownership transfer.

lempire123:
Use the transfer-accept pattern for ownership transfer.

This is implemented by adding a `proposed_owner_id` field to the `EngineState` struct.
The current `owner_id` account can then proposed a new owner, and said new owner can accept ownership.


Aurora-ClickUp:
Task linked: [CU-863g0gk60 AUR-20 new() May Be Called Multiple Times](https://app.clickup.com/t/863g0gk60)

PR #731, opened 2023-03-16
Fix: register_relayer() Functionality is Unnecessary

lempire123:
It is unnecessary to require relayers to register an EVM address before they can receive fees for deposits.

The function `register_relayer()` allows a user to register an EVM address that receives fees from the bridge. Fees are paid to an EVM address so a relayer must have an associated EVM address to receive funds.

However, since `near_account_to_evm_address() `provides a default mapping from a NEAR `AccountId` to an EVM address it is possible to allow relayers to receive funds without first registering an EVM address.


Aurora-ClickUp:
Task linked: [CU-85zrqmerc AUR-32 register_relayer() Functionality is Unnecessary](https://app.clickup.com/t/85zrqmerc)

mrLSD:
Related to #607 relayer will not receive `fee` at all. And some related logic was also deleted.

@lempire123 pls fix the tests. CI failed.

lempire123:
> Related to #607 relayer will not receive `fee` at all. And some related logic was also deleted.
> 
> @lempire123 pls fix the tests. CI failed.

I guess there's no point in fixing this specific report then, given the auditors weren't aware that the NEP-141 logic would be separated?

PR #732, opened 2023-03-20
Chore: update to borsh version 0.10

birchmd:
## Description

This PR updates the Engine contract to use borsh version 0.10. This is needed because [nearcore now uses borsh v0.10](https://github.com/near/nearcore/blob/a6d14feb03af14912f9c6d41857fd4013c8654b8/Cargo.toml#L92) and [borealis-engine-lib](https://github.com/aurora-is-near/borealis-engine-lib) must use both Engine and nearcore types in the same project.

Unfortunately there is a complexity here where the latest [near-sdk still depends on borsh v0.9](https://github.com/near/near-sdk-rs/blob/89f8c2d776c8d00feb45239fdb15725527f1894b/near-sdk/Cargo.toml#L27). Therefore the `aurora-engine-types` crate which is used with near-sdk in some other Near contracts (for example XCC router), actually must support both versions of borsh until there is a new release of near-sdk. This is accomplished by having a `borsh-compat` feature flag in the types crate.

N/A

## Testing

Existing tests.

### DIFF
--- a/engine-tests/src/tests/erc20_connector.rs
+++ b/engine-tests/src/tests/erc20_connector.rs
@@ -682,7 +682,12 @@ pub mod sim_tests {
         // deploy contract with simple exit to near method
         let constructor = TesterConstructor::load();
         let deploy_data = constructor.deploy(0, Address::zero()).data;
-        let submit_result: SubmitResult = aurora.call("deploy_code", &deploy_data).unwrap_borsh();
+        let submit_result = match aurora.call("deploy_code", &deploy_data).status() {
+            near_primitives::transaction::ExecutionStatus::SuccessValue(bytes) => {
+                SubmitResult::try_from_slice(&bytes).unwrap()
+            }
+            other => panic!("Unexpected status {other:?}"),
+        };
         let tester_address =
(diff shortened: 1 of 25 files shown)


engine-tests/src/tests/erc20_connector.rs
```diff
@@ -852,7 +857,12 @@ pub mod sim_tests {
             input: balance_tx.data,
         });
         let result = aurora.call("call", &call_args.try_to_vec().unwrap());
-        let submit_result: SubmitResult = result.unwrap_borsh();
+        let submit_result = match result.status() {
```
birchmd, line 860: This change (and the corresponding ones in `sanity.rs` and `xcc.rs`) is needed because `near-sdk-sim` is still on borsh v0.9. I figured changing this callsite was easier than going through the process of updating `near-sdk-sim` since it will be replaced with `aurora-workspace` soon.

aleksuss (approved):

engine-tests/src/tests/erc20_connector.rs
birchmd: I did consider doing that. But it didn't seem necessary when all the `near-sdk-sim` code is going to be removed soon.

PR #733, opened 2023-03-21
Fix: new() May Be Called Multiple Times

lempire123:
The engine initialisation function `new() `is not limited in the number of times it may be called by the owner.

Recommendations:

Only allow `new() `to be called once at deployment.


joshuajbouw:
Why is this a duplicate of https://github.com/aurora-is-near/aurora-engine/pull/730 ?

engine/src/lib.rs
```diff
@@ -177,6 +178,23 @@ mod contract {
...
 
+    #[no_mangle]
+    pub extern "C" fn get_upgrade_delay_blocks() {
+        let mut io = Runtime;
+        let state = state::get_state(&io).sdk_unwrap();
+        io.return_output(&state.upgrade_delay_blocks.to_le_bytes());
+    }
+
+    #[no_mangle]
+    pub extern "C" fn set_upgrade_delay_blocks() {
```
birchmd, line 189: When you add a new mutable method to the contract you also need to make sure the standalone engine knows how to process it by adding it to the `TransactionKind` enum. See the [set_owner PR for example](https://github.com/aurora-is-near/aurora-engine/pull/690).

birchmd (changes requested):

lempire123:
> Why is this a duplicate of #730 ?

The other PR got several commits behind develop and for some reason wasn't able to sync them so just created a new one, will close the other one now.

engine-standalone-storage/src/sync/types.rs
```diff
@@ -515,6 +518,7 @@ enum BorshableTransactionKind<'a> {
     ResumePrecompiles(Cow<'a, parameters::PausePrecompilesCallArgs>),
     Unknown,
     SetOwner(Cow<'a, parameters::SetOwnerArgs>),
+    SetUpgradeDelayBlocks(Cow<'a, parameters::SetUpgradeDelayBlocksArgs>),
```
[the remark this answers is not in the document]
lempire123: Could you elaborate why it's better to put `SetUpgradeDelayBlocks` after `SubmitWithArgs`?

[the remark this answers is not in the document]
aleksuss: Because of specification of the borsh. Take a look [here](https://github.com/aurora-is-near/aurora-engine/pull/696#discussion_r1113421269).

engine-standalone-storage/src/sync/mod.rs
```diff
@@ -438,6 +438,14 @@ fn non_submit_execute<'db>(
             prev.owner_id = args.clone().new_owner;
             state::set_state(&mut io, &prev)?;
 
+            None
+        }
+        TransactionKind::SetUpgradeDelayBlocks(args) => {
+            let mut prev = state::get_state(&io)?;
+
+            prev.upgrade_delay_blocks = args.clone().upgrade_delay_blocks;
```
aleksuss, line 446: `clone` is redundant here. You copy primitive type, which implements `Copy` trait by default.
```suggestion
            prev.upgrade_delay_blocks = args.upgrade_delay_blocks;
```

joshuajbouw:
Resolve conflicts please.

engine-standalone-storage/src/sync/types.rs
```diff
@@ -518,6 +521,7 @@ enum BorshableTransactionKind<'a> {
     Unknown,
     SetOwner(Cow<'a, parameters::SetOwnerArgs>),
     SubmitWithArgs(Cow<'a, parameters::SubmitArgs>),
+    SetUpgradeDelayBlocks(Cow<'a, parameters::SetUpgradeDelayBlocksArgs>),
```
aleksuss, line 524: It should be after `FundXccSubAccound` because `FundXccSubAccound` is already [merged](https://github.com/aurora-is-near/aurora-engine/pull/735) and might be used already.

aleksuss (changes requested):

aleksuss (approved):

birchmd (approved):
Looks good, thanks @lempire123 !

PR #735, opened 2023-03-23
Feat: allow creating and funding XCC sub-accounts from external NEAR accounts

birchmd:
## Description

The purpose of this PR is to add a new function to the Engine for creating XCC sub-accounts that are funded from an external Near account. The reason this is useful is because it allows the XCC feature to be used in an instance of the Engine where the wNEAR token has not been bridged (this is not the case on mainnet `aurora`, but could be the case on other Engine instances).

Note: this allows creating XCC accounts without requiring the Engine to know about wNEAR, but if a user wants to attach NEAR to their cross-contract calls then then Engine (and that user) will still need wNEAR. This should not be too much of a limitation since many things can be done on Near without spending the base token.

## Testing

One new XCC test to show this works to create a sub-account and use it for XCC in an Engine where wNEAR is not bridged.

### DIFF
--- a/engine-tests/src/tests/xcc.rs
+++ b/engine-tests/src/tests/xcc.rs
@@ -20,6 +20,7 @@ use std::fs;
 use std::path::Path;
 
 const WNEAR_AMOUNT: u128 = 10 * near_sdk_sim::STORAGE_AMOUNT;
+const ONE_NEAR: u128 = 1_000_000_000_000_000_000_000_000;
 
 #[test]
 #[allow(clippy::too_many_lines)]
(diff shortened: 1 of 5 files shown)


aleksuss (approved):
Overall looks good

engine-tests/src/tests/xcc.rs
```diff
@@ -158,6 +159,90 @@ fn test_xcc_eth_gas_cost() {
...
 
+#[test]
+fn test_xcc_external_fund() {
+    // In this test we intentionally do not bridge wNEAR into the Engine.
+    // The purpose of the `fund_xcc_sub_account` functionality is to allow using
+    // the XCC feature in an Engine instance where there is no bridged wNEAR.
+
+    // Set up Engine contract
+    let aurora = deploy_evm();
+    let chain_id = AuroraRunner::default().chain_id;
```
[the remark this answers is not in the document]
birchmd: All the tests for XCC use `near-sdk-sim` (will eventually be replaced with `aurora-workspace`) because they require resolving promises. So the `aurora` instance here is actually [AuroraAccount](https://github.com/aurora-is-near/aurora-engine/blob/56f5c90857edf48252c24cbd2d2f93fb582ff235/engine-tests/src/tests/state_migration.rs#L77) not `AuroraRunner`. The former does not include a `chain_id` field. I suppose we could add that field, but I don't want to put additional effort into the code we have `near-sdk-sim` when it is being replaced.

joshuajbouw (approved):
I do like the version tracking and if deploys need to be made. Good work. Thanks for the tests.

engine-tests/src/tests/xcc.rs
```diff
@@ -158,6 +159,90 @@ fn test_xcc_eth_gas_cost() {
...
+    let sub_account_id = format!(
+        "{}.{}",
+        signer_address.encode(),
+        aurora.contract.account_id.as_str()
+    );
+    let sub_account = aurora
+        .user
+        .borrow_runtime()
+        .view_account(&sub_account_id)
+        .unwrap();
```
joshuajbouw, line 212: Sorry forgot to mention that we shouldn't have unwrap on prod code.

joshuajbouw (changes requested):

engine-tests/src/tests/xcc.rs
aleksuss: Even in tests?

birchmd: Thanks for the review @joshuajbouw ! I think panics in test code is acceptable, but I did remove a couple panics from the production code in https://github.com/aurora-is-near/aurora-engine/pull/735/commits/597114b7802261b8d8a0d70985c6bd73e1d3b970

joshuajbouw: Oh sorry, I jumped the gun a bit given the PR on the other repo. Yeah here it is fine of course.

PR #736, opened 2023-03-29
chore: use owner account in engine authorizer

aleksuss:
## Description

The PR allows using an owner account in the `EngineAuthorizer` rather than a hardcoded `aurora` account.

There should be no changes in the performance.

### DIFF
--- a/engine/src/engine.rs
+++ b/engine/src/engine.rs
@@ -1025,11 +1025,11 @@ pub fn compute_block_hash(chain_id: [u8; 32], block_height: u64, account_id: &[u
 }
 
 #[must_use]
-pub fn get_authorizer() -> EngineAuthorizer {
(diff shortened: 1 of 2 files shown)


birchmd (approved):
LGTM

mandreyel:
There is no performance impact despite reading from storage because this is a read only method, correct?

aleksuss:
I guess you mean increasing gas costs? I think it increases gas costs a bit, even if this is a read-only operation because read-only ops are not free. But we can not worry about that because it's a rare operation (`pause_precompiles`). What we do care about are `submit` and `call` and their inner functions, which work with storage.

mrLSD (approved):

mandreyel (approved):

PR #737, opened 2023-03-30
Fix(modexp): do not return an error when length cannot be cast to usize

birchmd:
## Description

EVM precompiles generally do not return errors; they take whatever input users give and do their best to return a sensible output. For example

```
go-ethereum$ ./build/bin/evm \
    --input 0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000040000000000000000000c000000000000000000000000000000000000000000000000000000000000000071000000000000ff600000000000000000000000000000000000000000000000 \
    --receiver 0x0000000000000000000000000000000000000005 \
    run

0x
```

We see this input calls `modexp` with the size of both base and modulus both equal to zero, and the size of the exponent equal to `u64::MAX`. Since the `modexp` precompile output is bounded by the size of the modulus, we know the output is empty because the size of the modulus is zero, even though an exponent of the requested size could not actually exist in most hardware.

In our implementation we return an error from the `modexp` precompile if the lengths do not fit in a `usize` value, which in the case of Wasm is 32 bits. Therefore our implementation returns an error on the above input instead of retuning the empty value that geth returns. This PR fixes the issue by using a saturating cast into `usize` from the given input. This should be fine because non-trivial inputs (i.e. inputs where the modulus is non-zero) will have lengths that fit into `usize` no problem, or else the computation will run out of gas.

This problem of returning an error for the `usize` cast is actually a little worse than just incompatibility with Ethereum. It also creates an inconsistency with the standalone engine because Wasm has 32-bit `usize` while the standalone tends to run on 64-bit platforms. This means that the above input (with exponent length equal to `u64::MAX`) returns an error in the Wasm contract, but passed in the standalone engine. After the change in this PR it passes on both compilation targets (as verified by the new integration test case).

## Testing

Additional test case in `modexp` integration tests.

### DIFF
--- a/engine-precompiles/src/modexp.rs
+++ b/engine-precompiles/src/modexp.rs
@@ -41,11 +41,11 @@ impl<HF: HardFork> ModExp<HF> {
         }
     }
 
-    fn run_inner(input: &[u8]) -> Result<Vec<u8>, ExitError> {
+    fn run_inner(input: &[u8]) -> Vec<u8> {
         let (base_len, exp_len, mod_len) = parse_lengths(input);
-        let base_len = usize::try_from(base_len).map_err(utils::err_usize_conv)?;
-        let exp_len = usize::try_from(exp_len).map_err(utils::err_usize_conv)?;
-        let mod_len = usize::try_from(mod_len).map_err(utils::err_usize_conv)?;
+        let base_len = usize::try_from(base_len).unwrap_or(usize::MAX);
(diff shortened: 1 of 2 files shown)


0x3bfc (approved):
LGTM

aleksuss (approved):

joshuajbouw (approved):

PR #741, opened 2023-04-09
Fix: deploy_upgrade() Does Not Clear Upgrade Index 

lempire123:
There is an issue in `deploy_upgrade()` that the upgrade index is not cleared after each call.

Therefore, `deploy_upgrade() `will pass the check `io.block_height() <= index + state.upgrade_delay_blocks` since
`index` remains unchanged from previous calls.

### DIFF
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -201,7 +201,7 @@ mod contract {
     /// Deploy staged upgrade.
     #[no_mangle]
     pub extern "C" fn deploy_upgrade() {
-        let io = Runtime;
+        let mut io = Runtime;
         let state = state::get_state(&io).sdk_unwrap();
         require_owner_only(&state, &io.predecessor_account_id());
         let index = internal_get_upgrade_index();
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -209,6 +209,10 @@ mod contract {
             sdk::panic_utf8(errors::ERR_NOT_ALLOWED_TOO_EARLY);
         }
         Runtime::self_deploy(&bytes_to_key(KeyPrefix::Config, CODE_KEY));
+        io.write_storage(
+            &bytes_to_key(KeyPrefix::Config, CODE_STAGE_KEY),
+            &u64::MIN.to_le_bytes(),
+        );
     }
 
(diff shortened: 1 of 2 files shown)


aleksuss (approved):

engine/src/lib.rs
```diff
@@ -201,14 +201,15 @@ mod contract {
...
-        let io = Runtime;
+        let mut io = Runtime;
         let state = state::get_state(&io).sdk_unwrap();
         require_owner_only(&state, &io.predecessor_account_id());
         let index = internal_get_upgrade_index();
         if io.block_height() <= index {
             sdk::panic_utf8(errors::ERR_NOT_ALLOWED_TOO_EARLY);
         }
         Runtime::self_deploy(&bytes_to_key(KeyPrefix::Config, CODE_KEY));
+        io.remove_storage(&bytes_to_key(KeyPrefix::Config, CODE_STAGE_KEY));
```
birchmd, line 212: I'm not sure that we should automatically remove staged code. It could cause an issue for us in the future if we ever have a non-trivial state migration. See [near-plugins documentation](https://github.com/aurora-is-near/near-plugins/blob/master/near-plugins/src/upgradable.rs#L117) for more information.

aleksuss: Then we probably need an additional transaction for removing staged code manually?

birchmd: Couldn't we call `stage_upgrade` with empty input to clear the old staged code?

lempire123: > I'm not sure that we should automatically remove staged code. It could cause an issue for us in the future if we ever have a non-trivial state migration. See [near-plugins documentation](https://github.com/aurora-is-near/near-plugins/blob/master/near-plugins/src/upgradable.rs#L117) for more information.

Might be misunderstanding something, but doesn't` io.remove_storage(&bytes_to_key(KeyPrefix::Config, CODE_STAGE_KEY));`  just clear the upgrade index (ie. the block height at which the staged code is allowed to be deployed), and not the actual staged code?

birchmd: Ah, yes, my mistake! Thanks for pointing that out. This is a nice way to prevent applying an upgrade multiple times without removing the staged code!

birchmd (approved):

PR #744, opened 2023-04-18
chore: add build flow for xcc-router smart contract

aleksuss:
## Description

The PR adds flow for compiling [xcc-router](https://github.com/aurora-is-near/aurora-engine/tree/develop/etc/xcc-router) smart contract.


Makefile.toml
```diff
@@ -262,6 +301,18 @@ script = '''
 docker run --volume $PWD:/host -w /host -i --rm nearprotocol/contract-builder:master-b8fc60809b907543d909ee75fcc1bd7b68cbe2fd-amd64 ./scripts/docker-entrypoint.sh ${PROFILE}
 '''
 
+[tasks.build-xcc-router-docker-inner]
+condition = { profiles = ["mainnet", "testnet"] }
+env = { "RUSTFLAGS" = "-C strip=symbols --remap-path-prefix ${HOME}=/path/to/home/ --remap-path-prefix ${PWD}=/path/to/source/", "CARGO_FEATURES" = "${CARGO_FEATURES_BUILD}", "RELEASE" = "--release", "TARGET_DIR" = "release" }
```
birchmd, line 306: We should add the additional flag `-C link-arg=-s`. This is important to keep the size of the contract small. If the contract is too large then the 2 wNEAR fee we charge to create an XCC account will not be enough to cover the storage cost. The compiled contract size should be less than 150 kB.

birchmd (changes requested):

birchmd (approved):
Nice work!

joshuajbouw (approved):

joshuajbouw:
Note, that the push to master is intentional.

mrLSD (approved):

PR #745, opened 2023-04-21
Fix: Lack of Checks of Array Bounds

lempire123:
Several instances of lack of array bounds checks were identified in the `engine-standalone-storage` and `engine-standalone-tracing` crates.

**Recommendations**

Implement checks to ensure values used when slicing arrays or referencing items from arrays are within array bound- aries.

Use safer alternatives, such as `.get(n..m)` or `.get_mut(n..m) `when accessing array elements.

### DIFF
--- a/engine-standalone-tracing/src/types/mod.rs
+++ b/engine-standalone-tracing/src/types/mod.rs
@@ -233,10 +233,7 @@ impl Index<usize> for Logs {
     type Output = TraceLog;
 
     fn index(&self, index: usize) -> &Self::Output {
-        if index >= self.0.len() {
-            panic!("index out of bounds");
-        }
-        &self.0[index]
+        &self.0.get(index).unwrap()
     }
 }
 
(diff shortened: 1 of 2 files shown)


engine-standalone-tracing/src/types/mod.rs
```diff
@@ -233,7 +233,7 @@ impl Index<usize> for Logs {
     type Output = TraceLog;
 
     fn index(&self, index: usize) -> &Self::Output {
-        &self.0[index]
+        self.0.get(index).unwrap()
```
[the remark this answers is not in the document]
lempire123: Oh right! So maybe use `.expect(<Error message>)` ? To at least give the reason for the error?

engine-standalone-tracing/src/types/mod.rs
```diff
@@ -319,7 +319,7 @@ impl StepTransactionTrace {
     /// `None`.
     #[allow(dead_code)]
     pub fn step(&mut self) -> Option<&TraceLog> {
-        if self.step > self.inner.struct_logs.len() {
+        if self.step > self.inner.struct_logs.len() - 2 {
```
aleksuss, line 322: Why `-2`? I'd like to see some description for that.

lempire123: To avoid index out of bounds. The else block does `self.step += 1;` so has to be - 2. Eg. length is 10, index is 8, becomes 9 after incrementing, so it accesses the last element.

aleksuss: I mean to add that comment to the code.

birchmd (approved):

aleksuss (approved):

mrLSD (approved):

PR #746, opened 2023-04-25
feat: add possibility to use fixed gas cost (silo)

aleksuss:
## Description

The PR adds the possibility to set fixed gas costs per EVM transaction. The feature could be switched on by calling the transaction `set_fixed_gas_cost`. Also, four types of white lists have been added for applying restrictions to deploy code and submit transactions.

**Additions**: The PR introduces a new feature: `ext-connector`, and two additional profiles: `mainnet-silo` and `testnet-silo`. The smart contract built with one of these profiles implies the usage of an external eth-connector.

There are insignificant gas cost increases. That is because we read from the storage fixed gas cost in the execution of the `submit` method. 

## Testing

The unit and integration tests which test the new logic have been added.


engine-types/src/lib.rs
```diff
@@ -36,3 +36,7 @@ mod v0 {
 }
 
 pub use v0::*;
+
+pub trait AsBytes {
```
birchmd, line 40: Why add a new trait as opposed to using `AsRef<[u8]>`?

engine/src/silo/whitelist.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn is_enabled(&self) -> bool {
+        // White list is enabled by default. So return `true` if the key doesn't exist.
+        let key = self.key(STATUS);
+        self.io
+            .read_storage(&key)
+            .map_or(true, |value| value.to_vec() == [1])
+    }
+
+    fn key(&self, value: &[u8]) -> Vec<u8> {
+        let mut bytes = Vec::new();
```
birchmd, line 64: Performance nit: could use `Vec::with_capacity(1 + value.len());`

engine-tests/src/tests/silo.rs
```diff
@@ -0,0 +1,752 @@
...
+
+    // validate post-state
+    validate_address_balance_and_nonce(
+        &runner,
+        sender,
+        INITIAL_BALANCE - TRANSFER_AMOUNT - FEE,
+        (INITIAL_NONCE + 1).into(),
+    );
+    validate_address_balance_and_nonce(&runner, receiver, TRANSFER_AMOUNT, INITIAL_NONCE.into());
+    validate_address_balance_and_nonce(&runner, relayer, ZERO_BALANCE, INITIAL_NONCE.into());
```
birchmd, line 243: Why is the relayer balance still zero? The fee is simply burned?

birchmd (commented):
Very clean code, nice job!

engine-tests/src/tests/silo.rs
aleksuss: Good question. I think in terms of SILO we don't refund any gas to the relayer. But not sure. Good catch. Thanks. I will check it out.

birchmd (approved):

engine/src/engine.rs
```diff
@@ -844,6 +858,19 @@ pub fn submit<I: IO + Copy, E: Env, P: PromiseHandler>(
             .map_err(|_e| EngineErrorKind::InvalidSignature)?
     };
 
+    let fixed_gas_cost = silo::get_fixed_gas_cost(&io);
+    let transaction = if fixed_gas_cost.is_some() {
```
Karkunow, line 862: @aleksuss why do we push gas limit to the max in the case of a Silo?

aleksuss: To exclude a `OutOfGas` error.

mrLSD (approved):

engine-tests/src/tests/promise_results_precompile.rs
```diff
@@ -133,7 +133,7 @@ fn test_promise_result_gas_cost() {
         total_gas1 / NEAR_GAS_PER_EVM
     );
     assert!(
-        test_utils::within_x_percent(6, evm2, total_gas2 / NEAR_GAS_PER_EVM),
```
mandreyel, line 136: Why has this changed?

engine/src/engine.rs
mandreyel: Can you add this as a comment?

engine/src/engine.rs
```diff
@@ -1427,6 +1471,30 @@ unsafe fn schedule_promise_callback<P: PromiseHandler>(
...
 }
 
+fn assert_access<I: IO + Copy, E: Env>(
+    io: &I,
+    env: &E,
+    fixed_gas_cost: &Option<Wei>,
+    transaction: &NormalizedEthTransaction,
+) -> Result<(), EngineError> {
+    if fixed_gas_cost.is_some() {
+        let is_allow = if transaction.to.is_some() {
```
mandreyel, line 1481: Maybe instead of `is_allow` just `allowed` would be more natural sounding.

engine/src/silo/whitelist.rs
```diff
@@ -0,0 +1,156 @@
...
+
+use crate::prelude::Vec;
+use crate::silo::parameters::{WhitelistKindArgs, WhitelistStatusArgs};
+
+const STATUS: &[u8] = b"LIST_STATUS";
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub enum WhitelistKind {
+    Admin = 0x0,
+    EvmAdmin = 0x1,
```
mandreyel, line 14: Can you document what the difference is between admin and EVM admin? In fact, documenting all of them would be nice.

engine/src/silo/whitelist.rs
```diff
@@ -0,0 +1,156 @@
...
+            WhitelistKind::EvmAdmin => 0x1,
+            WhitelistKind::Account => 0x2,
+            WhitelistKind::Address => 0x3,
+        }
+    }
+}
+
+/// `WhiteList` for checking access.
+/// * io - I/O trait handler
+pub struct WhiteList<I> {
```
mandreyel, line 32: nit: whitelist is spelled as a single word so this should be `Whitelist` IMO.
```suggestion
pub struct Whitelist<I> {
```

engine/src/silo/whitelist.rs
```diff
@@ -0,0 +1,156 @@
...
+        let key = self.key(STATUS);
+        self.io.write_storage(&key, &[0]);
+    }
+
+    pub fn is_enabled(&self) -> bool {
+        // White list is enabled by default. So return `true` if the key doesn't exist.
+        let key = self.key(STATUS);
+        self.io
+            .read_storage(&key)
+            .map_or(true, |value| value.to_vec() == [1])
```
mandreyel, line 60: nit: Does it make sense to allocate a vector here?

engine/src/silo/whitelist.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    fn key(&self, value: &[u8]) -> Vec<u8> {
+        let mut bytes = Vec::with_capacity(1 + value.len());
+
+        bytes.push(u8::from(self.kind));
+        bytes.extend_from_slice(value);
+        bytes_to_key(KeyPrefix::WhiteList, &bytes)
```
mandreyel, line 68: nit: Same here, I think this should be `KeyPrefix::Whitelist`.

mandreyel (changes requested):

engine/src/silo/whitelist.rs
aleksuss: Yes. It does. In such a way, we provide a place in the memory via mutable reference to allocated vector for writing bytes that we read from the near storage.

mandreyel: I meant, can we avoid a dynamic allocation just for a comparison? It's probably not a big deal but I'm not sure how much dynamic allocations affect gas usage.

engine/src/silo/whitelist.rs
```diff
@@ -10,9 +10,19 @@ const STATUS: &[u8] = b"LIST_STATUS";
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
 pub enum WhitelistKind {
+    /// The whitelist of this type is for storing NEAR accounts. Accounts stored in this whitelist
+    /// have an admin role. The admin role allows to add new admins and add new entities
+    /// (`AccountId` and `Address`) to whitelists. Also, this role allows to deploy of EVM code
+    /// and submit transactions.
     Admin = 0x0,
+    /// The whitelist of this type is for storing EVM addresses. Addresses included in this
+    /// whitelist can deploy EVM code.
```
mandreyel, line 19: The description here is the same as for `Address`. What is the correct documentation?

engine/src/silo/whitelist.rs
aleksuss: I understand your question and concern 😄 . The short answer is No. We can't.

mandreyel (approved):

.github/workflows/lints.yml
```diff
@@ -29,10 +29,8 @@ jobs:
       - name: Clone the repository
         uses: actions/checkout@v3
       - run: cargo make build-contracts
-      - name: Run Contract cargo clippy
-        run: cargo make clippy
-      - name: Run cargo clippy
-        run: cargo clippy
```
joshuajbouw, line 35: Why is clippy removed here?

aleksuss: Because the `cargo make check` flow includes this kind of check.

engine/src/silo/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    if let Some(cost) = cost {
+        io.write_storage(&key, &cost.to_bytes());
+    } else {
+        io.remove_storage(&key);
+    }
+}
+
+/// Add an entry to a white list depending on a kind of list types in provided arguments.
+pub fn add_entity_to_whitelist<I: IO + Copy>(io: &I, args: &WhitelistArgs) {
```
Karkunow, line 40: @aleksuss should we rename this from add_entity to add_entry? just for consistency

engine/src/silo/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+#[cfg(feature = "contract")]
+pub fn assert_admin<I: IO + Env + Copy>(io: &I) -> Result<(), EngineErrorKind> {
+    if !is_owner(io) && !is_admin(io) {
+        return Err(EngineErrorKind::NotAllowed);
+    }
+
+    Ok(())
+}
+
+/// Check if user has writes to deploy EVM code.
```
Karkunow, line 81: @aleksuss just a small type here "writes" -> "rights"

engine/src/silo/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+pub fn is_allow_deploy<I: IO + Copy>(io: &I, account: &AccountId, address: &Address) -> bool {
+    let admin_list = Whitelist::init(io, WhitelistKind::Admin);
+    let evm_admin_list = Whitelist::init(io, WhitelistKind::EvmAdmin);
+
+    (!admin_list.is_enabled() || admin_list.is_exist(account))
+        && (!evm_admin_list.is_enabled() || evm_admin_list.is_exist(address))
+}
+
+/// Check if user has writes to submit transaction.
```
Karkunow, line 90: @aleksuss same here

engine/src/silo/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+    if !is_owner(io) && !is_admin(io) {
+        return Err(EngineErrorKind::NotAllowed);
+    }
+
+    Ok(())
+}
+
+/// Check if user has writes to deploy EVM code.
+pub fn is_allow_deploy<I: IO + Copy>(io: &I, account: &AccountId, address: &Address) -> bool {
+    let admin_list = Whitelist::init(io, WhitelistKind::Admin);
```
Karkunow, line 83: @aleksuss this part can be simplified a bit if you reuse `is_admin` function here? Optionally, you can also define a similar `is_evm_admin` fn.

aleksuss:
@Karkunow, thank you for the review 👍🏻

mrLSD (approved):

engine-precompiles/src/native.rs
```diff
@@ -553,6 +591,10 @@ impl<I: IO> Precompile for ExitToEthereum<I> {
         let mut input = input;
         let flag = input[0];
         input = &input[1..];
+        #[cfg(not(feature = "ext-connector"))]
+        let eth_connector_account_id = self.current_account_id.clone();
+        #[cfg(feature = "ext-connector")]
+        let eth_connector_account_id = get_eth_connector_contract_account(&self.io)?;
```
birchmd, line 597: I wonder if it makes sense to try an simplify some of this conditional compilation by having the implementation of `get_eth_connector_contract_account` depend on the feature (if it present read from state, otherwise use current account id).

This will make the definition of `get_eth_connector_contract_account` more complicated, but simplify all the call-sites.

engine-standalone-storage/src/relayer_db/mod.rs
```diff
@@ -124,7 +124,7 @@ where
                 if tx_succeeded {
                     println!(
                         "WARN: Transaction with NEAR hash {near_tx_hash:?} expected to succeed, but failed with error message {e:?}",
-                    );
+					);
```
birchmd, line 127: Unnecessary whitespace change. Also one below.

engine-standalone-storage/src/sync/types.rs
```diff
@@ -759,9 +817,10 @@ enum BorshableTransactionKind<'a> {
     PausePrecompiles(Cow<'a, parameters::PausePrecompilesCallArgs>),
     ResumePrecompiles(Cow<'a, parameters::PausePrecompilesCallArgs>),
     Unknown,
+    SetEthConnectorContractAccount(Cow<'a, parameters::SetEthConnectorContractAccountArgs>),
```
birchmd, line 820: I'm worried about the order of the variants here. In borsh encoding the first byte is used to determine which variant of the enum to try to deserialize. But adding this variant above other, already established variants such as `SetOwner` and `PauseContract`, it changes the mapping of that first byte to the different variants. This is a breaking change because old transactions would not be deserialized correctly anymore.

But I'm not exactly sure how to resolve this because I know that silo deployments already existed and would have had some transactions with certain variant identifiers, and those might be different from those in the mainnet engine. If it is true that none of the methods after `SetEthConnectorContractAccount` have been called before then there is no backwards compatibility issue, but if they have been then we might have a conflict between slios and mainnet Aurora.

In the latter case I think mainnet Aurora should take priority since I think its transaction history is more important than silos (I don't think there are any produciton silos yet where the history is critical, but we should double check with @JonathanLogan ).

aleksuss: Good catch 👍🏻 Definitely the`SetEthConnectorContractAccount` must be below existing transactions on the mainnet. Regarding breaking changes on the silo so I discussed it with Sergey and he said that we can make breaking changes in the silo smart contract since we don't have any smart contract in production. So I think it shouldn't be a problem.

engine-precompiles/src/native.rs
aleksuss: I'm not sure that it makes sense. We should add two implementations for the `ExitToNear` and `ExitToEthereum` to get access to the `self.current_account_id` field. So we will have two identical methods instead of one function.

birchmd (approved):

engine-standalone-storage/src/sync/mod.rs
```diff
@@ -424,97 +457,116 @@ fn non_submit_execute<I: IO + Copy>(
...
+            {
+                let handler = crate::promise::NoScheduler { promise_data };
+                contract_methods::connector::ft_resolve_transfer(io, env, &handler)?;
+            }
 
             None
         }
-
         TransactionKind::FtTransfer(_) => {
+            #[cfg(not(feature = "ext-connector"))]
```
karim-en, line 497: Just wondering, why we don't panic if this kind of transaction isn't supported in this mode?

engine-precompiles/src/native.rs
```diff
@@ -327,18 +354,21 @@ impl<I: IO> Precompile for ExitToNear<I> {
         let (refund_address, mut input) = parse_input(input)?;
         #[cfg(not(feature = "error_refund"))]
         let mut input = parse_input(input)?;
-        let current_account_id = self.current_account_id.clone();
+        #[cfg(not(feature = "ext-connector"))]
```
karim-en, line 357: Just wondering, why this check/feature wasn't moved to the `get_eth_connector_contract_account`?

karim-en (approved):
LGTM! just have some minor questions.

engine-tests/src/utils/workspace.rs
```diff
@@ -85,12 +131,14 @@ pub async fn transfer_nep_141_to_erc_20(
 
 pub async fn nep_141_balance_of(nep_141: &RawContract, account_id: &AccountId) -> u128 {
     nep_141
-        .view("ft_balance_of")
+        .call("ft_balance_of") // XCC requires gas
```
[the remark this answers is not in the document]
aleksuss: Because we could make XCC if the eth-connector is external.

engine-standalone-storage/src/sync/mod.rs
aleksuss: It allows us not to modify the code of the `borealis-engine` (we will do it when split the eth-connector from the main smart contract). It will just skip the eth-connector's transactions if we compile the borealis-engine with the feature `ext-connector`.

engine-precompiles/src/native.rs
aleksuss: Because we don't have access to the `self.current_account_id` inside the `get_eth_connector_contract_account`.

PR #749, opened 2023-05-03
Fix: remove unused state variable `bridge_prover_id`

birchmd:
This PR is a modified version of #726 

Normally I would not make a whole separate PR, but I ended up starting this work from scratch off of `develop` because I could not understand why the gas costs in `repro` were increased so much in the old PR. I have since figured out that it was some change in `Cargo.lock` that was causing the problem, but by that point I already had all this code finished in a separate branch so I thought it would be easier to make a separate PR at that point.

Key differences from #726 :

1. The `repro` tests pass without any changes to the gas values (i.e. this implementation does not create a significant gas increase). This is mostly due to leaving `Cargo.lock` untouched, but there are a few other improvements as well (see below).
2. Automatically migrate the state if the V1 state is found. This change means that after the first transaction, the fallback deserialization logic will not be executed and hence reading the state will be as efficient as it always was (as exemplified by the `repro` gas costs remaining unchanged).
3. Use `Cow` in `BorshableEngineState` so that `From<&EngineState> for BorshableEngineState` does not need to clone the owner account id.
4. Ensure removing the `bridge_prover_id` field from `NewCallArgs` is done in a backwards compatible way. This is not a performance improvement, but it is still a critical change. We must ensure that changes to the format of top-level contract functions are always backwards compatible otherwise the Borealis Engine and Borealis Refiner will not be able to parse old transactions (it will fail to deserialize the arguments).


aleksuss:
I'm really curious about the changes in the `Cargo.lock`, which increases the gas cost. Were you able to find out what they are?

birchmd:
> I'm really curious about the changes in the `Cargo.lock`, which increases the gas cost. Were you able to find out what they are?

I didn't dig into it further because there were quite a few changes. I just reverted the whole file back to the version on `develop` and the gas issues went away. But I agree that it would be nice to figure out exactly which dependency was the issue.

engine/src/engine.rs
```diff
@@ -1023,7 +1023,7 @@ pub fn compute_block_hash(chain_id: [u8; 32], block_height: u64, account_id: &[u
 }
 
 #[must_use]
-pub fn get_authorizer<I: IO>(io: &I) -> EngineAuthorizer {
+pub fn get_authorizer<I: IO + Copy>(io: &I) -> EngineAuthorizer {
```
aleksuss, line 1026: Out of curiosity, why did you add the `Copy` bounding for the `I` type?

engine/src/state.rs
```diff
@@ -1,60 +1,160 @@
...
-    /// Use empty to not use base token as bridged asset.
-    pub bridge_prover_id: AccountId,
     /// How many blocks after staging upgrade can deploy it.
     pub upgrade_delay_blocks: u64,
 }
 
-impl From<NewCallArgs> for EngineState {
-    fn from(args: NewCallArgs) -> Self {
+impl EngineState {
+    pub fn to_bytes(&self) -> Result<Vec<u8>, error::EngineStateError> {
```
aleksuss, line 27: The `to_bytes` is an ambiguous name, IMO.  I would rename it to `to_borsh_bytes` or at least add some comment docs to explain what bytes we get here.

aleksuss (approved):
Overall looks good. But with a couple of nitpicks.

engine-tests/src/tests/repro.rs
```diff
@@ -152,6 +152,9 @@ fn repro_common(context: &ReproContext) {
     let tx_hex = std::fs::read_to_string(input_path).unwrap();
     let tx_bytes = hex::decode(tx_hex.trim()).unwrap();
 
+    // Make a random call that touches the Engine state to force the lazy migration
+    runner.call("get_chain_id", "relay.aurora", Vec::new());
```
0x3bfc, line 156: This is interesting. Why I would need to send a random call ?

0x3bfc (approved):

engine-tests/src/tests/repro.rs
aleksuss: There is a comment above 😄

engine/src/engine.rs
birchmd: `Copy` is needed now because I am cheating on Rust mutability in the `get_state` function. In the case where a migration is needed the `get_state` function mutates the state, but since this happens exactly once I figured it would be overkill to change the type signature to `&mut I` and instead I copy `& I` into a mutable instance to pass to `set_state`. I know this bad practice in general, but in this case I think it is acceptable. Though if you feel otherwise then I am willing to change it.

aleksuss: Got it. Thanks.

engine-tests/src/tests/repro.rs
birchmd: By doing the migration in a separate transaction, the cost of the migration is not rolled into the benchmarks. This makes sense because the migration will only happen one time in production, so we care much more about the performance after the migration rather than caring about the migration itself.

engine/src/state.rs
birchmd: Fixed in https://github.com/aurora-is-near/aurora-engine/pull/749/commits/b1015b0fe9e51a30ecd5ae4b2034faaf6fbd26f3

aleksuss (approved):

joshuajbouw:
Wouldn't it be good to add a state migration for this? As in, when stage deploy and self deploy is done, the migration takes place then. That way we can update V1 -> V2.

joshuajbouw (approved):

mrLSD (approved):

PR #750, opened 2023-05-04
Unwrap NEAR on Aurora->NEAR transfers

karim-en:
## Description

NEAR instead of wNEAR on AuroraStoring native NEAR tokens on aurora account on NEAR - issues:
*  This creates issues for accounting and monitoring - need to verify which part of `aurora` balance is used for storage, which is used for bridged NEAR, and which is free and available.
* Could create a  lack of NEAR collateral for the bridged wNEAR - some users may not be able to withdraw their wNEAR from Aurora.

The solution:
* Keep holding wNEAR on `aurora` balance on NEAR side.
	* This way, each time the user bridges NEAR to Aurora, it's being wrapped to wNEAR and sent to Aurora.
	* We preserve the same ERC-20 address of wNEAR on Aurora without any additional steps.
*  When withdrawing wNEAR from Aurora, make unwrap in wNEAR (call  `wrap.near near_withdraw()`) and then send native NEAR tokens for users.
	* Use optional suffix `{}:unwrap` so it won't break the cross-chain swaps for [Ref.Finance](http://ref.finance/).

Implementation:
Modify the `ExitToNear` precompile:
* Call `near_withdraw` instead of `ft_transfer` if the `erc20_address` is a `wnear_address` and the `recipient` end with the suffix `{}:unwrap`.
* Rename `refund_on_error` callback to `exit_to_near_precompile_callback`.
* Pass to it args with optional `TransferNearCallArgs` and an already existing `RefundCallArgs` but make it optional.
* In the `exit_to_near_precompile_callback` transfer near on successful promise result and refund on the failed result.

The gas consumption increased a little bit due to a callback call.

## Testing
Integration tests are added

The reviewer should focus on the changed components that are described in the `implementation` section, and verify that the implementation is backward compatible and doesn't break any other components like `XCC` and `refund` logic.


joshuajbouw:
Why treat WNEAR differently than every other ERC-20 bridged token?

sept-en:
@joshuajbouw this is because $NEAR is a native token in NEAR, and not NEP-141. So this change could allow users to create and/or top-up NEAR accounts easier and in more convenient way (it doesn't require a separate NEAR transaction and thus NEAR gas to unwrap wNEAR from the user).

Also, on NEAR->Aurora transfers of $NEAR we automatically wrap them so users receive wNEAR on Aurora. This PR will allow unwrapping on the way back.

mrLSD (approved):

engine-precompiles/src/native.rs
```diff
@@ -395,30 +442,44 @@ impl<I: IO> Precompile for ExitToNear<I> {
...
         let transfer_promise = PromiseCreateArgs {
             target_account_id: nep141_address,
-            method: "ft_transfer".to_string(),
+            method: method.to_string(),
             args: args.as_bytes().to_vec(),
             attached_balance: Yocto::new(1),
             attached_gas: costs::FT_TRANSFER_GAS,
         };
 
-        #[cfg(feature = "error_refund")]
```
joshuajbouw, line 414: `error_refund` shouldn't be removed.

joshuajbouw (changes requested):

engine-precompiles/src/native.rs
```diff
@@ -357,22 +382,44 @@ impl<I: IO> Precompile for ExitToNear<I> {
 
                 validate_amount(amount)?;
 
-                if let Ok(receiver_account_id) = AccountId::try_from(input) {
+                if let Some(recipient) = parse_recipient(input) {
+                    let (args, method, transfer_near_args) =
+                        if recipient.message.unwrap_or_default() == UNWRAP_WNEAR_MSG
```
aleksuss, line 387:
```suggestion
                        if recipient.message == Some(UNWRAP_WNEAR_MSG)
```

engine-precompiles/src/native.rs
```diff
@@ -243,6 +252,22 @@ fn validate_amount(amount: U256) -> Result<(), ExitError> {
...
+    let mut iter = str::from_utf8(recipient).ok()?.splitn(2, ':');
+    let receiver_account_id = iter.next()?.parse().ok()?;
+    let message = iter.next().map(ToString::to_string);
+
+    Some(Recipient {
+        receiver_account_id,
+        message,
+    })
+}
+
```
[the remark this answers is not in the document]
karim-en: Improved in 6ebfc127dd40a32c3bb4572d185165d6c54ba773

engine-precompiles/src/native.rs
```diff
@@ -356,29 +386,46 @@ impl<I: IO> Precompile for ExitToNear<I> {
                 input = &input[32..];
 
                 validate_amount(amount)?;
+                let recipient = parse_recipient(input)?;
 
-                if let Ok(receiver_account_id) = AccountId::try_from(input) {
+                let (args, method, transfer_near_args) = if recipient.message
```
aleksuss, line 391: We can get rid of the creation `String` here by using [as_deref](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_deref) method.  
```rust
if recipient.message.as_deref() == Some(UNWRAP_WNEAR_MSG) {
...
```

karim-en: 6e294387a84ee2e7fb386a905b1dca812f677c85

engine-precompiles/src/native.rs
```diff
@@ -243,6 +252,27 @@ fn validate_amount(amount: U256) -> Result<(), ExitError> {
     Ok(())
 }
 
+struct Recipient {
+    receiver_account_id: AccountId,
+    message: Option<String>,
+}
+
+fn parse_recipient(recipient: &[u8]) -> Result<Recipient, ExitError> {
```
aleksuss, line 260: Tiny nitpick. I would add some unit tests for this function. I believe It would be trivial.

engine-precompiles/src/native.rs
```diff
@@ -689,4 +753,46 @@ mod tests {
     fn test_exit_with_valid_amount() {
         validate_amount(U256::from(u128::MAX)).unwrap();
     }
+
+    #[test]
+    fn test_parse_recipient() {
+        assert_eq!(
+            parse_recipient("test.near".as_bytes()).unwrap(),
```
aleksuss, line 760:
```suggestion
            parse_recipient(b"test.near").unwrap(),
```

karim-en: 6a9f820b0fb5ab81f9b283937e4913c64b4aaccc

birchmd (changes requested):
Build with the `error_refund` feature is currently failing. (We should probably add this check to CI)

```
[cargo-make] INFO - Execute Command: "cargo" "build" "--target" "wasm32-unknown-unknown" "--release" "--no-default-features" "--features=mainnet,integration-test,error_refund" "-p" "aurora-engine" "-Z" "avoid-dev-deps"
   Compiling aurora-engine-precompiles v1.0.0 (/home/birchmd/upstream/aurora-is-near/aurora-engine/engine-precompiles)
error[E0252]: the name `PromiseWithCallbackArgs` is defined multiple times
  --> engine-precompiles/src/native.rs:25:47
   |
4  |     parameters::{PromiseWithCallbackArgs, RefundCallArgs},
   |                  ----------------------- previous import of the type `PromiseWithCallbackArgs` here
...
25 |         ExitToNearPrecompileCallbackCallArgs, PromiseWithCallbackArgs, TransferNearCallArgs,
   |                                               ^^^^^^^^^^^^^^^^^^^^^^^--
   |                                               |
   |                                               `PromiseWithCallbackArgs` reimported here
   |                                               help: remove unnecessary import
   |
   = note: `PromiseWithCallbackArgs` must be defined only once in the type namespace of this module

warning: unused import: `PromiseWithCallbackArgs`
 --> engine-precompiles/src/native.rs:4:18
  |
4 |     parameters::{PromiseWithCallbackArgs, RefundCallArgs},
  |                  ^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

error[E0308]: mismatched types
   --> engine-precompiles/src/native.rs:449:21
    |
449 |             refund: refund_args,
    |                     ^^^^^^^^^^^ expected `Option<RefundCallArgs>`, found `RefundCallArgs`
    |
    = note: expected enum `Option<RefundCallArgs>`
             found struct `RefundCallArgs`
help: try wrapping the expression in `core::prelude::v1::Some`
    |
449 |             refund: core::prelude::v1::Some(refund_args),
    |                     ++++++++++++++++++++++++           +

Some errors have detailed explanations: E0252, E0308.
For more information about an error, try `rustc --explain E0252`.
warning: `aurora-engine-precompiles` (lib) generated 1 warning
error: could not compile `aurora-engine-precompiles` (lib) due to 2 previous errors; 1 warning emitted
[cargo-make] ERROR - Error while executing command, exit code: 101
[cargo-make] WARN - Build Failed.
```

engine-precompiles/src/native.rs
```diff
@@ -243,6 +252,28 @@ fn validate_amount(amount: U256) -> Result<(), ExitError> {
...
+    receiver_account_id: AccountId,
+    message: Option<String>,
+}
+
+fn parse_recipient(recipient: &[u8]) -> Result<Recipient, ExitError> {
+    let recipient = str::from_utf8(recipient)
+        .map_err(|_| ExitError::Other(Cow::from("ERR_INVALID_RECEIVER_ACCOUNT_ID")))?;
+    let (receiver_account_id, message) = recipient.split_once(':').map_or_else(
+        || (recipient, None),
+        |(recipient, msg)| (recipient, Some(msg.to_string())),
```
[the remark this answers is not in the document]
karim-en: Improved in a615fcb28f07e8470c918d92c559e454dd39df33

engine/src/xcc.rs
```diff
@@ -19,7 +19,7 @@ pub const CODE_KEY: &[u8] = b"router_code";
 pub const VERSION_UPDATE_GAS: NearGas = NearGas::new(5_000_000_000_000);
 pub const INITIALIZE_GAS: NearGas = NearGas::new(15_000_000_000_000);
 pub const UNWRAP_AND_REFUND_GAS: NearGas = NearGas::new(25_000_000_000_000);
-pub const WITHDRAW_GAS: NearGas = NearGas::new(30_000_000_000_000);
+pub const WITHDRAW_GAS: NearGas = NearGas::new(40_000_000_000_000);
```
birchmd, line 22: This change was needed now that more gas is consumed by the withdraw to Near precompile in the case that it needs a callback (either because of the auto-unwrap or because of the error_refund feature).

This change makes everything work, but it makes me realize that there is a more involved change that should be made. Effectively we are duplicating the auto-unwrap logic in both XCC and in the exit precompile now (XCC always auto-unwrapped wNEAR, now the exit precompile does so too). We could remove that duplication by adding `:unwrap` to the exit message and removing the unwrapping logic from XCC itself. That should probably be done as a separate PR though.

aleksuss: I'm surprised that the change doesn't break any test 🤔

aleksuss (approved):

engine-precompiles/src/native.rs
```diff
@@ -395,30 +440,44 @@ impl<I: IO> Precompile for ExitToNear<I> {
...
+        let precompile_call_args = ExitToNearPrecompileCallbackCallArgs {
+            #[cfg(feature = "error_refund")]
+            refund: Some(refund_args),
+            #[cfg(not(feature = "error_refund"))]
+            refund: None,
+            transfer_near: transfer_near_args,
         };
+
+        let callback_promise =
+            if precompile_call_args == ExitToNearPrecompileCallbackCallArgs::default() {
```
mandreyel, line 453: nitpick: Seems a bit cleaner the following way, but your call:
```rust
let is_exit_to_near_precompile = precompile_call_args != ExitToNearPrecompileCallbackCallArgs::default();
let callback_promise = is_exit_to_near_precompile.then(|| PromiseCreateArgs {
    target_account_id: self.current_account_id.clone(),
    method: "exit_to_near_precompile_callback".to_string(),
    args: precompile_call_args.try_to_vec().unwrap(),
    attached_balance: Yocto::new(0),
    attached_gas: costs::EXIT_TO_NEAR_CALLBACK_GAS,
});
```

engine-precompiles/src/native.rs
```diff
@@ -689,4 +750,46 @@ mod tests {
...
+                receiver_account_id: "test.near".parse().unwrap(),
+                message: Some("some_msg:with_extra_colon")
+            }
+        );
+
+        assert_eq!(
+            parse_recipient(b"test.near:").unwrap(),
+            Recipient {
+                receiver_account_id: "test.near".parse().unwrap(),
+                message: Some("")
```
mandreyel, line 784: Should this be `None`?

engine-tests/src/tests/erc20_connector.rs
```diff
@@ -482,6 +494,104 @@ pub mod sim_tests {
...
+        );
+
+        // Check wnear balances
+        assert_eq!(
+            nep_141_balance_of(ft_owner.account_id.as_str(), &wnear, &aurora),
+            FT_TOTAL_SUPPLY - FT_TRANSFER_AMOUNT
+        );
+        assert_eq!(
+            nep_141_balance_of(aurora.contract.account_id.as_str(), &wnear, &aurora),
+            FT_TRANSFER_AMOUNT - FT_EXIT_AMOUNT
```
mandreyel, line 538: I was a bit confused about this constant due to its naming. I know this PR didn't introduce it, so just commenting: would it be more accurate to name this constant `FT_ERC20_AMOUNT`?

engine-tests/src/tests/erc20_connector.rs
```diff
@@ -482,6 +494,104 @@ pub mod sim_tests {
...
+        let total_tokens_burnt: u128 = result
+            .promise_results()
+            .iter()
+            .map(|r| r.as_ref().unwrap().outcome().tokens_burnt)
+            .sum();
+
+        // Check that the wnear tokens are properly unwrapped and transferred to `ft_owner`
+        assert_eq!(
+            get_account_balance(ft_owner.account_id.as_str(), &aurora),
+            ft_owner_balance - total_tokens_burnt + FT_EXIT_AMOUNT
```
mandreyel, line 528: I'm new to this part of the code base and I'm missing some context. What does it mean for tokens to be burned here?

engine-tests/src/tests/erc20_connector.rs
```diff
@@ -701,7 +812,47 @@ pub mod sim_tests {
...
+        assert_eq!(
+            nep_141_balance_of(aurora.contract.account_id.as_str(), &wnear, &aurora),
+            FT_TRANSFER_AMOUNT
+        );
+        assert_eq!(
+            erc20_balance(&wnear_erc20, ft_owner_address, &aurora),
+            FT_TRANSFER_AMOUNT.into()
+        );
+
+        // 5. Deploy NEP-141
```
mandreyel, line 855: Is this just some random token (so not wNEAR)? What is the purpose of it? Most of the steps look the same to the above wNEAR deployment so I'm not sure what this means.

engine-tests/src/tests/erc20_connector.rs
```diff
@@ -744,16 +895,27 @@ pub mod sim_tests {
             nep_141,
             erc20,
             aurora,
+            wnear,
+            wnear_erc20,
         }
     }
 
+    fn get_account_balance(account: &str, aurora: &AuroraAccount) -> u128 {
```
mandreyel, line 903: This returns the amount of NEAR stored on the Aurora account on NEAR, correct?

karim-en: Yes, it is.

engine-tests/src/tests/erc20_connector.rs
karim-en: The `wNEAR` is more than just nep141 token, it is a native near custodian, so it has extra api/methods.
Anyway, it is a generic context for the tests, so we need to deploy both of them to cover the nep141 and wNEAR tokens behaviors.

engine-tests/src/tests/erc20_connector.rs
karim-en: It is the amount of native near tokens burnt corresponding to the burnt gas amount.

engine-tests/src/tests/erc20_connector.rs
karim-en: To be honest, this is a subjective perception, the `FT_ERC20_AMOUNT` is also a little bit confusing because it sounds like the origin of the token is a native erc20 token that deployed on near.

engine-precompiles/src/native.rs
karim-en: I think it should be an empty string due to the colon at the end.

engine-precompiles/src/native.rs
karim-en: I've moved this part of code down to improve the readability 870f97a9d33b648be8f7e1b0d0978eea05d0fb69

mandreyel (approved):

joshuajbouw:
Looks like CI is still failing.

joshuajbouw:
Can you resolve conflicts please? Thanks.

mrLSD (approved):

birchmd (changes requested):
PR looks good except for one mistake in the `TransactionKindTag`

birchmd (approved):

mrLSD (approved):

PR #751, opened 2023-05-05
chore: update near dependencies and bump rust version

aleksuss:
## Description

The PR updates near's crates from crates.io rather than our forks. The dependencies are used in the tests. Also, the PR bumps the version of the nightly rust compiler.  

A tiny gas costs regression occurs probably because of updating some dependency.

### DIFF
--- a/engine-precompiles/src/xcc.rs
+++ b/engine-precompiles/src/xcc.rs
@@ -31,7 +31,7 @@ pub mod costs {
     ///
     /// This process is done in the `test_xcc_eth_gas_cost` test in
     /// `engine-tests/src/tests/xcc.rs`.
-    pub const CROSS_CONTRACT_CALL_BASE: EthGas = EthGas::new(323_000);
+    pub const CROSS_CONTRACT_CALL_BASE: EthGas = EthGas::new(343_650);
     /// Additional EVM gas cost per bytes of input given.
     /// See `CROSS_CONTRACT_CALL_BASE` for estimation methodology.
     pub const CROSS_CONTRACT_CALL_BYTE: EthGas = EthGas::new(3);
--- a/engine-tests/src/test_utils/mod.rs
+++ b/engine-tests/src/test_utils/mod.rs
@@ -1,8 +1,9 @@
-use aurora_engine::engine::EngineError;
+use aurora_engine::engine::{EngineError, EngineErrorKind, GasPaymentError};
 use aurora_engine::parameters::{SubmitArgs, ViewCallArgs};
 use aurora_engine_types::account_id::AccountId;
 use aurora_engine_types::types::{NEP141Wei, PromiseResult};
 use borsh::{BorshDeserialize, BorshSerialize};
+use evm::ExitFatal;
 use libsecp256k1::{self, Message, PublicKey, SecretKey};
 use near_primitives::runtime::config_store::RuntimeConfigStore;
 use near_primitives::version::PROTOCOL_VERSION;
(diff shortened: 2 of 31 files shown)


engine-tests/src/test_utils/mod.rs
```diff
@@ -112,29 +113,28 @@ impl<'a> OneShotAuroraRunner<'a> {
...
+    ) -> anyhow::Result<VMOutcome> {
         AuroraRunner::update_context(
             &mut self.context,
             caller_account_id,
             caller_account_id,
             input,
         );
 
-        match near_vm_runner::run(
+        let outcome = near_vm_runner::run(
```
birchmd, line 137: I'm curious how this is able to correctly pass on all the errors. It seems like it is working via `anyhow` magic, but I don't understand how it works. There is a new `aborted` field in [VMOutcome](https://docs.rs/near-vm-logic/0.16.1/near_vm_logic/struct.VMOutcome.html) which captures [FunctionCallError](https://docs.rs/near-vm-errors/0.16.1/near_vm_errors/enum.FunctionCallError.html)s like panics in our contract (which would be raised by not enough intrinsic gas for example). How does `anyhow` know that this field being set means that it should convert to the error variant?

birchmd (approved):
Nice work! I like that we can use the cargo-published Near dependencies now!

engine-tests/src/test_utils/mod.rs
aleksuss: 1. `VMRunnerError` derives `thiserror::Error`, which makes `VMRunnerError` suitable to use in `anyhow::Rusult`.
2. There is no magic, by the way. We just use the `aborted` field for checking errors. Like [this](https://github.com/aurora-is-near/aurora-engine/blob/chore/aleksuss/update_tests/engine-tests/src/tests/ghsa_3p69_m8gg_fwmf.rs#L46).
3. We just convert anyhow::Result<VMOutcome> into anyhow::Result<VMOutcome, EngineError> [here](https://github.com/aurora-is-near/aurora-engine/blob/chore/aleksuss/update_tests/engine-tests/src/test_utils/mod.rs#L235).

But obviously, the situation should be improved, and I have some thoughts about that.

birchmd: Ok, no magic involved, we need to inspect the `field` directly. I think we should do that directly in this function instead of requiring it to be checked at the call-site. Something like

```rust
let outcome = near_vm_runner::run(...)?;
if let Some(error) = outcome.aborted {
    return Err(error.into());
}
outcome
```

aleksuss: If `outcome.aborted` is `Some`, the standalone won't process a transaction, but it must. I guess.

birchmd: The standalone should not process transactions which are aborted. In production, the Borealis Engine skips all failed Near transactions because they have no impact on the Aurora Engine state.

mrLSD (approved):

engine-tests/src/test_utils/mod.rs
aleksuss: I mean *try to process* and of course, we'll get the `EngineError` as a result of standalone execution if the `FunctionCallError` is some. But obviously, it's better to try to convert the `FunctionCallError` into `EngineError` and return it before the standalone execution.

P.S. I know that It looks like a cheating approach for getting `EngineError` without converting `FunctionCallError` 😄

birchmd (approved):

engine-precompiles/src/xcc.rs
```diff
@@ -31,7 +31,7 @@ pub mod costs {
     ///
     /// This process is done in the `test_xcc_eth_gas_cost` test in
     /// `engine-tests/src/tests/xcc.rs`.
-    pub const CROSS_CONTRACT_CALL_BASE: EthGas = EthGas::new(323_000);
+    pub const CROSS_CONTRACT_CALL_BASE: EthGas = EthGas::new(343_650);
```
mandreyel, line 34: I guess this is what you were talking about. It's a little concerning that a dependency update would so significantly increase gas cost.

aleksuss: Yes. Exactly.

birchmd: Yes, raising this in particular is bad because it breaks backwards compatibility (the same transaction prior to this change will cost less than after this change). But XCC was only just released on mainnet so I think it's ok to change it now.

mandreyel (approved):

PR #762, opened 2023-05-18
chore: add bench-modexp flow to CI script

aleksuss:
## Description

The goal of the PR is to speed up passing CI flow. Also, it adds an additional job for testing the `bech_modexp` flow.

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -31,8 +31,6 @@ jobs:
       - run: cargo make build-contracts
       - name: Run Contract cargo clippy
         run: cargo make clippy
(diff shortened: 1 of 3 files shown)


aleksuss:
I thought to speed up the CI flow by adding a matrix with profiles (mainnet and testnet), but I had no effect because the build server doesn't allow running so many parallel jobs. So the pr only adds an additional job for testing the `bench-modexp` flow. But if it's unimportant, the PR could be closed. 
cc @birchmd

birchmd (approved):

mrLSD (approved):

PR #763, opened 2023-05-22
Implement `serd::Deserialize` for `AccountId`

karim-en:
## Description
The AccountId validation is missing on parsing a json input, this leads to the possibility to call `ft_transfer` with an invalid `receiver_id`. here is an [example](https://explorer.near.org/transactions/HKLVudra9nYdtuFBncFipSveZ5d6D1TbHGfAnM4S2rWg) that we received.

This pull request removes the default implementation of the `serd::Deserialize` and implements a custom one.

### DIFF
--- a/engine-types/src/account_id.rs
+++ b/engine-types/src/account_id.rs
@@ -16,17 +16,7 @@ pub const MAX_ACCOUNT_ID_LEN: usize = 64;
 ///
 /// This guarantees all properly constructed `AccountId`'s are valid for the NEAR network.
 #[derive(
-    Default,
-    BorshSerialize,
-    Serialize,
-    Deserialize,
-    Eq,
-    Ord,
-    Hash,
-    Clone,
-    Debug,
-    PartialEq,
-    PartialOrd,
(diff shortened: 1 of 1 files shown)


birchmd (approved):

aleksuss (commented):
Overall looks good. But my only concern is that it would be inconsistent with our borsh version of the deserialization. In the borsh version, we [don't validate](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-types/src/account_id.rs#L107) `AccountId` for an empty string because as I remember we have such cases on chain.

As an alternative, you can use `#[serde(deserialize_with = "path")]` on a field of the structure used in the `ft_transfer` method. It allows leaving default deserialization and applying more strict rules for concrete structs, and it would be more flexible.
cc @joshuajbouw @birchmd

sept-en (approved):

mrLSD (approved):

frol:
FYI, I would recommend just using/re-exporting [`near-account-id` crate](https://crates.io/crates/near-account-id).

joshuajbouw (approved):

PR #765, opened 2023-05-25
Restrict receiving erc20 tokens

karim-en:
## Description
When users bridge from Ethereum to Aurora Cloud (some specific silo), we need to verify that the recipient, are whitelisted in that specific Silo (if this whitelist mode is enabled). This is required to prevent non-whitelisted users from receiving tokens on the silo and breaking permission requirements, also having their tokens stuck. When a transfer happens and the recipient  (and sender if feasible) don't satisfy our whitelisting requirements, that transfer is blocked, meaning that the recipient won't receive his tokens on the Silo.

It's also important that when the transfer is blocked, we instead transfer blocked tokens to our fallback address for tokens. This will help to make the manual recovery easier and won't add a significant security risk (as opposed to fallback just to the Silo/Aurora account itself and using admin access for the recovery).

There are insignificant gas cost increases. That is because we read from the storage the ERC20 fallback address in the execution of the `ft_on_transfer -> receive_erc20_tokens` method. 

## Testing

The integration tests that test the new logic have been added.

## Additional information
Implementation:
- Added new API `set_silo_params` to init/enable/disable the silo mode.
- The `set_fixed_gas_cost` is left for backward compatibility, but it can't be used to enable/disable the silo mode anymore.
- Verify if the recipient address is whitelisted on `ft_on_transfer -> receive_erc20_tokens` call, if it is not, then replace the passed `recipient` with the `fallback_address`

The reviewer should focus on the changed components that are described in the `implementation` section, and verify that the implementation is backward compatible, and the integrity of the storage was preserved.


engine/src/engine.rs
```diff
@@ -723,6 +723,12 @@ impl<'env, I: IO + Copy, E: Env> Engine<'env, I, E> {
             )))
         };
 
+        if let Some(fallback_address) = silo::get_erc20_fallback_address(&self.io) {
+            if !silo::is_allow_receive_erc20_tokens(&self.io, &recipient) {
```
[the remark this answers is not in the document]
karim-en: But why the `erc20_fallback_address` can't be also an indicator of silo mode on?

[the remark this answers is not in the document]
aleksuss: Ok. The question is follow. Is it possible case when the silo mode is on but the `erc20_fallback_address` is not set? Or `erc20_fallback_address` must be set always in silo mode?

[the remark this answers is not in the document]
karim-en: it's hard for me to unequivocally answer, but yeh, I think the `erc20_fallback_address` should be used only if the silo mode is on.
Let's make things less flexible to avoid vulnerabilities, so let's add a function `is_silo_mode_enabled()` and use it instead of the `fixed_gas_cost` everywhere.
The `is_silo_mode_enabled` could be implemented in two way just `fixed_gas_cost.is_some()` or `fixed_gas_cost.is_some() && erc20_fallback_address.is_some()`.

also, we can combine the `fixed_gas_cost` and `erc20_fallback_address` into one struct under one storage key.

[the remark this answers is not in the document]
aleksuss: Then I would propose the following changes. Change the structure of the input parameter of the `set_fixed_gas_cost` function by adding one more field: `fallback_address`. Obviously, `set_fixed_gas_cost` should be renamed to `set_silo_params` or something. Also, we should not be able to set fixed costs without a fallback address. In such a way, presenting a fixed cost or fallback address means that silo mode is on. It allows us to eliminate `is_silo_mode_enabled`, which adds additional storage read.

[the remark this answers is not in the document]
karim-en: Yeh, sounds good!

engine-tests/src/tests/silo.rs
```diff
@@ -762,3 +765,219 @@ fn call_function<T: BorshSerialize + Debug>(runner: &mut AuroraRunner, func: &st
...
+            aurora,
+            fallback_account,
+            fallback_address,
+            ft_owner,
+            ft_owner_address,
+            nep_141,
+            erc20,
+        } = init_silo();
+
+        set_whitelist_status(&aurora, WhitelistKind::Address, true);
```
aleksuss, line 798: Nitpick. It's redundant. By default, whitelists are enabled.

aleksuss (commented):
Overall looks good. But I would add a test with a positive case in the silo module.

aleksuss (approved):

engine-standalone-storage/src/sync/types.rs
```diff
@@ -133,7 +133,7 @@ pub enum TransactionKind {
     FactorySetWNearAddress(Address),
     FundXccSubAccount(FundXccArgs),
     /// Silo operations
-    SetFixedGasCost(silo::parameters::FixedGasCostArgs),
+    SetSiloParams(Option<silo::parameters::SiloParamsArgs>),
```
birchmd, line 136: This is technically a breaking change because the `set_fixed_gas_cost` is being removed from the list of known transactions. This means we will no longer be able to index old transactions calling that function using the Borealis infrastructure (refiner/engine). This might be ok if we will never need to index old transactions because the first production silo is not yet deployed. But I wanted to call it out in case this breaking change might cause a problem for any demos we have set up.

karim-en: We can return the `set_fixed_gas_cost` back but with an extra check for `erc20_fallback_address != None` if the passed  `fixed_gas_cost != None`. and set both values as None if the passed `fixed_gas_cost` is None.

aleksuss: Sounds good to me.

karim-en: Here is the implementation https://github.com/aurora-is-near/aurora-engine/pull/765/commits/2ea368ceed5f1440b5f57702c5df19c63ed13683.

For the existing demo, the admin needs to call the `set_silo_params` to be able to call the `set_fixed_gas_cost` again.

birchmd: Yes, to be backwards compatible (able to index all previous transactions) we must include all past and present top-level contract functions in `TransactionKind`.

joshuajbouw:
Karim please adhere to the rules of the PR and do include the rest of the information in the description of the PR. There a number of areas missing as it is a requirement for all code changes to include a full completion of the PR template.

engine-standalone-storage/src/sync/types.rs
karim-en: Reverted in bd13bf876a88b80552fb961e29a0a12e26919c8d

karim-en: @birchmd @aleksuss reverted in bd13bf876a88b80552fb961e29a0a12e26919c8d

birchmd (approved):

mrLSD (approved):

PR #769, opened 2023-06-01
Feat(precompiles): Abstract modexp impl

birchmd:
## Description

This makes the modexp precompile take a generic type parameter for the implementation of the algorithm. This allows the standalone engine (or other downstream users of Aurora Engine) to select their own modexp implementation. The implementation used for our Wasm contract remains unchanged (the generic is filled with a type using the implementation we already had).

N/A

## Testing

N/A

### DIFF
--- a/engine/src/engine.rs
+++ b/engine/src/engine.rs
@@ -1,6 +1,7 @@
 use crate::parameters::{
     CallArgs, NEP141FtOnTransferArgs, ResultLog, SubmitArgs, SubmitResult, ViewCallArgs,
 };
+use aurora_engine_types::PhantomData;
 use core::mem;
 use evm::backend::{Apply, ApplyBackend, Backend, Basic, Log};
 use evm::executor;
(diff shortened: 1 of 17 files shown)


joshuajbouw (approved):
Good work, and thanks for acting fast on this.

engine/src/engine.rs
```diff
@@ -958,7 +965,7 @@ pub fn refund_on_error<I: IO + Copy, E: Env, P: PromiseHandler>(
     if let Some(erc20_address) = args.erc20_address {
         // ERC-20 exit; re-mint burned tokens
         let erc20_admin_address = current_address(&current_account_id);
-        let mut engine =
+        let mut engine: Engine<I, E, aurora_engine_modexp::AuroraModExp> =
```
mandreyel, line 968: Above the engine is instantiated with a generic `ModexpAlgorithm` type parameter (e.g. in `submit`), but here a concrete implementation is used. Why?

aleksuss: Because another implementation of `ModExp` is needed in the Borealis infrastructure only.

birchmd:
Closing in favour of https://github.com/aurora-is-near/aurora-engine/pull/770

PR #772, opened 2023-06-07
chore: move input parameters to aurora-engine-types

aleksuss:
## Description

The goal of the PR is to allow to use input parameters of the engine without importing the `aurora-engine` itself. 

There are no changes because it's refactoring only.

## Testing

Some tests were moved only. 

Pay attention that parameters correctly have been placed in corresponding modules.


mandreyel (approved):

engine-types/src/parameters/connector.rs
```diff
@@ -0,0 +1,250 @@
+use base64::Engine;
+#[cfg(not(feature = "borsh-compat"))]
+use borsh::{BorshDeserialize, BorshSerialize};
+#[cfg(feature = "borsh-compat")]
+use borsh_compat::{self as borsh, BorshDeserialize, BorshSerialize};
+use primitive_types::{H160, H256};
```
mrLSD, line 6: I think better:
```
use crate::{H160, H256};
```

engine-types/src/parameters/mod.rs
```diff
@@ -1,330 +1,5 @@
-use crate::{
-    account_id::AccountId,
-    types::{Address, NEP141Wei, NearGas, RawU256, Yocto},
-    Box, String, Vec,
-};
-#[cfg(not(feature = "borsh-compat"))]
-use borsh::{maybestd::io, BorshDeserialize, BorshSerialize};
-#[cfg(feature = "borsh-compat")]
-use borsh_compat::{self as borsh, maybestd::io, BorshDeserialize, BorshSerialize};
+pub use promise::*;
```
mrLSD, line 1: Why it's a wildcard?

mrLSD: Maybe just:
```
pub use promise;
```
It will be more precise in importing.

aleksuss: That's for saving backward compatibility.

mrLSD (approved):

PR #777, opened 2023-06-20
chore: engine integration tests refactoring

aleksuss:
## Description

The main goal of the PR is to replace the deprecated `near-sdk-sim` crate with the `aurora-engine-workspace` crate, which is based on [workspaces](https://github.com/near/workspaces-rs). Also, some test files have been restructured. 

No changes in gas costs.


Cargo.toml
```diff
@@ -40,6 +40,7 @@ members = [
     "engine-tests",
     "engine-transactions",
     "engine-types",
+    "engine-workspace",
```
birchmd, line 43: This is now going to replace https://github.com/aurora-is-near/aurora-workspace right?

birchmd (approved):
Great job! It's amazing to finally have this done ❤️‍🔥

Cargo.toml
aleksuss: Yes. But this, internal crate is related to the engine only. Also, we have plans to continue working on the aurora-workspace by making it less dependable on the aurora-engine-types and more undependable.

mandreyel: Why duplicate the two? Can't the engine use aurora-workspace?

aleksuss: No. It can't. The reason is the issue related to the usage of `aurora-engine-types` from different sources. In the tests, we use the local path to the crate, but in the `aurora-workspace`, we use it from the git, and so we've got different versions of types in the engine tests.

mandreyel: Would it be possible for engine tests to use engine types defined in aurora workspace? That way we also ensure the workspace is kept up to date. Just an idea.

aleksuss: It's not so easy. I thought about this. But the problem is that we use these types further in the code in functions whose parameters are from the local crate.

engine-workspace/src/contract.rs
```diff
@@ -0,0 +1,393 @@
...
+    pub fn ft_transfer_call(
+        &self,
+        receiver_id: &AccountId,
+        amount: U128,
+        memo: Option<String>,
+        msg: String,
+    ) -> CallFtTransferCall {
+        CallFtTransferCall::call(&self.contract).args_json(json!({
+           "receiver_id": receiver_id,
+           "amount": amount.0.to_string(),
```
mrLSD, line 104: We don't need to unwrap `U128` and convert it to `String`. `U128` serialized to `String` itself.

engine-workspace/src/contract.rs
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    pub fn deposit(&self, raw_proof: Proof) -> CallDeposit {
+        CallDeposit::call(&self.contract).args_borsh(raw_proof)
+    }
+
+    pub fn set_eth_connector_contract_data(
+        &self,
+        prover_account: AccountId,
+        eth_custodian_address: String,
```
mrLSD, line 138: Why it's `String`?

engine-workspace/src/contract.rs
```diff
@@ -0,0 +1,393 @@
...
+
+    pub fn state_migration(&self) -> CallStateMigration {
+        CallStateMigration::call(&self.contract)
+    }
+
+    pub fn mint_account(
+        &self,
+        address: Address,
+        init_nonce: u64,
+        init_balance: u64,
```
mrLSD, line 252: We have the `Balance` type.

engine-workspace/src/contract.rs
```diff
@@ -0,0 +1,393 @@
...
+    ) -> CallMintAccount {
+        CallMintAccount::call(&self.contract).args_borsh((address, init_nonce, init_balance))
+    }
+
+    pub fn set_paused_flags(&self, flags: u8) -> CallSetPausedFlags {
```
mrLSD, line 257: AFAIR we have a special type for it.

joshuajbouw: For `CallSetPausedFlags`?

engine-workspace/src/contract.rs
aleksuss: Because of [this](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-types/src/parameters/connector.rs#L16).

engine-workspace/src/contract.rs
aleksuss: Yes, we have. But the engine waits for `u64` in this [case](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine/src/lib.rs#L1027).

engine-workspace/src/contract.rs
aleksuss: Yes. We have an alias `pub type PausedMask = u8;`

engine-workspace/src/contract.rs
aleksuss: Good catch.

mrLSD (approved):

joshuajbouw:
Let me know when this is rebased, I'll be pulling it in and including for the release.

aleksuss:
@joshuajbouw Done.

joshuajbouw (approved):

PR #778, opened 2023-06-21
Fix(modexp): use ibig implementation

birchmd:
## Description

Switch to using the `ibig` implementation for modexp. Its performance is better than current Aurora implementation in some (but not all) cases, and it is more battle-tested in terms of correctness.

### DIFF
--- a/engine-modexp/Cargo.toml
+++ b/engine-modexp/Cargo.toml
@@ -13,7 +13,7 @@ publish = false
 autobenches = false
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


joshuajbouw (approved):

joshuajbouw:
Ensure to note the TODO here and make an issue so that we can track it.

joshuajbouw (approved):

aleksuss (approved):

mrLSD (approved):

PR #789, opened 2023-06-28
chore: get rid of the byte-slice-cast crate in engine

aleksuss:
## Description

The PR removes the redundant `byte-slice-cast` crate from `aurora-engine` as it is used only in one place in the code. Also, vectors have been replaced by fixed-size arrays.

No performance changes.

## Testing

It doesn't require adding new tests.


mrLSD (approved):

joshuajbouw (approved):
Good improvement.

PR #803, opened 2023-07-25
fix: push0 should increment program counter by 1 instead of 2

mandreyel:
## Description

The PR bumps the `sputnikvm` version, which contains a bugfix of the `shanghai` implementation.

## Testing

Added test which validates EVM transaction created with parameter: `evmVersion: 'shanghai'`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -327,7 +327,7 @@ dependencies = [
 
 [[package]]
 name = "aurora-engine"
-version = "2.9.2"
+version = "2.10.0"
 dependencies = [
  "aurora-engine-modexp",
  "aurora-engine-precompiles",
@@ -1720,7 +1720,7 @@ checksum = "0206175f82b8d6bf6652ff7d71a1e27fd2e4efde587fd368662814d6ec1d9ce0"
 [[package]]
 name = "evm"
 version = "0.39.0"
-source = "git+https://github.com/aurora-is-near/sputnikvm.git?tag=v0.38.0-aurora#fcc538cc1f7f91156ef66564a8ac032a82bd4b9e"
+source = "git+https://github.com/aurora-is-near/sputnikvm.git?tag=v0.38.0-aurora#c5cdf0c9bf8da9f6baeb1e5b4a52a2a3405e123f"
 dependencies = [
  "auto_impl",
  "environmental",
@@ -1740,7 +1740,7 @@ dependencies = [
 [[package]]
 name = "evm-core"
 version = "0.39.0"
-source = "git+https://github.com/aurora-is-near/sputnikvm.git?tag=v0.38.0-aurora#fcc538cc1f7f91156ef66564a8ac032a82bd4b9e"
+source = "git+https://github.com/aurora-is-near/sputnikvm.git?tag=v0.38.0-aurora#c5cdf0c9bf8da9f6baeb1e5b4a52a2a3405e123f"
 dependencies = [
  "parity-scale-codec 3.6.3",
  "primitive-types 0.12.1",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -327,7 +327,7 @@ dependencies = [
 
 [[package]]
 name = "aurora-engine"
-version = "2.9.2"
+version = "2.10.0"
```
aleksuss, line 330: Why is this changed?

mandreyel: Not sure actually! I just ran `cargo update -p evm`.

mandreyel: Oh I know, it's because the version in `engine/Cargo.toml` has been changed to that version but apparently it hasn't been built since then and thus Cargo.lock was never updated. Think it's fine.

aleksuss: Yes. I see.

engine-tests/src/tests/sanity.rs
```diff
@@ -694,6 +694,27 @@ fn test_eth_transfer_incorrect_nonce() {
     utils::validate_address_balance_and_nonce(&runner, dest_address, Wei::zero(), 0.into());
 }
 
+#[test]
+fn test_shanghai_contract() {
+    let (mut runner, mut source_account, _) = initialize_transfer();
+
+    // Panicked with `EvmFatal(Other("internal error: entered unreachable code [engine/src/engine.rs:148:54]"))`
```
birchmd, line 701: Probably don't need this comment anymore

engine-tests/src/tests/sanity.rs
```diff
@@ -694,6 +694,27 @@ fn test_eth_transfer_incorrect_nonce() {
...
+                gas_price: 0.into(),
+                gas_limit: u64::MAX.into(),
+                to: None,
+                value: Wei::zero(),
+                data: hex::decode("6080604052348015600e575f80fd5b50607480601a5f395ff3fe6080604052348015600e575f80fd5b50600436106026575f3560e01c8063919840ad14602a575b5f80fd5b600560405190815260200160405180910390f3fea2646970667358221220cb01b9b9c75e5cd079a1980af2fe4397d2029888d12737d74cbbc10e0de65bd364736f6c63430008150033").unwrap(),
+            }
+        })
+        .unwrap();
+
+    println!("{result:?}");
```
birchmd, line 715: Println not needed, could think about adding an assert on the output instead or something.

engine-tests/src/tests/sanity.rs
```diff
@@ -694,6 +694,27 @@ fn test_eth_transfer_incorrect_nonce() {
...
+            aurora_engine_transactions::legacy::TransactionLegacy {
+                nonce,
+                gas_price: 0.into(),
+                gas_limit: u64::MAX.into(),
+                to: None,
+                value: Wei::zero(),
+                data: hex::decode("6080604052348015600e575f80fd5b50607480601a5f395ff3fe6080604052348015600e575f80fd5b50600436106026575f3560e01c8063919840ad14602a575b5f80fd5b600560405190815260200160405180910390f3fea2646970667358221220cb01b9b9c75e5cd079a1980af2fe4397d2029888d12737d74cbbc10e0de65bd364736f6c63430008150033").unwrap(),
+            }
+        })
+        .unwrap();
```
birchmd, line 713:
```suggestion
        .expect("Should be able to execute EVM bytecode including PUSH0");
```

birchmd (approved):

aleksuss (approved):

mrLSD (approved):

PR #807, opened 2023-07-28
feat: add function to return wNEAR address

aleksuss:
## Description

The PR adds a function for returning factory wNEAR address in borsh format.

### DIFF
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -498,7 +498,7 @@ mod contract {
         crate::xcc::set_wnear_address(&mut io, &Address::from_array(address));
     }
 
-    /// Returns the address for the `wNEAR` ERC-20 contract in hex format.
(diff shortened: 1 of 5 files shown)


engine/src/lib.rs
```diff
@@ -498,6 +498,15 @@ mod contract {
         crate::xcc::set_wnear_address(&mut io, &Address::from_array(address));
     }
 
+    /// Returns the address for the `wNEAR` ERC-20 contract in borsh format.
+    #[no_mangle]
+    pub extern "C" fn factory_wnear_address() {
```
birchmd, line 503: I wonder if we should call it `factory_get_wnear_address` to mirror `factory_set_wnear_address`

birchmd (approved):

mrLSD (approved):

PR #809, opened 2023-07-29
`engine-modexp` bug fixes and performance improvements

guidovranken:
## Description

This fixes the following issues with `engine-modexp`:

- addition overflow in big_sq that would lead to incorrect output (and a crash in debug mode)
- assert failure in sub_to_same_size that would lead to incorrect output
- slow modular exponentiation with even modulus. The bottleneck was the routine were the inverse of A mod M was computed, where M is a power of two. This inversion function has been replaced with the algorithm found in [this paper](https://eprint.iacr.org/2017/411.pdf), which is much faster.

The performance of modular exponentiation with even modulus is improved. Performance remains the same for odd modulus.

## Testing

Extensive fuzz testing, was was also the method used to detect the issues that this PR addresses.


joshuajbouw (approved):

joshuajbouw:
I believe the cargo fmt is failing due to the use of nightly cargo fmt check in the CI.

aleksuss:
@joshuajbouw no. We have pretty old nightly in our CI so it should be fixed, I guess.

birchmd (approved):

mrLSD (approved):

PR #814, opened 2023-08-04
`engine-modexp` bug fixes and performance improvements with unusual exponents

guidovranken:
## Description

- Fixes a panic with empty exponent
- Ensure correct result with zero exponent
- Fix performance issues with exponents with large zero padding

The performance is increased for some cases (those with exponents with a large amount of leading zeroes) and remains the same for all other cases.

## Testing

Fuzzing and tests added.

### DIFF
--- a/engine-modexp/src/mpnat.rs
+++ b/engine-modexp/src/mpnat.rs
@@ -22,6 +22,17 @@ pub struct MPNat {
 }
 
 impl MPNat {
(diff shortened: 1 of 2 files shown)


birchmd (approved):
Nice find!

joshuajbouw (approved):
Nice!

mrLSD (approved):

PR #821, opened 2023-08-10
Feat(silo): Test eth-connector with aurora contract

mrLSD:
# Description
Test using eth-connector for Silo via Aurora Engine contract (for v2.10.1).


engine-tests-connector/src/utils.rs
```diff
@@ -336,6 +502,19 @@ pub fn validate_eth_address(address: &str) -> Address {
     Address::decode(address).unwrap()
 }
 
+pub fn get_aurora_contract() -> Vec<u8> {
+    let file = "../bin/aurora-engine-mainnet.wasm";
```
aleksuss, line 506:
```suggestion
    let file = "../bin/aurora-mainnet.wasm";
```

engine-tests-connector/src/silo.rs
```diff
@@ -0,0 +1,31 @@
...
+    assert_eq!(res, 5 * ETH);
+
+    let (recv, _) = create_eth_acc();
+    let tx_args = set_submit_tx(0, recv, ETH);
+    let tx = prepare_submit_transaction(&sk, tx_args).await;
+    let res = contract.submit(tx).await.unwrap();
+    assert!(res.is_success());
+
+    let res = contract.get_eth_balance(&addr).await.unwrap();
+    assert_eq!(res, 3 * ETH);
```
aleksuss, line 28: I wonder why the balance is decreased to `2 * ETH`. As I can see, you execute only one transaction and the fixed gas cost is: `ETH`.

aleksuss (approved):

engine-tests-connector/src/utils.rs
mrLSD: we can't use
`    let file = "../bin/aurora-mainnet.wasm";`

This name conflicts with other builds

aleksuss:
Closed because a similar logic has already been added.

PR #824, opened 2023-08-21
Chore: Update to SputnikVM version 0.38.2

birchmd:
## Description

Updates to the latest version of the Aurora fork of SpunikVM. The previous version had a bug in the implementation of the `returndatacopy` opcode.

For some reason this also causes the EVM gas cost of the promise results precompile to increase. I am not sure why, but it's only a small increase (recall that any EVM transaction must spend at least 21000 gas), so I think's ok.

## Testing

Added an additional test for `returndatacopy`.


birchmd:
Superseded by #826

PR #827, opened 2023-08-23
Chore: factor logic out of all top-level contract methods

birchmd:
## Description

This PR is further preparation for the eventual introduction of the hashchain to Aurora Engine. This PR is only a refactor (no functional change to the code). The purpose of this refactor is to move all the logic related to state-changing contract methods from `lib.rs` into a public module that can be reused in the standalone engine. Making this change will ensure that the standalone engine and the Wasm contract both see the exact same input bytes and generate the exact same output bytes. This is important because the input and output both contribute to the hashchain.

This PR only refactors `lib.rs`. In a future PR I will change the standalone engine to use the new module introduced by this PR, replacing the logic that currently exists in `fn execute_transaction`.

No impact -- refactor only.

## Testing

Existing tests -- refactor only.

The logic that is removed from `lib.rs` should appear nearly verbatim in the new `contract_methods.rs` module. The main difference is being able to replace `sdk_unwrap` with `?` since all functions now return a `Result`.


aleksuss (approved):
Good idea. I guess we need to move other logic to corresponding modules as well. 

P.S. I would rename `contract_methods` to `modifying_methods` or something not too common.

birchmd:
> I would rename contract_methods to modifying_methods or something not too common.

Thanks for the suggestion @aleksuss . I decided to keep the `contract_methods` top-level module name, but split the functions into logical sub-modules. Does that look good to you?

aleksuss:
Yes. That sounds good.

mrLSD (approved):

PR #831, opened 2023-08-30
Feat(engine): Hashchain integration

birchmd:
## Description

This PR integrates the hashchain feature into the Aurora Engine contract (see https://github.com/aurora-is-near/AIPs/pull/8). This change is fully backwards compatible because by default there is no hashchain present in the contract state and therefore no hashchain computation is done.

The method to activate the hashchain is `start_hashchain` requires the contract to be paused and can only be called by a privileged account (I chose to use the key manager instead of introducing a new role). These restrictions are necessary because the hashchain is initialized with a value based on the Engine's transaction history (to enable validating Aurora blocks starting from the deployment of the contract).

After activating the hashchain there will be a small increase in gas usage because of the hashchain computation. It's not very much for most transactions (as seen in the performance regression tests), but the amount of gas is proportional to the size of the input + output of the transaction (because we compute a hash from this data).

## Testing

The tests have been changed to enable the hashchain by default. Therefore all existing tests are running the hashchain code and testing that the Wasm and Standalone results match (this is important because the Refiner will be computing hashchain values off-chain using the Standalone Engine). An additional test has been added for the hashchain feature itself.

## Additional information

Note: the diff in `contract_methods` is relatively large, but the code changes are minimal. All that happened is the existing function bodies were wrapped in `with_hashchain`, which changed the indentation. Viewing the PR with whitespace changes ignored may make this section easier to review.

### DIFF
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -36,6 +36,7 @@ pub mod deposit_event;
 pub mod engine;
 pub mod errors;
 pub mod fungible_token;
+pub mod hashchain;
 pub mod pausables;
 mod prelude;
 pub mod state;
@@ -114,7 +115,8 @@ mod contract {
     #[no_mangle]
     pub extern "C" fn new() {
         let io = Runtime;
-        contract_methods::admin::new(io)
+        let env = Runtime;
+        contract_methods::admin::new(io, &env)
             .map_err(ContractError::msg)
             .sdk_unwrap();
     }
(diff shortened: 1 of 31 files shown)


engine/src/lib.rs
```diff
@@ -525,6 +537,14 @@ mod contract {
         io.return_output(&value.0);
     }
 
+    #[no_mangle]
+    pub extern "C" fn get_latest_hashchain() {
```
aleksuss, line 541: Does having a similar method but with a specified height argument make sense?

aleksuss (approved):
Amazing work 👍🏻

mrLSD (approved):
LGTM!

PR #832, opened 2023-08-30
Feat: update clippy requirements to `nightly-2023-08-24` without changing toolchain

mrLSD:
# Description

Currently impossible to update toolchain, it's related to [NEAR issue](https://github.com/near/nearcore/issues/9143).

And it is impossible to do `cargo update`, because [issue](https://github.com/rust-cli/anstyle/issues/118#issuecomment-1695730543).

- Updated clippy requirements to `nightly-2023-08-24`, without updating `rust-toolchain`.
- Fixed clippy issues, related to clippy requirements for `nightly-2023-08-24` version.

### DIFF
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -105,6 +105,7 @@ mod contract {
     const CODE_KEY: &[u8; 4] = b"CODE";
     const CODE_STAGE_KEY: &[u8; 10] = b"CODE_STAGE";
 
+    #[allow(clippy::empty_line_after_doc_comments)]
     ///
     /// ADMINISTRATIVE METHODS
     ///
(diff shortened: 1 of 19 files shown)


engine/src/lib.rs
```diff
@@ -105,6 +105,7 @@ mod contract {
     const CODE_KEY: &[u8; 4] = b"CODE";
     const CODE_STAGE_KEY: &[u8; 10] = b"CODE_STAGE";
 
+    #[allow(clippy::empty_line_after_doc_comments)]
```
aleksuss: Would it be better to remove this empty line?

mrLSD: Actually, it's not `doc comments` at all. It's comments for codec section.

aleksuss (approved):

PR #834, opened 2023-09-05
fix: fix in the exit_to_eth precompile

aleksuss:
## Description

The PR fixes the exit to ethereum precompile.

There are no changes in performance.

## Testing

There are no tests for testing this part of the code, but it would be nice to have.

### DIFF
--- a/engine-precompiles/src/native.rs
+++ b/engine-precompiles/src/native.rs
@@ -6,7 +6,7 @@ use crate::prelude::{
     sdk::io::{StorageIntermediate, IO},
     storage::{bytes_to_key, KeyPrefix},
     types::{Address, Yocto},
-    vec, BorshSerialize, Cow, ToString, Vec, U256,
+    vec, BorshDeserialize, BorshSerialize, Cow, ToString, Vec, U256,
 };
 #[cfg(feature = "error_refund")]
 use crate::prelude::{
@@ -15,7 +15,10 @@ use crate::prelude::{
 };
 use crate::PrecompileOutput;
 use aurora_engine_types::account_id::AccountId;
+use aurora_engine_types::parameters::connector::WithdrawSerializeType;
+use aurora_engine_types::parameters::WithdrawCallArgs;
 use aurora_engine_types::storage::EthConnectorStorageId;
+use aurora_engine_types::types::NEP141Wei;
 use evm::backend::Log;
 use evm::{Context, ExitError};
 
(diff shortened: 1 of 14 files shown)


engine-precompiles/src/native.rs
```diff
@@ -537,7 +533,7 @@ impl<I: IO> Precompile for ExitToEthereum<I> {
                     .try_into()
                     .map_err(|_| ExitError::Other(Cow::from("ERR_INVALID_RECIPIENT_ADDRESS")))?;
                 (
-                    self.current_account_id.clone(),
+                    get_eth_connector_contract_account(&self.io)?,
                     // There is no way to inject json, given the encoding of both arguments
                     // as decimal and hexadecimal respectively.
                     WithdrawCallArgs {
```
birchmd, line 539: One more change is needed here. As @karim-en pointed out in our call, the Engine Withdraw used Borsh serialization for the arguments and [so does the new eth-connector](https://github.com/aurora-is-near/aurora-eth-connector/blob/master/eth-connector/src/lib.rs#L488), however regular NEP-141 tokens use JSON (see ERC-20 implementation below).

I think there are two options for what to do:

1. Change the eth-connector to use JSON args for withdraw, thus making it consistent with the other bridged NEP-141 tokens, and then make the corresponding change here.
2. Add another config to the Engine to say what format the withdraw arguments are supposed to be in.

I personally think 1 is the better solution.

birchmd (changes requested):

engine-precompiles/src/native.rs
karim-en: I also like the first option, but this change should be backward compatible, because this change can break any services or contracts that interact with the engine contract.
To make it backward compatible I think we can try to deserialise the input twice, as an borsh, and then try to deserialise it as a json if it was failed.

mrLSD: @birchmd what existing problem or business goal is it to solve by introducing changes that break the existing API? I do not have enough context to understand the essence of such a controversial proposal.

mrLSD (approved):

engine-precompiles/src/native.rs
aleksuss: @karim-en, what is the JSON format for this message should be?

birchmd: @karim-en Backwards compatibility is maintained by the main Aurora Engine `withdraw` (which delegates to the new connector's `engine_withdraw`) remaining borsh-encoded. I don't think we need fallback serialization on the new `withdraw` method of the new connector.

@aleksuss the format will be identical to [the ERC-20 case](https://github.com/aurora-is-near/aurora-engine/blob/77c284a5c231f84c1ec4c2080841528170e0413c/engine-precompiles/src/native.rs#L593).

@mrLSD This is not a breaking change to the existing API because the Aurora Engine maintains backwards compatibility. The reason to API of the new contract to be different from Aurora Engine is because then it is consistent with all other bridged NEP-141 tokens (i.e. all ERC-20 tokens bridged from Ethereum). All NEP-141 tokens having a consistent interface allows us to use any bridged token as the base token for a silo, not just ETH. Having alternate base tokens is a business requirement.

aleksuss: @birchmd don't we need to support both versions of deserialization in the main aurora contract in the `withdraw` function? In case we will use the main contract as an eth-connector for using ETH as the base token in silo?

engine-types/src/parameters/connector.rs
```diff
@@ -165,6 +165,13 @@ pub struct StorageUnregisterCallArgs {
 #[derive(Debug, Clone, BorshSerialize, BorshDeserialize, PartialEq, Eq)]
 pub struct SetEthConnectorContractAccountArgs {
     pub account: AccountId,
+    pub withdraw_serialize_type: WithdrawSerializeType,
```
birchmd, line 168: Should maybe be optional with default value `WithdrawSerializeType::Borsh` for backwards compatibility

birchmd (approved):
LGTM

engine-types/src/parameters/connector.rs
aleksuss: I thought that we don't have any production contract with such functionality and therefore I skipped it. But why not?

aleksuss: By the way. I'm not sure that borsh allows us to use default value if there are no bytes for field like serde does. I mean that we can't get: 
```rust
SetEthConnectorContractAccountArgs {
    account: "eth-connector.near".parse().unwrap(),
    withdraw_serialize_type: WithdrawSerializeType::Borsh,
}
```
from
```
 "eth-connector.near".parse::<AccountId>().unwrap().try_to_vec().unrwap()
```

If only to write custom deserialization for SetEthConnectorContractAccountArgs. But does it make sense? As I said we don't use `set_eth_connector_contract_account` api in production.

birchmd: Yeah, I agree it doesn't make sense to put in so much work when there are no production silos that use it yet.

birchmd (approved):

Issue #818, opened 2023-08-10
wasm-opt for improving gas consumption


Tarnadas:
Hey,

I looked through your build steps and haven't found usage of [wasm-opt](https://github.com/WebAssembly/binaryen), which honestly surprised me. I basically use it for every contract I develop for a free bonus in runtime perf and code size.

I am currently developing a contract that uses Aurora SDK to do interactions from Aurora to the contract on Near. I have a couple of integration tests and I wanted to check what wasm-opt would change regarding gas consumption. Most of the gas is not used from Aurora contract, but from the cross contract calls and contracts on native Near, but you can clearly see a small improvement. I precompiled the Aurora contract for the integration tests like you're doing it in the `Makefile`, but for the optimized version I invoked `wasm-opt -O4 res/aurora_engine.wasm -o res/aurora_engine_opt.wasm --strip-debug --vacuum`.

Here are the relevant logs from my result:
Unoptimized:
```
    approve wNear for Solidity contract gas burnt: 7.526 TGas
    storageDeposit gas burnt: 109.358 TGas
    approve ETH for Solidity contract gas burnt: 7.757 TGas
    approve wNear for Solidity contract gas burnt: 7.522 TGas
    storageDeposit gas burnt: 73.606 TGas
    approve ETH for Solidity contract gas burnt: 7.633 TGas
    deposit gas burnt: 100.992 TGas
    createLimitOrder gas burnt: 47.737 TGas
    swap gas burnt: 126.609 TGas
    withdrawAll gas burnt: 44.345 TGas
    withdrawAll gas burnt: 63.940 TGas
    withdrawAll gas burnt: 44.189 TGas
    withdrawAll gas burnt: 44.197 TGas
```

Optimized:
```
    approve wNear for Solidity contract gas burnt: 7.503 TGas
    storageDeposit gas burnt: 108.474 TGas
    approve ETH for Solidity contract gas burnt: 7.733 TGas
    approve wNear for Solidity contract gas burnt: 7.480 TGas
    approve ETH for Solidity contract gas burnt: 7.591 TGas
    deposit gas burnt: 99.882 TGas
    createLimitOrder gas burnt: 46.957 TGas
    swap gas burnt: 125.282 TGas
    withdrawAll gas burnt: 43.621 TGas
    withdrawAll gas burnt: 63.172 TGas
    withdrawAll gas burnt: 43.464 TGas
    withdrawAll gas burnt: 43.472 TGas
```

[...]

PR #835, opened 2023-09-07
Feat: added wasm optimization

mrLSD:
## Description

- Added wasm optimization with `wasm-opt`.
- Refactored test gas cost.

## Gas cost

Gas costs decreased by 3-5%.

## Binary size

Binary size decreased to 9%.

## Tests

Refactored test gas cost.


mrLSD:
Related to #818

aleksuss (changes requested):

aleksuss:
Also, we need to modify building artifacts [script](https://github.com/aurora-is-near/aurora-engine/blob/develop/.github/workflows/builds.yml).

birchmd (commented):
Cool! Nice to see low-hanging fruit like this giving a performance improvement.

We will probably also need to update the README or something to say that wasm-opt is now a required tool for building the project.

birchmd (approved):

aleksuss (approved):

aleksuss:
@mrLSD, why don't we use `wasm-opt` for `xcc-router` smart contract?

mrLSD:
> @mrLSD, why don't we use `wasm-opt` for `xcc-router` smart contract?

Because it's not part of the test-flow:
```
[tasks.test-flow]
category = "Test"
dependencies = [
    "build-test",
    "test-contracts",
    "test-workspace",
    "bench-modexp",
]
```

aleksuss:
I also want to note that applying the `wasm-opt` utility for the contract file resolves the deserialisation error if the contract is built with Rust > 1.81.0. That is precisely how the `cargo-near` crate resolves the issue.
```
 --no-wasmopt
          Do not run equivalent of `wasm-opt -O` on the generated output as a post-step
          
          This `wasm-opt -O` step is required for correctness of build on rustc >= 1.82.
          Otherwise, if the post-step is skipped, nearcore runtime won't be able to parse and validate deployed wasm.
          In these cases a `CompilationError(PrepareError(Deserialization))` is expected to be emitted
          when performing calls to wasm with unsupported features.
```

PR #837, opened 2023-09-12
feat: possibility to set metadata for erc20 contracts

aleksuss:
## Description

The PR adds functions for setting (`set_erc20_metadata`) and getting (`get_erc20_metadata`) metadata of ERC-20 contracts deployed with `deploy_erc20_token` transaction.

There are no performance changes.

## Testing

## Additional information

Probably, there should be a new role for `set_erc20_metadata` transaction. I guess it should be added in a new PR.


engine/src/contract_methods/connector.rs
```diff
@@ -404,3 +404,55 @@ pub fn set_paused_flags<I: IO + Copy, E: Env>(io: I, env: &E) -> Result<(), Cont
...
+        predecessor_address(&env.predecessor_account_id()),
+        current_account_id,
+        io,
+        env,
+    );
+    let metadata = engine.get_erc20_metadata(erc20_address)?;
+
+    io.return_output(
+        metadata
+            .try_to_vec()
```
birchmd, line 453: I wonder if this function would be easier to use if it returned the result as JSON instead of Borsh. Many Near tools work better with JSON-encoded output

engine-types/src/parameters/connector.rs
```diff
@@ -220,6 +220,36 @@ impl rlp::Encodable for LogEntry {
...
+    pub decimals: u8,
+}
+
+impl Default for Erc20Metadata {
+    fn default() -> Self {
+        Self {
+            name: "Token".to_string(),
+            symbol: "TKN".to_string(),
+            decimals: 18, // To be compatible with MetaMask.
+        }
```
birchmd, line 249: The original default values were `"Empty", "EMPTY", 0`. Should we leave them as they were? cc @karim-en

birchmd (commented):
Nice work! Just a few comments.

engine-types/src/parameters/connector.rs
aleksuss: At least these changes don't affect our tests at all 😄. Also, I thought about adding a possibility to set optional metadata while deploying and mirroring a new ERC-20 contract.

engine/src/contract_methods/connector.rs
karim-en: Yeh, json for the input/output is better definitely and much easier to call with the cli

engine-types/src/parameters/connector.rs
karim-en: I think we should keep the `"Empty", "EMPTY", 0` as a default metadata.

aleksuss:
One bad moment of changing borsh to JSON is that it increases the number of functions and size of the smart contract.
cc @birchmd

birchmd:
Increasing the number of functions is ok in my opinion. The reason for counting them is because there is a [hard limit on the number of functions Near allows in a contract](https://github.com/near/nearcore/pull/4954). But this limit is 10000, so we are not close to exceeding it. 

How much bigger is the contract size? Will it be offset by the wasm opt improvment?

aleksuss:
> How much bigger is the contract size? Will it be offset by the wasm opt improvement?

`1166962` vs `1060706` (develop), diff: 106256 bytes. After applying `wasm-opt` the size becomes `1041185` bytes.

birchmd (approved):
LGTM!

I think the increase in contract size is worth the utility these methods provide.

mrLSD (approved):

engine/src/contract_methods/connector.rs
```diff
@@ -404,3 +404,51 @@ pub fn set_paused_flags<I: IO + Copy, E: Env>(io: I, env: &E) -> Result<(), Cont
...
+#[named]
+pub fn set_erc20_metadata<I: IO + Copy, E: Env, H: PromiseHandler>(
+    io: I,
+    env: &E,
+    handler: &mut H,
+) -> Result<SubmitResult, ContractError> {
+    with_hashchain(io, env, function_name!(), |io| {
+        let state = state::get_state(&io)?;
+        require_running(&state)?;
+        // TODO: Define special role for this transaction. Potentially via multisig?
```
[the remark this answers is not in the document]
karim-en: I proposed to have another role for this and grant this role to some internal DAO with participants from engine/bridge teams as a long term solution, and use the account `metadata.factory.bridge.near` as a short term.
But I'm not sure that this should be done in the scope of this PR, because looks like the roles system is not implemented in the engine, so this task can be more compacted than I expect.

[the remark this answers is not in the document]
karim-en: Also a while ago I proposed to retrieve the metadata from the nep141 using the cross contract call, but there were objections on this from Kirill to this because in some potential scenarios we can set a metadata that differ from the nep141.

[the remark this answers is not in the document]
sept-en: I think the right set of scenarios and permissions should be the following:
* Special role is able to obtain the metadata from the NEP-141 contract and set it.
* DAO can set any metadata if needed (for the rare special cases).

PR #845, opened 2023-10-04
feat: mirroring erc-20 contracts

aleksuss:
## Description

The PR adds a possibility to mirror previously deployed ERC-20 contracts on the main Aurora contract. The transaction `mirror_erc20_token` is allowed in the Silo mode only.

There are no changes in the performance/gas consumption.

## Testing


engine-types/src/parameters/connector.rs
```diff
@@ -252,6 +252,17 @@ impl rlp::Encodable for LogEntry {
...
 
+/// Borsh-encoded parameters for `mirror_erc20_token` function.
+#[derive(BorshSerialize, BorshDeserialize, Debug, Eq, PartialEq, Clone)]
+pub struct MirrorErc20TokenArgs {
+    /// AccountId of the main Aurora contract which has previously deployed ERC-20.
+    pub contract_id: AccountId,
+    /// AccountId of the bridged NEP-141 token.
+    pub nep141: AccountId,
+    /// Metadata of the ERC-20 token.
+    pub erc20_metadata: Option<Erc20Metadata>,
```
birchmd, line 263: Should this also be looked up from the other Engine instance instead of passed as an argument?

karim-en: Yes, we also need a method to mirror the metadata.

engine/src/contract_methods/connector/mod.rs
```diff
@@ -382,14 +384,15 @@ pub fn set_erc20_metadata<I: IO + Copy, E: Env, H: PromiseHandler>(
...
-        let result = engine.set_erc20_metadata(args.erc20_address, args.erc20_metadata, handler)?;
+        let result = engine.set_erc20_metadata(&args.erc20_identifier, args.metadata, handler)?;
 
         Ok(result)
     })
 }
 
 pub fn get_erc20_metadata<I: IO + Copy, E: Env>(mut io: I, env: &E) -> Result<(), ContractError> {
-    let erc20_address = io.read_input_arr20().map(Address::from_array)?;
+    let erc20_identifier =
```
aleksuss, line 394: Technically it is a breaking change, but I think it's okay because we haven't deployed this function yet.

karim-en:
@aleksuss Could you please add the details about the added public methods/api with its arguments (json/borsh) in the PR description?

birchmd (approved):
Nice work!

mrLSD (approved):

PR #847, opened 2023-10-11
feat: add transaction for adding full access key

aleksuss:
## Description

The PR adds a transaction which allows adding the full access key into the account of the smart contract. 

No changes.

## Testing


engine/src/contract_methods/admin.rs
```diff
@@ -426,6 +427,35 @@ pub fn get_latest_hashchain<I: IO>(io: &mut I) -> Result<(), ContractError> {
...
+    let current_account_id = env.current_account_id();
+    let action = PromiseAction::AddFullAccessKey {
+        public_key,
+        nonce: 0, // not actually used - depends on block height
+    };
+    let promise = PromiseBatchAction {
+        target_account_id: current_account_id,
+        actions: vec![action],
+    };
+    let promise_id = unsafe { handler.promise_create_batch(&promise) };
```
birchmd, line 452:
```suggestion
    // SAFETY: This action is dangerous because it adds a new full access key (FAK) to the Engine account.
    // However, it is safe to do so here because of the `require_owner_only` check above; only the
    // (trusted) owner account can add a new FAK.
    let promise_id = unsafe { handler.promise_create_batch(&promise) };
```

birchmd (approved):

vzctl (approved):

mrLSD (approved):

PR #849, opened 2023-10-12
ci: use github heave runners for tests

aleksuss:
## Description

Attempt to improve CI tests flow by using GitHub hosted heavy runners.


birchmd (commented):
I think we used to have mainnet and testnet in a matrix a while back, but we switched it to sequential because they were somehow interfering with each other on the self-hosted runners. If it works now though then don't worry about this comment.

vzctl (approved):

PR #854, opened 2023-10-23
feat: change the logic of gas cost calculation in silo mode

aleksuss:
## Description

The PR changes the logic of gas calculation in Silo mode. Fixed gas cost has been replaced with a fixed amount of gas per transaction.

There are no changes related to performance.

## Testing

Corresponding tests have been added.


engine-types/src/types/wei.rs
```diff
@@ -188,6 +192,14 @@ impl Sub for Wei {
     }
 }
 
+impl Mul for Wei {
```
birchmd, line 195: I do not like this kind of implementation for newtypes because the units do not work properly. For example, if I have a new type representing meters (the unit for length) and there is this kind of implementation, then we would be able to write code where meters times meters gives you back meters (a measure of length) instead of meters^2 (a measure of area). Clearly this is incorrect.

It is ok to have implementations for `Add` and `Sub` because it does make sense to add together meters and get back meters. It is also ok to have multiplication by a pure scalar number (e.g. `u64`) because it does make sense to multiply a unit by a pure number.

I suggest if multiplication is needed then we have `impl Mul<U256> for Wei` instead.

engine/src/engine.rs
```diff
@@ -1078,16 +1065,19 @@ pub fn submit_with_alt_modexp<
...
-            Err(_e) => {
-                return Err(EngineErrorKind::GasOverflow.into());
-            }
-            Ok(intrinsic_gas) => {
-                if transaction.gas_limit < intrinsic_gas.into() {
-                    return Err(EngineErrorKind::IntrinsicGasNotMet.into());
-                }
+    // Check that fixed gas is less than gasLimit from the transaction.
+    if fixed_gas.map_or(false, |gas| gas.raw() > transaction.gas_limit) {
+        return Err(EngineErrorKind::FixedGasOverflow.into());
```
birchmd, line 1070: The comment and the implementation don't match. The comment says the fixed gas should be less than the tx gas limit, but the implementation checks that the fixed gas is greater.

birchmd (changes requested):

engine/src/engine.rs
```diff
@@ -456,31 +458,26 @@ impl<'env, I: IO + Copy, E: Env, M: ModExpAlgorithm> Engine<'env, I, E, M> {
         sender: &Address,
         transaction: &NormalizedEthTransaction,
         max_gas_price: Option<U256>,
-        fixed_gas_cost: Option<Wei>,
+        fixed_gas: Option<Wei>,
```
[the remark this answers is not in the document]
aleksuss: > It's not an amount of the base currency any more, it is an amount of gas that is still multiplied by the gas price to get the amount of base currency that the user must pay.

Yes. That's correct.

engine-types/src/types/wei.rs
aleksuss: I got your point and agree with it. Good catch 👍🏻

engine-tests/src/tests/silo.rs
```diff
@@ -20,10 +21,13 @@ const INITIAL_BALANCE: Wei = Wei::new_u64(10u64.pow(18) * 10);
 const ZERO_BALANCE: Wei = Wei::zero();
 const INITIAL_NONCE: u64 = 0;
 const TRANSFER_AMOUNT: Wei = Wei::new_u64(10u64.pow(18) * 4);
-const FEE: Wei = Wei::new_u64(10u64.pow(18));
+const FIXED_GAS: EthGas = EthGas::new(10u64.pow(18));
+const ONE_GAS_PRICE: Wei = Wei::new_u64(1);
```
birchmd, line 25: If we were being very strict about units, a gas price would have units of `Wei / EthGas`, not simply `Wei`. But it is common to talk about Ethereum gas prices in units of `Wei` (ignoring the "per gas" part), so maybe this is ok.

birchmd (approved):

mrLSD (approved):

PR #863, opened 2023-11-07
Fix(standalone): properly compute runtime random value

birchmd:
## Description

The Near runtime has a random value available to smart contracts via a cost function. Aurora exposes that random value to EVM smart contracts in two ways: (1) via a custom precompile, and (2) via [EVM opcode 0x44](https://www.evm.codes/#44?fork=shanghai) since it was changed from `DIFFICULTY` to `PREVRANDAO` after the merge. Therefore it is important for the standalone engine to be able to correctly provide the same random value as would be present on-chain.

In this PR I fix the standalone engine to be able to correctly reproduce the Near runtime random value based on the implementation present in nearcore. There will also be a follow-up PR to `borealis-engine-lib` to make use of this change in the Borealis Engine there.

The random value is computed as `sha256(block_random_value || action_hash)` where the `block_random_value` comes from the protocol-level VRF (this was the value which previously we were using directly) and the `action_hash` is derived from the (`FunctionCall`) `Action` that is being executed in the Near runtime. To have the `action_hash` available to the standalone engine I am adding a new field to `TransactionMessage` which stores this value.

In the tests the `action_hash` field is populated in a reasonable way, but not exactly as it would be in nearcore because there is no actual `Action` (we skip straight to the Wasm execution). However, in the follow-up PR in `borealis-engine-lib` the field will be populated from the Near data. Once this change is made, it will fix a bug in Borealis Engine where it was not correctly reproducing the execution of contracts involving randomness.

N/A : change is to standalone engine only.

## Testing

The test for the random precompile has been updated to reflect this change. A test related to tracing is also changed to no longer depend on the randomness precompile because the test did not rely on the specifics of which precompile was used and now using randomness in the tests has extra setup steps.

### DIFF
--- a/engine-standalone-storage/src/relayer_db/mod.rs
+++ b/engine-standalone-storage/src/relayer_db/mod.rs
@@ -155,6 +155,7 @@ where
             transaction: crate::sync::types::TransactionKind::Submit(tx),
             promise_data: Vec::new(),
             raw_input: transaction_bytes,
+            action_hash: H256::default(),
         };
         storage.set_transaction_included(tx_hash, &tx_msg, &diff)?;
     }
@@ -268,6 +269,7 @@ mod test {
                         transaction: TransactionKind::Unknown,
                         promise_data: Vec::new(),
                         raw_input: Vec::new(),
+                        action_hash: H256::default(),
                     },
                     &diff,
                 )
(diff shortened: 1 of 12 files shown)


aleksuss (approved):
Good work 👍🏻

mrLSD (approved):

mrLSD:
LGTM!

PR #865, opened 2023-11-14
Fix panic in ExitToNear

guidovranken:
## Description

If `input` is empty a panic occurs. This PR sets `flag` to the default value in that case.

## Testing

Fuzzing.

Ensure that `flag` assuming the default value if `input` is empty is the intended behavior.

## Additional information

### DIFF
--- a/engine-precompiles/src/native.rs
+++ b/engine-precompiles/src/native.rs
@@ -349,7 +349,7 @@ impl<I: IO> Precompile for ExitToNear<I> {
         // First byte of the input is a flag, selecting the behavior to be triggered:
         //      0x0 -> Eth transfer
         //      0x1 -> Erc20 transfer
-        let flag = input[0];
+        let flag = input.first().copied().unwrap_or_default();
         #[cfg(feature = "error_refund")]
         let (refund_address, mut input) = parse_input(input)?;
         #[cfg(not(feature = "error_refund"))]


birchmd (approved):
Good catch, thanks!

If the input is empty it will fail the `validate_input_size` check in `parse_input` immediately after the assignment of `flag`, so the default value will never be used, which I think is a good thing. Instead of assigning a default value, we could potentially move flag assignment input `parse_input` after the size is already validated. But maybe that's more complexity than necessary.

aleksuss:
Yes. But the only problem is that the `parse_input` returns modified input. I would propose modifying the `parse_input` function so it will return the flag as well:
```rust
#[cfg(feature = "error_refund")]
let (flag, refund_address, mut input) = parse_input(input)?;
#[cfg(not(feature = "error_refund"))]
let (flag, mut input) = parse_input(input)?;
```

birchmd:
@aleksuss  Yes, that is what I suggested above. My worry about complexity is because of the conditional compilation you would need to change both versions of the `parse_input` function (duplicating the flag assignment logic). It's not a big deal to duplicate that one line, and I agree it's nicer to encapsulate all the input handling into one function (even if there are two different versions of that function). So I think I agree with you that it's the better solution overall.

aleksuss:
On the other hand, even if we get the default value for the flag, we will fail later because we don't pass the validation. So, I guess it's absolutely fine and could be merged.

aleksuss (approved):

PR #867, opened 2023-11-16
Feat(xcc): ses native bridge logic to unwrap wNEAR

birchmd:
## Description

Since #750 has been merged, the ERC-20 connector is able to unwrap wrapped Near tokens into Near native tokens. This logic was duplicated by the XCC router, but in this PR we remove the logic from XCC and use the bridge's unwrapping instead. This has a nice side-effect of saving us 20 Tgas from the XCC overhead.

No impact on regular transactions, 20 Tgas improvement to XCC transactions.

## Testing

Existing tests. Specifically this relies heavily on the upgrade test introduced in #866

### DIFF
--- a/etc/xcc-router/src/lib.rs
+++ b/etc/xcc-router/src/lib.rs
@@ -4,7 +4,7 @@ use aurora_engine_types::parameters::{
 };
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::collections::{LazyOption, LookupMap};
-use near_sdk::json_types::{U128, U64};
+use near_sdk::json_types::U64;
 use near_sdk::BorshStorageKey;
 use near_sdk::{
     env, near_bindgen, AccountId, Gas, PanicOnDefault, Promise, PromiseIndex, PromiseResult,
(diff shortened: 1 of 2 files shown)


etc/xcc-router/src/lib.rs
```diff
@@ -209,19 +172,14 @@ impl Router {
...
-            .unwrap_or_else(|| env::panic_str("ERR_CONTRACT_NOT_INITIALIZED"));
-
+        let parent = self.require_parent_caller();
         Promise::new(parent).transfer(REFUND_AMOUNT)
     }
 }
 
 impl Router {
-    fn require_parent_caller(&self) {
+    fn require_parent_caller(&self) -> AccountId {
```
aleksuss, line 182: I guess it's out of the scope of this PR, but it looks like a bad design of the method. 1. The name doesn't correspond to what it really does. 2. And looks like more than one responsibility.

aleksuss (approved):

etc/xcc-router/src/lib.rs
birchmd: Good point. I have refactored it in https://github.com/aurora-is-near/aurora-engine/pull/867/commits/c847479fff268be0c45d881ecd3c3ce993411d8a

karim-en (commented):
LGTM!

PR #869, opened 2023-11-21
Feat(standalone): SubmitResult from bridge transactions is available

birchmd:
## Description

Explorers like Blockscout rely on EVM logs to index certain events such as ERC-20 token transfers/minting/burning. When NEP-141 tokens are bridged to Aurora via `ft_on_transfer` there are ERC-20 token events that should be produced and indexed by explorers. This PR changes how `ft_on_transfer` calls are handled to expose the EVM logs generated to the standalone engine, which in turn will make them available to the Refiner and thus also to block explorers.

There will be a follow-up PR in `borealis-engine-lib` which uses the `SubmitResult` to include logs in the transactions corresponding to `ft_on_transfer`.

N/A no significant changes to contract code

## Testing

Existing tests. (A new test will be added to `borealis-engine-lib` when we do integration there)

### DIFF
--- a/engine/src/engine.rs
+++ b/engine/src/engine.rs
@@ -30,7 +30,7 @@ use crate::prelude::precompiles::Precompiles;
 use crate::prelude::transactions::{EthTransactionKind, NormalizedEthTransaction};
 use crate::prelude::{
     address_to_key, bytes_to_key, sdk, storage_to_key, u256_to_arr, vec, AccountId, Address,
-    BTreeMap, BorshDeserialize, KeyPrefix, PromiseArgs, PromiseCreateArgs, Vec, Wei, Yocto,
+    BTreeMap, BorshDeserialize, Cow, KeyPrefix, PromiseArgs, PromiseCreateArgs, Vec, Wei, Yocto,
     ERC20_DIGITS_SELECTOR, ERC20_MINT_SELECTOR, ERC20_NAME_SELECTOR, ERC20_SET_METADATA_SELECTOR,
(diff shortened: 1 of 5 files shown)


mrLSD (approved):

engine/src/engine.rs
```diff
@@ -751,24 +728,45 @@ impl<'env, I: IO + Copy, E: Env, M: ModExpAlgorithm> Engine<'env, I, E, M> {
...
+                    kind: EngineErrorKind::EvmError(ExitError::Other(Cow::Borrowed(
+                        INVALID_MESSAGE,
+                    ))),
+                    gas_used: 0,
+                });
+            }
+            let address_bytes: [u8; 20] = hex::decode(&message[..40])
+                .ok()
+                .and_then(|bytes| bytes.as_slice().try_into().ok())
+                .ok_or_else(engine_err(INVALID_MESSAGE))?;
```
aleksuss, line 767: I guess `hex::decode_to_slice` will be more effective here:

```rust
let mut address_bytes = [0; 20];
hex::decode_to_slice(&message[..40], &mut address_bytes)
    .map_err(|_| engine_err(INVALID_MESSAGE))?;
```

aleksuss (approved):
Overall looks good

engine/src/engine.rs
birchmd: Good call! Refactored in https://github.com/aurora-is-near/aurora-engine/pull/869/commits/b21eb801c785eeb5ef411b283efc7f42bd8695cf

karim-en (commented):
LGTM!

PR #882, opened 2023-12-07
Remove eth-connector fee logic

karim-en:
## Description
Remove unused and redundant `Fee` logic of the eth-connector.
The primary purpose of this change is to unify the message format for transferring ERC20 and native base tokens.
These changes should simplify the interactions with silo contracts, especially when the base token is an ERC20 token.

Slight reduction in gas

## Testing
Added unit and workspace tests

Verify the `ft_on_transfer` logic for both internal and external connectors.
This change should be backward compatible, not breaking any internal or external components/services.

## Additional information
N/A

### DIFF
--- a/engine/src/contract_methods/connector/deposit_event.rs
+++ b/engine/src/contract_methods/connector/deposit_event.rs
@@ -122,8 +122,6 @@ impl FtTransferMessageData {
 
     /// Prepare message for `ft_transfer_call` -> `ft_on_transfer`
     pub fn prepare_message_for_on_transfer(
-        relayer_account_id: &AccountId,
-        fee: Fee,
         recipient: String,
     ) -> Result<Self, errors::ParseEventMessageError> {
         let address = if recipient.len() == 42 {
@@ -143,10 +141,7 @@ impl FtTransferMessageData {
         #[allow(deprecated)]
         Ok(Self {
             recipient: recipient_address,
-            fee: Some(FtTransferFee {
-                relayer: relayer_account_id.clone(),
-                amount: fee,
-            }),
+            fee: None,
         })
     }
 }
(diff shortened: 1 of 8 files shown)


engine/src/contract_methods/connector/deposit_event.rs
```diff
@@ -18,9 +18,16 @@ pub type EventParams = Vec<EventParam>;
 #[derive(BorshSerialize, BorshDeserialize)]
 #[cfg_attr(not(target_arch = "wasm32"), derive(Debug, PartialEq, Eq))]
 pub struct FtTransferMessageData {
-    pub relayer: AccountId,
     pub recipient: Address,
-    pub fee: Fee,
+    #[deprecated]
+    pub fee: Option<FtTransferFee>,
```
aleksuss, line 23: Why do we still need this field? Saving backward compatibility?

karim-en: Yes, just to save the backward compatibility and make less aggressive changes.
Removing this field and keeping the backward compatibility parse logic is possible, but more cleanup and test adjustments will be needed. Let me know if you think it is better to go that way.

engine/src/contract_methods/connector/deposit_event.rs
```diff
@@ -103,10 +140,13 @@ impl FtTransferMessageData {
         let recipient_address = Address::decode(&address)
             .map_err(errors::ParseEventMessageError::EthAddressValidationError)?;
 
+        #[allow(deprecated)]
         Ok(Self {
-            relayer: relayer_account_id.clone(),
             recipient: recipient_address,
-            fee,
+            fee: Some(FtTransferFee {
```
birchmd, line 146: Should this function be refactored to use `fee: None`? It is used to pass on the fee from the deposit event logs, but maybe we should be ignoring the fee in those logs as well.

birchmd (approved):

mrLSD (approved):

aleksuss (approved):

PR #885, opened 2023-12-15
сhore: bump rust nightly to 2023-12-15

RomanHodulak:
## Description

* Upgrades rust version 1.70 => 1.76
* Resolves all issues reported by clippy

Does not change.

## Testing

The clippy run succeeded and reported no further issues.

Check that the code changes are introduced in a proper code style.

## Additional information

None


Cargo.toml
```diff
@@ -44,14 +44,13 @@ ibig = { version = "0.3", default-features = false, features = ["num-traits"] }
 impl-serde = { version = "0.4", default-features = false}
 lazy_static = "1"
 libsecp256k1 = { version = "0.7", default-features = false }
-near-crypto = "0.17"
+near-crypto = { git = "https://github.com/aleksuss/nearcore", rev = "6832608" }
```
aleksuss, line 47: Here the fork is used because there is an [issue](https://github.com/near/nearcore/issues/10374) in the nearcore.

engine-precompiles/src/prelude.rs
```diff
@@ -1,6 +1,5 @@
 pub use aurora_engine_sdk as sdk;
-pub use aurora_engine_types::account_id::*;
-pub use aurora_engine_types::borsh::{BorshDeserialize, BorshSerialize};
+pub use aurora_engine_types::borsh::BorshSerialize;
```
aleksuss, line 2: Removing the public imports should be safe here because the module `prelude` is private.
cc @RomanHodulak

engine-tests/src/tests/xcc.rs
```diff
@@ -257,28 +256,6 @@ fn test_xcc_exec_gas() {
             outcome.burnt_gas,
             router_exec_cost
         );
-
-        assert_eq!(
```
aleksuss, line 261: This section has been removed since the `VMOutcome` doesn't contain `action_receipts` anymore.

mrLSD (approved):

engine-tests/src/tests/xcc.rs
aleksuss: Yes. There is a [PR](https://github.com/near/nearcore/pull/9332) about that. Try to find out.

aleksuss: I guess you can collect receipts by using this [method](https://github.com/near/nearcore/blob/master/runtime/near-vm-runner/src/logic/dependencies.rs#L321) of this [trait](https://github.com/near/nearcore/blob/master/runtime/near-vm-runner/src/logic/dependencies.rs#L121). We already have a [type](https://github.com/aurora-is-near/aurora-engine/blob/develop/engine-tests/src/utils/mocked_external.rs#L24) which implements this trait.

engine-tests/src/utils/mocked_external.rs
```diff
@@ -76,27 +79,128 @@ impl near_vm_logic::External for MockedExternalWithTrie {
...
 
-    fn get_trie_nodes_count(&self) -> near_primitives::types::TrieNodesCount {
-        let db_reads = self.new_trie_node_count.get();
-        let mem_reads = self.cached_trie_node_count.get();
-        near_primitives::types::TrieNodesCount {
-            db_reads,
-            mem_reads,
-        }
+    fn get_trie_nodes_count(&self) -> TrieNodesCount {
+        self.underlying.get_trie_nodes_count()
```
birchmd, line 87: This is incorrect and is partially contributing to the lowered gas measurements. The default mock provided by nearcore [always returns 0 for touched trie nodes](https://github.com/near/nearcore/blob/master/runtime/near-vm-runner/src/logic/mocks/mock_external.rs#L142). In production Aurora has a large state so the number of trie nodes touched by state access is certainly not zero. This is the reason for the `new_trie_node_count` and `cached_trie_node_count` present in the original code.

engine-tests/src/tests/one_inch.rs
```diff
@@ -104,7 +104,7 @@ fn test_1_inch_limit_order_deploy() {
     // at least 45% of which is from wasm execution
     let wasm_fraction = 100 * profile.wasm_gas() / profile.all_gas();
     assert!(
-        (45..=55).contains(&wasm_fraction),
+        (55..=60).contains(&wasm_fraction),
```
birchmd, line 107: This increase in the wasm fraction was a red-flag to me. If the new version of the rust compiler was making the Wasm code more efficient then the fraction of gas spent on Wasm computation should go down, not up. The reason it goes up is because of now missing the touching trie node costs (see comment on `MockedExternalWithTrie::get_trie_nodes_count`). Please change this back to `45..=55` and revisit after resolving that comment.

engine-tests/src/tests/uniswap.rs
```diff
@@ -49,19 +49,19 @@ fn test_uniswap_exact_output() {
 
     let (_result, profile) =
         context.add_equal_liquidity(LIQUIDITY_AMOUNT.into(), &token_a, &token_b);
-    utils::assert_gas_bound(profile.all_gas(), 32);
+    utils::assert_gas_bound(profile.all_gas(), 23);
     let wasm_fraction = 100 * profile.wasm_gas() / profile.all_gas();
     assert!(
-        (40..=50).contains(&wasm_fraction),
+        (50..=60).contains(&wasm_fraction),
```
birchmd, line 55: Revisit after resolving the comment on `MockedExternalWithTrie::get_trie_nodes_count`

engine-tests/src/tests/uniswap.rs
```diff
@@ -49,19 +49,19 @@ fn test_uniswap_exact_output() {
...
     );
 
     let (_amount_in, profile) =
         context.exact_output_single(&token_a, &token_b, OUTPUT_AMOUNT.into());
-    utils::assert_gas_bound(profile.all_gas(), 17);
+    utils::assert_gas_bound(profile.all_gas(), 13);
     let wasm_fraction = 100 * profile.wasm_gas() / profile.all_gas();
     assert!(
-        (45..=55).contains(&wasm_fraction),
+        (50..=60).contains(&wasm_fraction),
```
birchmd, line 64: Revisit after resolving the comment on `MockedExternalWithTrie::get_trie_nodes_count`

engine-tests/src/tests/xcc.rs
birchmd: Looks like the `action_log` should be the right place to find this. I do not immediately see why the promises are not recorded. I think we should continue investigating this though.

birchmd (changes requested):
The change to `MockedExternalWithTrie::get_trie_nodes_count` needs to be reverted because otherwise we are missing out on a large part of the gas cost.

aleksuss (approved):

engine-tests/src/tests/standard_precompiles.rs
```diff
@@ -60,25 +60,25 @@ fn profile_modexp() {
...
     let profile = precompile_execution_profile("test_ecmul");
-    utils::assert_gas_bound(profile.all_gas(), 7);
+    utils::assert_gas_bound(profile.all_gas(), 6);
 }
 
 #[test]
 fn profile_ecpair() {
     let profile = precompile_execution_profile("test_ecpair");
-    utils::assert_gas_bound(profile.all_gas(), 102);
+    utils::assert_gas_bound(profile.all_gas(), 115);
```
birchmd, line 75: A 10% increase in the gas cost here is a little concerning. Is the new rust compiler worse at optimizing things in some cases now? Or is it something to do with the Near Runtime that is different now?

birchmd (approved):
Looks good me! Thanks for all the work upgrading the Near Runtime dependencies.

I'm a little worried that some gas costs went up with the new compiler version when I think all other rust version changes have lowered the gas costs. But I suppose there is nothing we can do about it.

engine-tests/src/tests/standard_precompiles.rs
aleksuss: Good question. I'm not sure that we can compile the code with the previous Rust compiler since there are a lot of updated crates which don't support it.

PR #888, opened 2024-01-09
Prevent subtraction underflow in CrossContractCall

guidovranken:
## Description

`call.promise_count()` can be 0, in which case `call.promise_count() - 1` causes a subtraction underflow:

https://github.com/aurora-is-near/aurora-engine/blob/1213f2c7c035aa523601fced8f75bef61b4728ab/engine-precompiles/src/xcc.rs#L141

This PR makes the function return to caller with an error if (and only if) this is the case.

## Testing

Fuzzing.


birchmd (approved):

aleksuss (approved):

aleksuss:
Probably this PR should be merged after #885.

PR #889, opened 2024-01-09
Fix balance and gas overflows in CrossContractCall

guidovranken:
## Description

Specific bytecode inputs to the EVM could lead to overflowing additions in gas and balance variables. This PR makes turns those additions into `saturating_add` so that balance/gas requirements will always exceed the capacity of the caller to incur those costs, and the call will fail.

## Testing

Fuzzing.


engine-types/src/parameters/promise.rs
```diff
@@ -134,7 +134,9 @@ impl NearPromise {
     pub fn total_near(&self) -> Yocto {
         match self {
             Self::Simple(x) => x.total_near(),
-            Self::Then { base, callback } => base.total_near() + callback.total_near(),
+            Self::Then { base, callback } => {
+                base.total_near().saturating_add(callback.total_near())
+            }
             Self::And(promises) => {
                 let total = promises.iter().map(|p| p.total_near().as_u128()).sum();
```
birchmd, line 141:
```suggestion
                let total = promises.iter().map(|p| p.total_near().as_u128()).fold(0, |a, b| a.saturating_add(b));
```

birchmd (commented):
I think the definitions in `PromiseArgs` and `SimpleNearPromise` should probably also be changed to saturating adds.

birchmd (approved):

aleksuss (approved):

PR #922, opened 2024-04-02
feat: pause FT transfers

karim-en:
## Description
Add the possibility of pausing the ft transfers for the internal eth-connector; this is needed to avoid pausing the whole Aurora engine during migration.
A new flag, `PAUSE_FT`, has been added that pauses all the public methods that could change the internal balances of the eth-connector token [`ft_transfer`, `ft_transfer_call`, `storage_deposit`, `storage_unregister`].

## Testing
Some tests have been added to test the pause and unpause logic.

### DIFF
--- a/engine-tests/src/utils/workspace.rs
+++ b/engine-tests/src/utils/workspace.rs
@@ -239,26 +239,30 @@ pub async fn transfer_nep_141(
 }
 
 #[cfg(not(feature = "ext-connector"))]
+pub async fn storage_deposit_nep141(
+    nep_141: &AccountId,
+    source: &Account,
+    dest: &str,
+) -> anyhow::Result<ExecutionFinalResult> {
+    source
+        .call(nep_141, "storage_deposit")
+        .args_json(json!({
+            "account_id": dest,
+        }))
+        .deposit(STORAGE_AMOUNT)
+        .max_gas()
+        .transact()
+        .await
+}
+
+#[cfg(not(feature = "ext-connector"))]
 pub async fn transfer_call_nep_141(
     nep_141: &AccountId,
     source: &Account,
     dest: &str,
     amount: u128,
     msg: &str,
-    storage_deposit: bool,
 ) -> anyhow::Result<ExecutionFinalResult> {
-    if storage_deposit {
-        let _ = source
-            .call(nep_141, "storage_deposit")
-            .args_json(json!({
-                "account_id": dest,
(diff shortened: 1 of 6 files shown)


engine-tests/src/utils/workspace.rs
```diff
@@ -236,6 +238,40 @@ pub async fn transfer_nep_141(
     Ok(())
 }
 
+pub async fn transfer_call_nep_141(
```
aleksuss, line 241: I believe it should be annotated with `#[cfg(not(feature = "ext-connector"))]`

engine/src/contract_methods/connector/errors.rs
```diff
@@ -118,6 +119,7 @@ impl AsRef<[u8]> for FtTransferCallError {
             Self::InsufficientAmountForFee => errors::ERR_NOT_ENOUGH_BALANCE_FOR_FEE,
             Self::Transfer(e) => e.as_ref(),
             Self::BalanceOverflow(e) => e.as_ref(),
+            Self::Paused => errors::ERR_PAUSED,
```
aleksuss, line 122: I would introduce a new error: `error::ERR_FT_PAUSED`.

aleksuss (commented):
Overall looks good with couple nitpicks.

engine-tests/src/utils/workspace.rs
karim-en: 219007bf9b048c152dd5f94abafd7126e9316331

engine/src/contract_methods/connector/errors.rs
karim-en: 4a6accd250b4312bc51315f4b452c61170f4c9d0

engine-tests/src/utils/workspace.rs
```diff
@@ -236,6 +238,40 @@ pub async fn transfer_nep_141(
...
+    nep_141: &AccountId,
+    source: &Account,
+    dest: &str,
+    amount: u128,
+    msg: &str,
+    storage_deposit: bool,
```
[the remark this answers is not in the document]
karim-en: The storage deposit was moved to a separate function d90c9c4d33970bc17d9e28b3acb5154a9316cb76

birchmd (approved):

aleksuss (approved):

mrLSD (approved):

PR #926, opened 2024-04-08
Feat: add CANCUN hard fork support

mrLSD:
## Description

➡️  Added **EVM CANCUN** hard fork support. Based on SputnikVM release [v0.42.0-aurora](https://github.com/aurora-is-near/sputnikvm/releases/tag/v0.42.0-aurora)
➡️  Changed `engine-tests` that related to **CACUN** changes
➡️  Solidity: updated to `0.8.25` which support **CANCUN** Features
➡️  Extended Transactions fields for **CANCUN** hard fork requirements

### Cancun hard fork

➡️  EVM impelentation of Cancun hard fork: aurora-is-near/sputnikvm/pull/39

### Breaking changes

➡️  Transaction extended according to [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844) requirements

### Gas cost
Gas cost changed as expected insignificantly.
- 2 tests: gas increased to `1 TGas`
- 1 test: gas decreased to `1 TGas`


engine-tests/src/utils/solidity/mod.rs
```diff
@@ -191,7 +191,7 @@ where
             &source_mount_arg,
             "-v",
             &output_mount_arg,
-            "ethereum/solc:0.8.24", // TODO: 0.8.25 introduces support of the Dencun hard fork.
+            "ethereum/solc:0.8.25", // TODO: 0.8.25 introduces support of the Dencun hard fork.
```
aleksuss, line 194: Let's rollback it to `stable`.

engine-transactions/src/eip_4844.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub nonce: U256,
+    /// Determined by the sender and is optional. Priority Fee is also known as Miner Tip as it is
+    /// paid directly to block producers.
+    pub max_priority_fee_per_gas: U256,
+    /// Maximum amount the sender is willing to pay to get their transaction included in a block.
+    pub max_fee_per_gas: U256,
+    /// The maximum amount of gas the sender is willing to consume on a transaction.
+    pub gas_limit: U256,
+    /// The receiving address (`None` for the zero address)
+    pub to: Option<Address>,
```
birchmd, line 35: According to EIP-4844, `to` is a required field (not allowed to be `None`). Therefore I think we should have `to: Address`. If a transaction is intended for the zero address they can explicitly specify the zero address.

engine-transactions/src/eip_4844.rs
```diff
@@ -0,0 +1,160 @@
...
+use crate::Error;
+use alloc::vec::Vec;
+use aurora_engine_precompiles::secp256k1::ecrecover;
+use aurora_engine_types::types::{Address, Wei};
+use aurora_engine_types::{H256, U256};
+use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
+
+/// Type indicator (per EIP-2718) for access list transactions
+/// [EIP-4844 Specification](https://eips.ethereum.org/EIPS/eip-4844#specification)
+pub const BLOB_TX_TYPE: u8 = 0x03;
```
aleksuss, line 11: I would rename it to `TYPE_BYTE` for consistency.

engine/src/engine.rs
```diff
@@ -1918,6 +1918,14 @@ impl<'env, I: IO + Copy, E: Env, M: ModExpAlgorithm> Backend for Engine<'env, I,
     fn original_storage(&self, address: H160, index: H256) -> Option<H256> {
         Some(self.storage(address, index))
     }
+
+    fn blob_gasprice(&self) -> Option<u128> {
+        Some(1)
```
aleksuss, line 1923: Should it be 0 instead?

engine-transactions/src/eip_4844.rs
mrLSD: Actually, it align with the specification: https://eips.ethereum.org/EIPS/eip-4844#specification
I'm not sure that renaming it for "_just good look_" is good reason enough.

engine/src/engine.rs
mrLSD: It's a good question. I decided that if: `MIN_BASE_FEE_PER_BLOB_GAS = 1` then we (_probably_) should put it to `1`.

engine-transactions/src/eip_4844.rs
aleksuss: Specification it's excellent, but **actually** we have  our own naming convention (such historically happens) for these constants: `eip_1559::TYPE_BYTE`, `eip_2930::TYPE_BYTE`
And I'm not sure that changing this convention to "_just to be like in the specification_" is a good reason enough.

mrLSD: Changed: 5ba5c12abaa8356357d4f23aa216d88ef72fb39d

engine-transactions/src/eip_4844.rs
mrLSD: Changed: 5ba5c12abaa8356357d4f23aa216d88ef72fb39d

mrLSD:
> I left a few comments on the EIP-4844 transaction implementation, but now I am wondering if we should remove the implementation and simply return an error saying that transaction type is unsupported.
> [...]
> What do you think @aleksuss @mrLSD ?

@birchmd We didn't plan to support `EIP-4844`. But the transaction type was added for compatibility reasons.

But you reasonably point out that it's better just to remove supporting that type of transaction.

My reason for not support `EIP-4844` on aurora-engine side is:

- it solves very specific Ethereum Network problems like temporary
blob data storage. It's the unnatural solution for Aurora on top NEAR.

- Ethereum full node should operate on it's own to manage and delete blob data
approximately after 4 weeks. Again as top on NEAR solution, it looks unreasonable at all.
It's hard to manage. Also, it's related to large amount of data.

- it's part of the way to Ethereum sharding. And it solves very specific Ethereum network problems.
Aurora as a solution top on NEAR does not reach that specific kind of problem, as this part is NEAR Protocol
related solution, but not Aurora as smarcontract on NEAR.

- And as a conclusion, it doesn't meet the concrete current requirement of Aurora to use blob data storage (and blob hashes as proof verifications)

birchmd:
Great, thanks @mrLSD . Since we agree EIP-4844 is not relevant to Aurora please remove the implementation and just return an error if we get a transaction with that type byte.

aleksuss:
My point was that we can know about the EIP-4844 transaction but are not obligated to execute it and just respond to a user that we do not support this type of transaction. But, of course, if we can understand the type of transaction by type byte then the implementation could be removed, and we don't need to deserialize it completely.

engine-transactions/src/eip_4844.rs
```diff
@@ -0,0 +1,158 @@
+use crate::eip_2930::AccessTuple;
+use crate::Error;
+use alloc::vec::Vec;
+use aurora_engine_precompiles::secp256k1::ecrecover;
+use aurora_engine_types::types::{Address, Wei};
+use aurora_engine_types::{H160, H256, U256};
+use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
+
+/// Type indicator (per EIP-2718) for access list transactions
```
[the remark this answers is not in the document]
mrLSD: Changed here: 353963cf3fa556151d3b875309e402a82639c83c

birchmd (approved):

aleksuss (approved):

raventid (approved):

PR #930, opened 2024-05-03
Feat: Implementation `EIP-3607`

mrLSD:
## Description

➡️ Implementation of [EIP-3607](https://eips.ethereum.org/EIPS/eip-3607) - reject transaction for accounts with not empty code.


engine/src/engine.rs
```diff
@@ -1032,6 +1034,10 @@ pub fn submit_with_alt_modexp<
         tx.try_into()
             .map_err(|_e| EngineErrorKind::InvalidSignature)?
     };
+    // Retrieve the signer of the transaction:
+    let sender = transaction.address;
+    // EIP-3706
+    check_empty_code(&io, &sender)?;
```
aleksuss, line 1040: Looks like here is not the best place for this check. Is it possible to do it at the EVM level?

mrLSD: Definitely not. `SpuntikVM` operates as an execution layer. `EIP-3607` about transaction validation. It's same a ga  preparation in `Submit` call.
For example in REVM all the same - It's part of transaction validation before EVM execution.

engine/src/engine.rs
```diff
@@ -1032,6 +1034,10 @@ pub fn submit_with_alt_modexp<
         tx.try_into()
             .map_err(|_e| EngineErrorKind::InvalidSignature)?
     };
+    // Retrieve the signer of the transaction:
+    let sender = transaction.address;
+    // EIP-3706
```
birchmd, line 1039:
```suggestion
    // EIP-3607
```

engine-tests/src/tests/promise_results_precompile.rs
```diff
@@ -119,14 +119,15 @@ fn test_promise_result_gas_cost() {
 
     let total_gas1 = y1 + baseline.all_gas();
     let total_gas2 = y2 + baseline.all_gas();
+
     assert!(
-        utils::within_x_percent(15, evm1, total_gas1 / NEAR_GAS_PER_EVM),
+        utils::within_x_percent(17, evm1, total_gas1 / NEAR_GAS_PER_EVM),
```
birchmd, line 124: I'm curious about why this seemingly small change has an impact on gas cost. Is the cost of reading that one key really so high? I wonder if it makes sense to go through `self.contract_code_cache` since then at least we are only reading the key once (assuming it is already being ready elsewhere in the sputnikVM code).

mrLSD: Unfortunately yes, it's related to storage read. It's not a part of SputnikVM, as it's a stage of transaction validation and not part of an Execution Layer.

@birchmd do you probably remember, why we do not validate transactions according to `eip-3607`? As it's 3 years old EIP. Or we just missed it?

Actually, I discovered that issue when it was tested `SpuntikVM` with all ethereum-json tests. Because current tests for SputnikVM skip a lot of tests, including `EIP-3607`.

birchmd: Yes I'm sure it was just an oversight that EIP-3607 was not included earlier.

I still think it is worth trying recording the read of the code into the cache just in case SputnikVM also tries to read the same key later.

mrLSD: @birchmd as you proposed added to `Engine` cache request.

birchmd: Can this value be lowered back to 15 with that change?

mrLSD:
**NOTE:** ➡️  CI but related to that issue: Near-One/aurora-eth-connector/issues/83

engine-tests/src/tests/promise_results_precompile.rs
birchmd: Thanks for double checking. It makes me wonder if we should revisit the `NEAR_GAS_PER_EVM` constant which this test is meant to check is accurate. A 17% deviation is a far cry from the 5% deviation this test started with. I suppose that can be done as a separate PR though.

birchmd (approved):

aleksuss (approved):

PR #933, opened 2024-06-27
Experimental: gas cost estination for `usize`

mrLSD:
## Description

That's the experimental usage of SpuntikVM with direct checking for `usize` casting overflow, which avoids `aurora-engine` panics.

## SputnikVM

Based on experimental PR: aurora-is-near/sputnikvm/pull/49

## Gas costs
- for some tests gas cost increase nearly `1%`
- for some tests gas cost increase nearly `10%`

### DIFF
--- a/engine-tests/src/tests/one_inch.rs
+++ b/engine-tests/src/tests/one_inch.rs
@@ -26,7 +26,7 @@ fn test_1inch_liquidity_protocol() {
 
     let (result, profile, pool_factory) = helper.create_pool_factory(&deployer_address);
     assert!(result.gas_used >= 2_800_000); // more than 2.8M EVM gas used
-    assert_gas_bound(profile.all_gas(), 9); // less than 9 NEAR Tgas used
+    assert_gas_bound(profile.all_gas(), 10); // less than 9 NEAR Tgas used
 
     // create some ERC-20 tokens to have a liquidity pool for
     let signer_address = utils::address_from_secret_key(&helper.signer.secret_key);
@@ -38,7 +38,7 @@ fn test_1inch_liquidity_protocol() {
     let (result, profile, pool) =
         helper.create_pool(&pool_factory, token_a.0.address, token_b.0.address);
     assert!(result.gas_used >= 4_500_000); // more than 4.5M EVM gas used
-    assert_gas_bound(profile.all_gas(), 18);
+    assert_gas_bound(profile.all_gas(), 20);
 
     // Approve giving ERC-20 tokens to the pool
     helper.approve_erc20_tokens(&token_a, pool.address());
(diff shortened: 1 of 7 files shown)


engine/src/engine.rs
```diff
@@ -1915,6 +1920,23 @@ impl<'env, I: IO + Copy, E: Env, M: ModExpAlgorithm> Backend for Engine<'env, I,
...
 
+    fn is_empty_storage(&self, address: H160) -> bool {
+        let address = Address::new(address);
+        if self
+            .contract_storage_cache
+            .borrow()
+            .contains_key(&(address, H256::zero()))
+        {
+            return true;
+        }
```
birchmd, line 1931: I don't think this implementation is correct. If the storage cache contains a key then it is not empty (should return `false`). Also what if `0x00` is not used, but some other slot is?

engine-tests/src/tests/one_inch.rs
```diff
@@ -26,7 +26,7 @@ fn test_1inch_liquidity_protocol() {
 
     let (result, profile, pool_factory) = helper.create_pool_factory(&deployer_address);
     assert!(result.gas_used >= 2_800_000); // more than 2.8M EVM gas used
-    assert_gas_bound(profile.all_gas(), 9); // less than 9 NEAR Tgas used
+    assert_gas_bound(profile.all_gas(), 10); // less than 9 NEAR Tgas used
```
birchmd, line 29: Gas costs going up a little is not ideal. But I think it is worth the trade off of not having panics in sputnikvm.

engine/src/engine.rs
mrLSD: In that particular case, we're just checking a storage cache.
But if it doesn't exist we'll just read storage by the key: (address, index), where index = 0;
Index zero should exist, as it's the impossible case in EVM to have index > 0, but don't have an index = 0;

Also, we can't store storage keys for an account, as it'll cost NEAR storage.

engine-tests/src/tests/one_inch.rs
mrLSD: In that case, it's `10%`. If you think it's normal - then no problem.

But also, we should charge gas for non-validation errors, from execution.

engine/src/engine.rs
birchmd: > it's the impossible case in EVM to have index > 0, but don't have an index = 0

Why is this the case? I thought you could write a value to any slot leaving the others unused. The storage usage convention in Solidity contracts is to store values for fields in the contract starting at index 0, but for arbitrary contracts written in EVM bytecode directly that does not need to be the case.

mrLSD: hm... it's problematic to get account storage status on `aurora-engine`. We don't know the account storage keys (indexes) for current EVM accounts. I believe it's the optimal solution for now to assume that `inxed=0` exists.
In the future, we can store account empty storage flag.

birchmd: Yeah, we'll need to think about a more robust solution to this problem. A flag is a nice idea. It'll require a storage migration for all existing accounts, but it's probably worth the effort to do this.

PR #935, opened 2024-07-05
Feat: refactore gas charge logic form EVM exit reason

mrLSD:
### Description

To be fully compatible with EVM, we should:
:arrow_right:  Success| Revert
:arrow_right:  ExitError - Execution errors should charge gas from users
:arrow_right:  ExitFatal - shouldn't charge user gas
:arrow_right:  Transactions validation errors should not charge user gas

This PR changed the logic for EVM exit reason gas charge for users, which includes EVM execution errors gas charge. Previously transaction just panicked, and user's gas was not charged.

#### EVM
Related to EVM changes:  aurora-is-near/sputnikvm/pull/52


engine-tests/src/tests/erc20.rs
```diff
@@ -12,8 +12,9 @@ use aurora_engine_types::account_id::AccountId;
 use aurora_engine_types::parameters::connector::{
     Erc20Identifier, Erc20Metadata, SetErc20MetadataArgs,
 };
-use aurora_engine_types::parameters::engine::SetOwnerArgs;
+use aurora_engine_types::parameters::engine::{SetOwnerArgs, TransactionStatusEvmErrorKind};
```
aleksuss, line 15: Could `TransactionStatusEvmErrorKind` be just `EvmErrorKind` cause it looks wired because the `TransactionStatusEvmErrorKind` is used in `TransactionStatus`?

engine-tests/src/tests/sanity.rs
```diff
@@ -1245,13 +1261,25 @@ mod workspace {
...
+        let res = aurora
+            .submit(rlp::encode(&signed_tx).to_vec())
+            .transact()
+            .await;
+        if let Err(err) = res {
+            println!("ERROR: {:#?}", err);
+        }
+
         let result = aurora
             .submit(rlp::encode(&signed_tx).to_vec())
```
aleksuss, line 1274: You sent here the same transaction a second time.

engine-tests/src/utils/solidity/exit_precompile.rs
```diff
@@ -94,7 +94,9 @@ impl Tester {
         match result.status {
             aurora_engine::parameters::TransactionStatus::Succeed(_) => Ok(result),
             aurora_engine::parameters::TransactionStatus::Revert(_bytes) => Err(Revert),
-            other => panic!("Unexpected status {other:?}"),
+            aurora_engine::parameters::TransactionStatus::Error(err) => {
+                panic!("Unexpected status {err:?}")
```
aleksuss, line 98:
```suggestion
                panic!("Unexpected status {:?}", err.as_ref());
```
?

engine-tests/src/tests/standalone/call_tracer.rs
```diff
@@ -296,10 +296,9 @@ fn test_contract_create_too_large() {
     let standalone_result = sputnik::traced_call(&mut listener, || {
         runner.submit_transaction(&signer.secret_key, tx)
     });
-    assert!(
-        standalone_result.is_err(),
-        "Expected contract too large error"
-    );
+    // As we charge users for the EVM errors, we should charge them for
+    // the gas used to deploy the large contract.
+    assert!(standalone_result.is_ok());
```
aleksuss, line 301: Should we check here error status like this:
```rust
assert!(matches!(
    standalone_result.unwrap().status,
    TransactionStatus::Error(EvmErrorKind::CreateContractLimit)
));
```
 ?

mrLSD: @birchmd as, you was author of that part - what do you thinks? Initially it was just check - is it contains errors.

engine-tests/src/utils/solidity/exit_precompile.rs
mrLSD: I decided that may be it will be more informative. As it'll return something like: `Error(GasOverflow)` but I can change it, obviously, if you think it will be more clear.

engine-sdk/src/caching.rs
```diff
@@ -10,7 +10,6 @@ impl<K: Ord, V> FullCache<K, V> {
     pub fn get_or_insert_with<F: FnOnce() -> V>(&mut self, k: K, f: F) -> &mut V {
         self.inner.entry(k).or_insert_with(f)
     }
-
```
aleksuss, line 13: I would rollback an empty line here.

engine/src/engine.rs
```diff
@@ -1920,27 +1961,23 @@ impl<'env, I: IO + Copy, E: Env, M: ModExpAlgorithm> Backend for Engine<'env, I,
...
-            .or_insert_with(|| get_generation(&self.io, &address));
-        // First we're checking cache to not produce read-storage operation
         if self
             .contract_storage_cache
             .borrow()
-            .contains_key(&(address, index))
+            .contains_key(&(address, H256::zero()))
         {
-            return false;
+            return true;
```
aleksuss, line 1973:
```suggestion
            return false;
```
?

aleksuss (approved):

engine-tests/src/utils/mod.rs
```diff
@@ -1000,11 +1000,15 @@ pub fn unwrap_revert_slice(result: &SubmitResult) -> &[u8] {
     }
 }
 
-pub fn panic_on_fail(status: TransactionStatus) {
+pub fn panic_on_fail(status: TransactionStatus) -> Option<String> {
```
birchmd, line 1003: Same comment as on other PR: https://github.com/aurora-is-near/aurora-engine/pull/947#discussion_r1716885738

engine-types/src/parameters/engine.rs
```diff
@@ -177,16 +177,79 @@ pub struct ResultLog {
...
 #[derive(Debug, Clone, BorshSerialize, BorshDeserialize, PartialEq, Eq)]
 #[cfg_attr(feature = "impl-serde", derive(Serialize, Deserialize))]
 pub enum TransactionStatus {
     Succeed(Vec<u8>),
     Revert(Vec<u8>),
-    OutOfGas,
-    OutOfFund,
-    OutOfOffset,
-    CallTooDeep,
+    Error(EvmErrorKind),
```
birchmd, line 252: This is a breaking change in the binary (borsh) format of `TransactionStatus`. For example, previously `TransactionStatus::OutOfGas` serialized as `0x02`, but now `TransactionStatus::Error(EvmErrorKind::OutOfGas)` will serialize as `0x020a`.

I think the safest thing to do is make this change backwards compatible by retaining the previous variants of `TransactionStatus` and adding `Error` as a new variant. Then the new `EvmErrorKind` should skip `OutOfGas`, `OutOfFund`, `OutOfOffset` and `CallTooDeep` so that we do not have duplicate representations of the same error.

New tests should be added checking that we catch the new error kinds that will appear in the new `TransactionStatus` variant.

birchmd (changes requested):

engine-types/src/parameters/engine.rs
mrLSD: First suggestion - I agree.

About tests, for me it's a bit unclear.
Do you mean all variants of `EmvErrorKind`?

For example: `test_contract_create_too_large` cover: `TransactionStatus::Error(EvmErrorKind::CreateContractLimit)`

birchmd: For example it would be good to add a test that covers `TransactionStatus::Error(EvmErrorKind::StackUnderflow)`.

mrLSD: Am I clear, that your point also to check that borsh serialization matches the expected results? As currently there are no tests for that.

birchmd: Yes, it would also be a good idea to have regression tests for the borsh serialization. I think it makes sense to test the serialization of `SubmitResult` which contains `TransactionStatus`.

birchmd (approved):

PR #936, opened 2024-07-17
Feat: Update dependencies & toolchain

mrLSD:


engine-standalone-storage/src/lib.rs
```diff
@@ -227,7 +227,7 @@ impl Storage {
     }
 
     fn process_transaction<F: Fn(&mut rocksdb::WriteBatch, &[u8], &[u8])>(
-        &mut self,
+        &self,
```
birchmd, line 230: I know that technically this doesn't mutate `self`, but it does internally because of the `self.db.write` call at the end of the function. I recommend adding an `allow` for whatever clippy lint is given an error on this function.

engine/src/engine.rs
```diff
@@ -816,10 +816,9 @@ impl<'env, I: IO + Copy, E: Env, M: ModExpAlgorithm> Engine<'env, I, E, M> {
                     gas_used: submit_result.gas_used,
                 }),
             })
-            .map_err(|e| {
-                sdk::log!("{:?}", e);
+            .inspect_err(|_e| {
+                sdk::log!("{:?}", _e);
```
birchmd, line 820:
```suggestion
            .inspect_err(|e| {
                sdk::log!("{:?}", e);
```

engine-tests/src/tests/repro.rs
```diff
@@ -26,7 +26,7 @@ fn repro_GdASJ3KESs() {
         block_timestamp: 1_645_717_564_644_206_730,
         input_path: "src/tests/res/input_GdASJ3KESs.hex",
         evm_gas_used: 706_713,
-        near_gas_used: 117,
+        near_gas_used: 114,
```
birchmd, line 29: It's always nice to see the gas costs decrease when we upgrade the compiler version!

birchmd (commented):
Overall looks good with just a few comments. Do you know why the tests are failing though?

mrLSD:
>  Do you know why the tests are failing though?

@birchmd it's good question, as in some tests state trie check fails. i'm still investigating it.

engine/src/engine.rs
aleksuss: The `_e` variable couldn't be used if the feature `log` is not provided.

aleksuss (approved):

birchmd (approved):

PR #939, opened 2024-07-22
chore: release notes 3.6.4

aleksuss:
## Description

Release notes for the 3.6.4 release.


birchmd (commented):
Is this also the first release that includes the Cancun upgrade? That probably deserves a mention in the release notes.

aleksuss:
No. I'm planning to add only the commit mentioned in the CHANGELOG. I think it's better to add CANCUN support after splitting eth connector. Just in case.

birchmd (approved):

mrLSD (approved):

PR #942, opened 2024-08-01
chore: remove unused `bytes_to_hex`

dwiekawki:


birchmd (commented):
I don't think this function is used in our codebase any more. Probably it makes sense just to remove it instead of improving it.

engine-types/src/types/mod.rs
```diff
@@ -101,7 +101,7 @@ const HEX_ALPHABET: &[u8; 16] = b"0123456789abcdef";
 #[allow(dead_code)]
 #[must_use]
 pub fn bytes_to_hex(v: &[u8]) -> String {
-    let mut result = String::new();
+    let mut result = String::with_capacity(v.len() << 1);
```
mrLSD, line 104: [This function used](https://github.com/aurora-is-near/aurora-engine/blob/c03a2d8610cd27a9decb91b3bddb107db2177b29/engine-types/src/types/mod.rs#L161)  only for testing itself.
I think it's better to just remove that code, as it's outdated.

dwiekawki: Removed

birchmd (approved):

aleksuss (approved):

mrLSD (approved):

PR #947, opened 2024-08-14
Chore: Update SputnikVM to v0.45.2-aurora

mrLSD:


aleksuss (approved):

engine-tests/src/utils/mod.rs
```diff
@@ -1000,11 +1000,15 @@ pub fn unwrap_revert_slice(result: &SubmitResult) -> &[u8] {
     }
 }
 
-pub fn panic_on_fail(status: TransactionStatus) {
+pub fn panic_on_fail(status: TransactionStatus) -> Option<String> {
```
birchmd, line 1003: Maybe `get_error_message` is a better name for this function now that it no longer panics.

engine/src/engine.rs
```diff
@@ -1920,27 +1961,23 @@ impl<'env, I: IO + Copy, E: Env, M: ModExpAlgorithm> Backend for Engine<'env, I,
...
-            .or_insert_with(|| get_generation(&self.io, &address));
-        // First we're checking cache to not produce read-storage operation
         if self
             .contract_storage_cache
             .borrow()
-            .contains_key(&(address, index))
+            .contains_key(&(address, H256::zero()))
         {
-            return false;
+            return true;
```
birchmd, line 1973:
```suggestion
            return false;
```

How does this incorrect logic pass our CI? Maybe new test that touches `is_empty_storage` is needed.

PR #949, opened 2024-08-26
chore: use plain enum for the TransactionStatus

aleksuss:
## Description

The PR proposes adding some code refactoring to the #935 

No performance changes.

## Testing

A regression test for the `TransactionStatus` has been added.


birchmd (approved):

engine-types/src/parameters/engine.rs
```diff
@@ -575,51 +570,47 @@ mod tests {
...
-        let res = borsh::to_vec(&TransactionStatus::Error(EvmErrorKind::PCUnderflow)).unwrap();
-        assert_eq!(&[6, 8], res.as_slice());
-        let res = borsh::to_vec(&TransactionStatus::Error(EvmErrorKind::CreateEmpty)).unwrap();
-        assert_eq!(&[6, 9], res.as_slice());
-        let res = borsh::to_vec(&TransactionStatus::Error(EvmErrorKind::Other(
-            crate::Cow::from(String::new()),
-        )))
+    fn test_serialization_transaction_status_regression() {
+        let bytes =
+            std::fs::read("../engine-tests/src/tests/res/transaction_status.borsh").unwrap();
```
mrLSD, line 575: I understand, why that solution was proposed: some guaranty or reference of transaction order.
But nothing can prevent the change that file. It means that the serialization generator can be used directly, as it will be difficult to miss ordering modifications for it.
Also keep in mind, that new EVM errors will be frequently adding.

mrLSD (approved):

PR #960, opened 2024-09-26
Feat: Add EIP-7702 - new type transaction

mrLSD:
## Description

➡️  Added EIP-7702 - new type transaction

### DIFF
--- a/engine-tests/src/tests/repro.rs
+++ b/engine-tests/src/tests/repro.rs
@@ -26,7 +26,7 @@ fn repro_GdASJ3KESs() {
         block_timestamp: 1_645_717_564_644_206_730,
         input_path: "src/tests/res/input_GdASJ3KESs.hex",
         evm_gas_used: 706_713,
-        near_gas_used: 115,
+        near_gas_used: 113,
     });
 }
 
@@ -51,7 +51,7 @@ fn repro_8ru7VEA() {
         block_timestamp: 1_648_829_935_343_349_589,
         input_path: "src/tests/res/input_8ru7VEA.hex",
         evm_gas_used: 1_732_181,
-        near_gas_used: 206,
+        near_gas_used: 202,
     });
 }
 
@@ -71,7 +71,7 @@ fn repro_FRcorNv() {
         block_timestamp: 1_650_960_438_774_745_116,
         input_path: "src/tests/res/input_FRcorNv.hex",
         evm_gas_used: 1_239_721,
-        near_gas_used: 168,
+        near_gas_used: 163,
     });
 }
 
@@ -88,7 +88,7 @@ fn repro_5bEgfRQ() {
         block_timestamp: 1_651_073_772_931_594_646,
         input_path: "src/tests/res/input_5bEgfRQ.hex",
         evm_gas_used: 6_414_105,
-        near_gas_used: 650,
+        near_gas_used: 508,
     });
 }
 
@@ -106,7 +106,7 @@ fn repro_D98vwmi() {
         block_timestamp: 1_651_753_443_421_003_245,
         input_path: "src/tests/res/input_D98vwmi.hex",
         evm_gas_used: 1_035_348,
-        near_gas_used: 169,
+        near_gas_used: 164,
     });
 }
 
@@ -125,7 +125,7 @@ fn repro_Emufid2() {
         block_timestamp: 1_662_118_048_636_713_538,
         input_path: "src/tests/res/input_Emufid2.hex",
         evm_gas_used: 1_156_364,
-        near_gas_used: 294,
+        near_gas_used: 293,
     });
 }
 
--- a/engine-transactions/src/eip_7702.rs
+++ b/engine-transactions/src/eip_7702.rs
@@ -1,0 +1,262 @@
+use crate::eip_2930::AccessTuple;
+use crate::Error;
+use aurora_engine_precompiles::secp256k1::ecrecover;
+use aurora_engine_types::types::{Address, Wei};
+use aurora_engine_types::{Vec, H160, U256};
+use evm::executor::stack::Authorization;
+use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+/// Type indicator (per EIP-7702)
+pub const TYPE_BYTE: u8 = 0x04;
+
+// EIP-7702 `MAGIC` number
+pub const MAGIC: u8 = 0x5;
+
+/// The order of the secp256k1 curve, divided by two. Signatures that should be checked according
+/// to EIP-2 should have an S value less than or equal to this.
+///
+/// `57896044618658097711785492504343953926418782139537452191302581570759080747168`
+pub const SECP256K1N_HALF: U256 = U256([
+    0xDFE9_2F46_681B_20A0,
+    0x5D57_6E73_57A4_501D,
+    0xFFFF_FFFF_FFFF_FFFF,
+    0x7FFF_FFFF_FFFF_FFFF,
+]);
+
+#[derive(Debug, Eq, PartialEq, Clone)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct AuthorizationTuple {
+    pub chain_id: U256,
+    pub address: H160,
+    pub nonce: u64,
+    pub parity: U256,
+    pub r: U256,
+    pub s: U256,
+}
+
+impl Decodable for AuthorizationTuple {
+    fn decode(rlp: &Rlp<'_>) -> Result<Self, DecoderError> {
+        let chain_id = rlp.val_at(0)?;
+        let address = rlp.val_at(1)?;
+        let nonce = rlp.val_at(2)?;
+        let parity = rlp.val_at(3)?;
+        let r = rlp.val_at(4)?;
+        let s = rlp.val_at(5)?;
+        Ok(Self {
+            chain_id,
+            address,
+            nonce,
+            parity,
+            r,
+            s,
+        })
+    }
+}
+
+/// EIP-7702 transaction kind from the Prague hard fork.
+///
+/// See [EIP-7702](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md)
+/// for more details.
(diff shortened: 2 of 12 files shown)


engine-transactions/src/eip_7702.rs
```diff
@@ -0,0 +1,262 @@
...
+use serde::{Deserialize, Serialize};
+
+/// Type indicator (per EIP-7702)
+pub const TYPE_BYTE: u8 = 0x04;
+
+// EIP-7702 `MAGIC` number
+pub const MAGIC: u8 = 0x5;
+
+/// The order of the secp256k1 curve, divided by two. Signatures that should be checked according
+/// to EIP-2 should have an S value less than or equal to this.
```
aleksuss, line 18: EIP-2 or EIP-7702 ?

aleksuss:
It would be good to add some integration tests/examples which demonstrate the new functionality.

engine-transactions/src/eip_7702.rs
```diff
@@ -0,0 +1,262 @@
...
+    pub nonce: U256,
+    /// Determined by the sender and is optional. Priority Fee is also known as Miner Tip as it is
+    /// paid directly to block producers.
+    pub max_priority_fee_per_gas: U256,
+    /// Maximum amount the sender is willing to pay to get their transaction included in a block.
+    pub max_fee_per_gas: U256,
+    /// The maximum amount of gas the sender is willing to consume on a transaction.
+    pub gas_limit: U256,
+    /// The receiving address (`None` for the zero address).
+    pub to: Option<Address>,
```
birchmd, line 76: I think we should model this as simply `pub to: Address`. The EIP-7702 documentation clearly states `a null destination is not valid`, so we should not allow `to: None`. If a user wants `to` equal to the zero address then they must explicitly set `to: 0x00`.

engine-transactions/src/eip_7702.rs
```diff
@@ -0,0 +1,262 @@
...
+            }
+            // EIP-2 validation
+            if auth.s > SECP256K1N_HALF {
+                return Err(Error::InvalidAuthorizationSignature);
+            }
+
+            // According to EIP-7702 step 1. validation, we should verify is `chain_id` is related to current network.
+            // We just ensure that `chain_id` is not zero, as it's possible different `chain_id` and
+            // `chain_id == 0` is related to Ethereum `mainnet`, that's not correct for Aurora.
+            let mut is_valid = !auth.chain_id.is_zero();
```
birchmd, line 187: This check does not look right to me. The documentation says `Verify the chain id is either 0 or the chain's current ID`. So 0 is always allowed regardless of the current chain and any non-zero value must match the current chain id. This check does neither of these things because it always rejects 0 and accepts any non-zero value.

Instead I think the expected chain id should be passed as an argument to this function and the check should be `auth.chain_id.is_zero() || auth.chain_id == expected_chain_id`.

engine-transactions/src/eip_7702.rs
```diff
@@ -0,0 +1,262 @@
...
+
+            // 2. Checking: authority = ecrecover(keccak(MAGIC || rlp([chain_id, address, nonce])), y_parity, r, s])
+            let mut rlp_stream = RlpStream::new();
+            rlp_stream.begin_list(4);
+            rlp_stream.append(&MAGIC);
+            rlp_stream.append(&auth.chain_id);
+            rlp_stream.append(&auth.address);
+            rlp_stream.append(&auth.nonce);
+
+            let signature_hash = aurora_engine_sdk::keccak(rlp_stream.as_raw());
```
birchmd, line 197: I don't think this matches the spec. I think it's supposed to be

```rust
let mut rlp_stream = RlpStream::new();
rlp_stream.begin_list(3);
rlp_stream.append(&auth.chain_id);
rlp_stream.append(&auth.address);
rlp_stream.append(&auth.nonce);

let message_bytes = [&[MAGIC], rlp_stream.as_raw()].concat();

let signature_hash = aurora_engine_sdk::keccak(&message_bytes);
```

birchmd (changes requested):

engine-transactions/src/eip_7702.rs
mrLSD: I know about `chain_id = 0`. We intended to simplify it as we do not pass `chain_id` to `NormalizedTransaction`.
it. I'm agree it doesn't look like a good idea and it should be strictly related to EIP.

mrLSD: Do you have an idea how to manage `current_chain_id` for:
```
impl TryFrom<EthTransactionKind> for NormalizedEthTransaction {
    type Error = Error;

    fn try_from(kind: EthTransactionKind) -> Result<Self, Self::Error> { ... }
```

birchmd: Hmm I see that doesn't quite work. Perhaps the validation logic should be separate from constructing `NormalizedEthTransaction`. Maybe it should exist as part of the `submit` function logic after the transaction is parsed.

mrLSD: Agree.

mrLSD: Solved: 976248fed50c0526337ea8e945021a72a79ca66e

engine-transactions/src/eip_7702.rs
```diff
@@ -0,0 +1,365 @@
...
+            &super::vrs_to_arr(self.parity, self.r, self.s),
+        )
+        .map_err(|_e| Error::EcRecover)
+    }
+
+    pub fn authorization_list(&self) -> Result<Vec<Authorization>, Error> {
+        if self.transaction.authorization_list.is_empty() {
+            return Err(Error::EmptyAuthorizationList);
+        }
+        let current_tx_chain_id = U256::from(self.transaction.chain_id);
```
birchmd, line 160: Good idea taking the current chain id from the transaction itself!

engine-tests/src/tests/repro.rs
```diff
@@ -88,7 +88,7 @@ fn repro_5bEgfRQ() {
         block_timestamp: 1_651_073_772_931_594_646,
         input_path: "src/tests/res/input_5bEgfRQ.hex",
         evm_gas_used: 6_414_105,
-        near_gas_used: 650,
+        near_gas_used: 508,
```
birchmd, line 91: It's always nice to see the gas costs go down when we move to a newer version of sputnikVM.

birchmd (commented):
The implementation looks good to me now, nice work. I still think it would be nice to have some integration tests for this feature. I think the following tests would be good to have:

- Happy path: EOA delegates code to a contract and then a transaction to that EOA executes successfully
- Error path: Delegation fails for incorrect nonce
- Error path: Delegation fails for incorrect chain id
- Error path: Delegation fails for invalid signature

mrLSD:
@birchmd @aleksuss for this PR was:
➡️ Changed EVM config to: `Prague`. In other way it's impossible to pass integration tests
➡️ For now PR related with #988 - Prague hard fork for Engine
➡️ Added unit tests for `EIP-7702`: nonce, signature
➡️ Added integration tests for `EIP-7702`. It's a bit tricky, as we can'y directly detect in integrations tests, that `auth_list` is wrong. So I invented some smartcontract code to simulate that situation and cover some wrong executions. It simulates new types of transactions and some interesting aspects. Keep in mind, that the most important EVM tests inside `SputnikVM` itself. For Engine most important to validate new transactions itself and creating `auth_list`. It was covered for 💯 .
➡️ As you can see in #988 - gas consumption was changed. Changes was merged from #988

engine-tests/src/tests/transaction.rs
```diff
@@ -161,6 +173,129 @@ fn test_access_list_tx_encoding_decoding() {
...
+    assert_eq!(runner.get_balance(contract_address), CONTRACT_BALANCE);
+    assert_eq!(runner.get_nonce(contract_address), CONTRACT_NONCE.into());
+    assert_eq!(runner.get_code(contract_address), contract_code);
+    // `EXTCODESIZE` should return zero, as `authorization_list` failed validation
+    assert_eq!(
+        runner.get_storage(contract_address, H256::zero()),
+        H256::zero()
+    );
+
+    // Authority address should increase Nonce
```
birchmd, line 293:
```suggestion
    // Authority address should not increase Nonce because authorization failed
```

engine-tests/src/tests/transaction.rs
```diff
@@ -161,6 +173,129 @@ fn test_access_list_tx_encoding_decoding() {
...
+    assert_eq!(runner.get_nonce(delegated_designator), 1.into());
+    // Get delegated designator address
+    assert_eq!(
+        hex::encode(runner.get_code(delegated_designator)),
+        "ef0100cccccccccccccccccccccccccccccccccccccccc"
+    );
+}
+
+#[test]
+fn test_eip_7702_wrong_auth_nonce() {
```
birchmd, line 251: Is the name of this test correct? It looks to me like the reason the authorization fails is because the chain id is wrong not because the nonce is wrong.

birchmd (approved):
Nice job, thanks for adding tests.

engine-tests/src/tests/transaction.rs
mrLSD: Absolutely! Both changes in: 69dd68a764d13ae1b2b4efc2868e21d5be602004

engine-tests/src/utils/mod.rs
```diff
@@ -25,17 +37,6 @@ use rlp::RlpStream;
...
-#[cfg(not(feature = "ext-connector"))]
-use crate::prelude::parameters::InitCallArgs;
-use crate::prelude::parameters::{StartHashchainArgs, SubmitResult, TransactionStatus};
-use crate::prelude::transactions::{
-    eip_1559::{self, SignedTransaction1559, Transaction1559},
-    eip_2930::{self, SignedTransaction2930, Transaction2930},
-    legacy::{LegacyEthSignedTransaction, TransactionLegacy},
-};
-use crate::prelude::{sdk, Address, Wei, H256, U256};
-use crate::utils::solidity::{ContractConstructor, DeployedContract};
```
aleksuss, line 37: For some reason, the above imports moved in the beginning. I think it was done unintentionally.

mrLSD: Looks like autoformat

aleksuss (approved):

PR #966, opened 2024-10-17
Chore: Update SputnikVM to v0.46.1-aurora

mrLSD:
## Description

🚀  Updated SputnikVM (EVM) to `v0.46.1-aurora`

### Gas cost

For some tests gas costs decreased.

### DIFF
--- a/engine-tests/src/tests/repro.rs
+++ b/engine-tests/src/tests/repro.rs
@@ -71,9 +71,6 @@ fn repro_FRcorNv() {
         block_timestamp: 1_650_960_438_774_745_116,
         input_path: "src/tests/res/input_FRcorNv.hex",
         evm_gas_used: 1_239_721,
-        #[cfg(not(feature = "ext-connector"))]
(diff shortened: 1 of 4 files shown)


engine-tests/src/tests/repro.rs
```diff
@@ -74,7 +74,7 @@ fn repro_FRcorNv() {
         #[cfg(not(feature = "ext-connector"))]
         near_gas_used: 168,
         #[cfg(feature = "ext-connector")]
-        near_gas_used: 167,
+        near_gas_used: 168,
```
birchmd, line 77: Funny that one gas cost went up while another went down.

birchmd (approved):

engine-tests/src/tests/repro.rs
aleksuss: I guess we can get back to only one field because now the value is the same.

aleksuss (approved):

PR #973, opened 2024-11-12
ci: fix installing nodejs deps in the tests

aleksuss:
## Description

The PR fixes breaking tests that are related to the execution in the CI only.

### DIFF
--- a/engine-tests/src/utils/one_inch/mod.rs
+++ b/engine-tests/src/utils/one_inch/mod.rs
@@ -14,8 +14,8 @@ pub static LIMIT_ORDER_PROTOCOL_PATH: LazyLock<PathBuf> =
 
 fn download_and_compile_solidity_sources(repo_name: &str) -> PathBuf {
     let target_dir = Path::new(env!("CARGO_MANIFEST_DIR")).join("target");
-    let lock_file = target_dir.join(&format!("{repo_name}.lock"));
-    // Using file lock to prevent parallel execution from the different processes in GitHub CI.
+    let lock_file = target_dir.join(format!("{repo_name}.lock"));
+    // Using a file lock to prevent parallel execution from the different processes in GitHub CI.
     let mut lock = fslock::LockFile::open(&lock_file).unwrap();
 
     lock.lock().unwrap();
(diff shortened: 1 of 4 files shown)


mrLSD (approved):

engine-tests/src/utils/one_inch/mod.rs
```diff
@@ -13,7 +13,8 @@ pub static LIMIT_ORDER_PROTOCOL_PATH: LazyLock<PathBuf> =
     LazyLock::new(|| download_and_compile_solidity_sources("limit-order-protocol"));
 
 fn download_and_compile_solidity_sources(repo_name: &str) -> PathBuf {
-    let sources_dir = Path::new("target").join(repo_name);
+    let process_id = std::process::id();
```
birchmd, line 16: I'm not sure why this fixes the issue. The purpose of downloading to a fixed sub-directory of `target` was so that the download would only need to happen once (since we cache the target directory). With this change the repo is downloaded fresh each time the tests are run and also still cached so that the size of `target` will grow over time.

If the repo needed to be re-downloaded because of some corrupted state in the CI runner then I think it would be better to clear the cache instead of modifying the code.

aleksuss: > I'm not sure why this fixes the issue.

It does. But it's not elegant :smiling. I added it to just check it out that the problem related to the running mulitple process in one env.

> The purpose of downloading to a fixed sub-directory of target was so that the download would only need to happen once (since we cache the target directory).

The directory with the solidity contracts is `./engine-tests/target`. I'm not sure that we're caching it.

> With this change the repo is downloaded fresh each time the tests are run and also still cached so that the size of target will grow over time.

Agreed. And because of that I'm thinking about using the `FileLock` instread.

But the main problem, as I can see, is that mutiple processes work with one directory.

birchmd: > I'm not sure that we're caching it.

Hmm, I think it would be useful if we did cache it. It seems wasteful to download fresh each time.

> I'm thinking about using the `FileLock` instead.

That sounds like a good solution!

birchmd (approved):

aleksuss:
Sorry, guys, for the changes after review, but I guess I realised that file lock is redundant here because it looks like the `target` folder in the root project folder can't be shared between processes that execute the tests. It seems the problem was related to the directory `engine-tests/target`. Anyway, the issue isn't reproduced after changing the `source_dir` from `engine-tests/target` to the `target`.

birchmd (approved):

PR #974, opened 2024-11-13
ci: use k8 self-hosted runners

aleksuss:
## Description

The PR changes the usage of legacy self-hosted runners to the new ones, k8, for some tasks. The tasks which run tests are still using GitHub runners, since on the k8 runners, the tests fail from time to time (probably because of sharing resources between workers).


kogeler:
@aleksuss I deployed new copy of k8s runners that reserve CPU and RAM. I've added it here, a test job takes about ~8 minutes now and should avoid the timeout issue at all times

aleksuss:
@kogeler the [issue](https://github.com/aurora-is-near/aurora-engine/actions/runs/12004669575/job/33460008835) is still reproduced.

kogeler:
> @kogeler the [issue](https://github.com/aurora-is-near/aurora-engine/actions/runs/12004669575/job/33460008835) is still reproduced.

I moved the workers to dedicated k8s nodes. It should work batter.

mrLSD (approved):

PR #975, opened 2024-11-14
ci: bump node up to 18 in scheduled script

aleksuss:
## Description

The PR bumps node up to 18 in the ci script for scheduled tasks.


.github/workflows/scheduled_lints.yml
```diff
@@ -41,8 +41,8 @@ jobs:
       - name: Setup Node and cache
         uses: actions/setup-node@v4
         with:
-          node-version: 16
-          cache: 'yarn'
+          node-version: 18
```
mrLSD, line 44: Why not LTS? Or even `v20`?

engine-tests/src/tests/promise_results_precompile.rs
```diff
@@ -101,7 +101,7 @@ fn test_promise_result_gas_cost() {
 
     assert!(
         utils::within_x_percent(
-            5,
+            10,
```
mrLSD, line 104: It's `2x`. Any ideas why?

aleksuss: Because we've got the [error](https://github.com/aurora-is-near/aurora-engine/actions/runs/11834374670/job/32974926364#step:8:1227) from time to time on the CI. On localhost I've got: 

`base cost: 107 and promise result base cost: 111`

which is within 3.6 % 
But 102 and 111 within 8.1 %.

.github/workflows/scheduled_lints.yml
aleksuss: Because the `hardhat@2.12.7` doesn't support 20.x.x

mrLSD (approved):

PR #982, opened 2024-12-26
Feat: update primitive types & refactoring

mrLSD:
## Description

Updated dependencies:
➡️  `primitive-types 0.13.1`
➡️  `sputnikvm 0.47-beta.1`
➡️  Refactored `ethabi` types transformation
➡️  Refactored `primitive-types` transformation
➡️  Refactored tests

### DIFF
--- a/engine-precompiles/src/native.rs
+++ b/engine-precompiles/src/native.rs
@@ -131,10 +131,10 @@ pub mod events {
                 self.amount.to_big_endian(),
             ))]);
             let topics = vec![
-                ethabi::Hash::from(EXIT_TO_ETH_SIGNATURE.0),
-                ethabi::Hash::from(encode_address(self.sender).0),
-                ethabi::Hash::from(encode_address(self.erc20_address).0),
-                ethabi::Hash::from(encode_address(self.dest).0),
+                EXIT_TO_ETH_SIGNATURE.0.into(),
+                encode_address(self.sender).0.into(),
+                encode_address(self.erc20_address).0.into(),
+                encode_address(self.dest).0.into(),
             ];
 
             ethabi::RawLog { topics, data }
@@ -518,7 +518,7 @@ impl<I: IO> Precompile for ExitToNear<I> {
             address: exit_to_near::ADDRESS.raw(),
             topics: exit_event_log
                 .topics
-                .iter()
+                .into_iter()
                 .map(|h| H256::from(h.0))
                 .collect(),
             data: exit_event_log.data,
--- a/engine-precompiles/src/native.rs
+++ b/engine-precompiles/src/native.rs
@@ -7,7 +7,7 @@ use crate::prelude::{
     storage::{bytes_to_key, KeyPrefix},
     str,
     types::{Address, Yocto},
-    vec, Cow, ToString, Vec, U256,
+    vec, Cow, ToString, Vec, H256, U256,
 };
 #[cfg(feature = "error_refund")]
 use crate::prelude::{parameters::RefundCallArgs, types};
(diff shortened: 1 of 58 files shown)


engine-precompiles/src/native.rs
```diff
@@ -122,12 +127,14 @@ pub mod events {
...
-            let data = ethabi::encode(&[ethabi::Token::Int(self.amount)]);
+            let data = ethabi::encode(&[ethabi::Token::Int(ethabi::Uint::from(
+                self.amount.to_big_endian(),
+            ))]);
             let topics = vec![
-                EXIT_TO_ETH_SIGNATURE,
-                encode_address(self.sender),
-                encode_address(self.erc20_address),
-                encode_address(self.dest),
+                ethabi::Hash::from(EXIT_TO_ETH_SIGNATURE.0),
```
aleksuss, line 134: Would be shorter:
```suggestion
                EXIT_TO_ETH_SIGNATURE.0.into(),
```

engine-precompiles/src/native.rs
```diff
@@ -509,7 +516,11 @@ impl<I: IO> Precompile for ExitToNear<I> {
         let exit_event_log = exit_event.encode();
         let exit_event_log = Log {
             address: exit_to_near::ADDRESS.raw(),
-            topics: exit_event_log.topics,
+            topics: exit_event_log
+                .topics
+                .iter()
```
aleksuss, line 521:
```suggestion
                .into_iter()
```

engine-tests-connector/src/connector.rs
```diff
@@ -506,14 +506,14 @@ async fn test_deposit_with_0x_prefix() -> anyhow::Result<()> {
...
-            event_schema.signature(),
+            H256::from(event_schema.signature().0),
             // the sender is not important
             H256::zero(),
         ],
         data: ethabi::encode(&[
             ethabi::Token::String(message),
-            ethabi::Token::Uint(U256::from(deposit_event.amount.as_u128())),
-            ethabi::Token::Uint(U256::from(deposit_event.fee.as_u128())),
+            ethabi::Token::Uint(ethabi::Uint::from(deposit_event.amount.as_u128())),
```
aleksuss, line 515:
```suggestion
            ethabi::Token::Uint(deposit_event.amount.as_u128().into()),
```

aleksuss (commented):
Overall looks good, but I would change the `from`s to `into`s.

aleksuss (approved):

mrLSD:
@aleksuss LGTM last your refactoring.

engine-tests/src/tests/repro.rs
```diff
@@ -88,7 +88,7 @@ fn repro_5bEgfRQ() {
         block_timestamp: 1_651_073_772_931_594_646,
         input_path: "src/tests/res/input_5bEgfRQ.hex",
         evm_gas_used: 6_414_105,
-        near_gas_used: 649,
+        near_gas_used: 508,
```
birchmd, line 91: Quite a significant decrease! 🚀

engine-tests/src/utils/solidity/uniswap.rs
```diff
@@ -214,17 +214,17 @@ impl PositionManager {
             .function("mint")
             .unwrap()
             .encode_input(&[ethabi::Token::Tuple(vec![
-                ethabi::Token::Address(params.token0.raw()),
-                ethabi::Token::Address(params.token1.raw()),
+                ethabi::Token::Address(params.token0.raw().0.into()),
```
birchmd, line 217: Looks like there is a lot of this kind of type conversion now. I assume this is because `ethabi` is no longer actively maintained and thus depends on an old version of the Ethereum primitive types. I wonder if we should look into migrating to the `alloy` library some time in the future.

birchmd (approved):
LGTM

engine-tests/src/utils/solidity/uniswap.rs
aleksuss: Yes. Exactly, that's because the `ethabi` is no longer supported. But it's a good point about the `alloy` crate. We've already discussed it with @mrLSD.

PR #984, opened 2025-01-07
chore: extend clippy coverage

aleksuss:
## Description

The PR includes:

- extends clippy linter coverage of the crates
- adds new syntax for describing lints to check
- adds a new job to check that the XCC router is built

Some insignificant gas changes.


mrLSD (approved):

engine-modexp/src/arith.rs
```diff
@@ -399,19 +399,8 @@ pub fn join_as_double(hi: Word, lo: Word) -> DoubleWord {
 
 #[test]
 fn test_monsq() {
-    check_monsq(1, 31);
-    check_monsq(6, 31);
     // This example is intentionally chosen because 5 * 5 = 25 = 0 mod 25,
```
birchmd, line 402: I think it makes sense to keep this comment with the `check_monsq(5, 25)` test case.

engine-tests/src/tests/standard_precompiles.rs
```diff
@@ -55,7 +55,7 @@ fn profile_identity() {
 #[test]
 fn profile_modexp() {
     let profile = precompile_execution_profile("test_modexp");
-    utils::assert_gas_bound(profile.all_gas(), 8);
+    utils::assert_gas_bound(profile.all_gas(), 9);
```
birchmd, line 58: Strange that linting changes made the execution less efficient. It's not by much, so it's probably fine, but unexpected none the less.

birchmd (approved):

PR #997, opened 2025-02-12
ci: update security workflow

diegofigs:
## Description

Updates security workflow

N/A

## Testing

N/A

N/A

## Additional information

Previous version of workflow was targeted specifically for contracts, newer version is generic and supports different project configurations apart from contracts and has stronger/up-to-date security tooling


mrLSD:
@diegofigs we runs `clippy` and other `engine`-related tools in a different way, via `cargo make`.

.github/workflows/contract-analysis.yml
```diff
@@ -9,9 +10,8 @@ on:
     - cron: "30 5 * * *"
   workflow_dispatch:
 
-name: "Contract Security Analysis"
 jobs:
   contract_analysis:
     name: "Shared"
-    uses: aurora-is-near/.github/.github/workflows/contract_analysis.yml@master
+    uses: aurora-is-near/.github/.github/workflows/security_analysis.yml@master
```
mrLSD, line 16: It looks like black box. And CI fails. It seems better to create engine proposals before such changes, as in engine we run all via `cargo make`.

aleksuss (approved):

PR #1000, opened 2025-03-07
fix: incorrect amount of gas provided to promised in ext connector

aleksuss:
## Description

The PR fixes the issue with an incorrect amount of gas attached to promises related to the ETH connector.

No performance changes.

## Testing

Added corresponding tests which test the logic.

## Additional information

This is the first part of the changes related to the external ETH connector. The second part will include modifications related to the modified API of the eth connector.

### DIFF
--- a/engine-tests-connector/src/connector.rs
+++ b/engine-tests-connector/src/connector.rs
@@ -8,7 +8,7 @@ use aurora_engine_types::{
     H256, U256,
 };
 use byte_slice_cast::AsByteSlice;
-use near_sdk::serde_json::json;
+use near_sdk::serde_json::{json, Value};
 use near_sdk::{json_types::U128, serde};
 use near_workspaces::types::NearToken;
 use near_workspaces::AccountId;
@@ -55,7 +55,7 @@ async fn test_aurora_ft_transfer() -> anyhow::Result<()> {
         .deposit(ONE_YOCTO)
         .transact()
         .await?;
-    assert!(res.is_success());
+    assert!(res.is_success(), "{res:#?}");
 
     let balance = contract
         .eth_connector_contract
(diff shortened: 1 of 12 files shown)


engine-tests-connector/src/connector.rs
```diff
@@ -1311,6 +1311,58 @@ async fn test_ft_metadata() -> anyhow::Result<()> {
...
+        .call("storage_balance_of")
+        .args_json(json!({
+            "account_id": "account.near",
+        }))
+        .max_gas()
+        .transact()
+        .await?
+        .json::<Value>()?;
+
+    assert_eq!(res, json!({"available": "0", "total": "0"}));
```
birchmd, line 1342: Please add a comment about why 0 is the expected response when a non-zero amount of tokens were deposited. Something along the lines of "The NEP-141 implementation of ETH intentionally set the storage deposit amount equal to 0 so any non-zero deposit amount is automatically returned to the user, leaving 0 storage balance behind."

engine/src/contract_methods/connector/internal.rs
```diff
@@ -444,6 +444,11 @@ pub fn ft_metadata<I: IO + Copy>(mut io: I) -> Result<(), ContractError> {
     Ok(())
 }
 
+pub fn get_bridge_prover<I: IO + Copy>(io: I) -> Result<(), ContractError> {
+    EthConnectorContract::init(io)?.get_bridge_prover();
```
birchmd, line 448: In the original code this value was returned via `io.return_output`. Perhaps a test of this function should be added as well?

aleksuss: Good catch 👍🏻

birchmd (approved):

mrLSD (approved):

PR #1003, opened 2025-03-18
deps: switch to aurora-evm crate

mrLSD:
## Description

🚀  Switched from **SputnikVM** to new release version of **Aurora EVM**: `aurora-evm v1.0.0` 

➡️ Changed all dependencies and tests

### DIFF
--- a/engine-tests/src/tests/repro.rs
+++ b/engine-tests/src/tests/repro.rs
@@ -26,7 +26,7 @@ fn repro_GdASJ3KESs() {
         block_timestamp: 1_645_717_564_644_206_730,
         input_path: "src/tests/res/input_GdASJ3KESs.hex",
         evm_gas_used: 706_713,
-        near_gas_used: 113,
+        near_gas_used: 115,
     });
 }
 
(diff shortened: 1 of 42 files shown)


aleksuss (approved):

engine-tests/src/tests/repro.rs
```diff
@@ -88,7 +88,7 @@ fn repro_5bEgfRQ() {
         block_timestamp: 1_651_073_772_931_594_646,
         input_path: "src/tests/res/input_5bEgfRQ.hex",
         evm_gas_used: 6_414_105,
-        near_gas_used: 508,
+        near_gas_used: 649,
```
birchmd, line 91: Why has the gas usage gone up so significantly on this test?

birchmd (approved):

engine-tests/src/tests/repro.rs
mrLSD: It's expected behavior. As we're using on Engine version related to Prague. And that particular release is Cancun.
@aleksuss asked me to use previous version, for precompilles release.
But after that I'll update to new version of `aurora-evm`. And I'm expected, that all optimized gas cost for engine will return back

PR #1004, opened 2025-03-24
Feat: move BLS12-381 precompiles execution implementations to engine-sdk

mrLSD:
## Description

➡️  Moved `BLS12-381` precompiles execution implementations to `engine-sdk`
➡️  Added `bls` feature flag to `engine-sdk`
➡️  Bunped version to `1.3.0` for: `engine-sdk, engine-precompiles, engine-transactions`
⚠️ It's related to Etnereum Prague hard fork

### Gas usage

- NEAR gas was decreased, as it's related to the new changes for the Prague hard fork.
- For some tests EVM gas was increased. It's expected behavior related to [EIP-7623](https://eips.ethereum.org/EIPS/eip-7623)


engine/src/engine.rs
```diff
@@ -2762,7 +2762,7 @@ mod tests {
         let mut handler = Noop;
         let actual_result = refund_on_error(io, &env, expected_state, &args, &mut handler).unwrap();
         let expected_result =
-            SubmitResult::new(TransactionStatus::Succeed(Vec::new()), 21344, Vec::new());
+            SubmitResult::new(TransactionStatus::Succeed(Vec::new()), 21860, Vec::new());
```
birchmd, line 2765: Why did this EVM gas amount change?

aleksuss:
Looks like the only half of the work has been done.

engine-tests/src/tests/repro.rs
```diff
@@ -26,7 +26,7 @@ fn repro_GdASJ3KESs() {
         block_timestamp: 1_645_717_564_644_206_730,
         input_path: "src/tests/res/input_GdASJ3KESs.hex",
         evm_gas_used: 706_713,
-        near_gas_used: 130,
+        near_gas_used: 126,
```
birchmd, line 29: Always nice to see Near gas amounts reduced 🚀

birchmd (commented):
The change looks ok, though I am not sure what the motivation is for only moving the standalone implementation of the BLS precompiles to engine-sdk. Could you elaborate on why the code should exist in that crate instead?

engine/src/engine.rs
mrLSD: It's related to EIP-7623 as the gas floor cost was increased

mrLSD:
@aleksuss 
	1.	In my opinion, the initial decision to move the implementation was a mistake and a sign of poor design. Moreover, it is inconsistent because the logic of other precompiles remained unchanged. If we think in such terms, a consistent decision would be to apply this approach to all precompiles.
	2.	I can assume that “half of the work done” refers to having a single entry point for functions, similar to `keccak`. However, in this particular case, my conscious decision was to implement it this way—to not hide the standalone implementation. Since the logic itself is complex, I follow Python’s philosophy: Explicit is better than implicit. You are proposing a Java-like approach—hiding implementation details. I don’t like the idea of having implicit standalone calls within BLS12-381. Yes, the code looks clean, but at a high cost—the implicit behavior of BLS precompiles.

mrLSD:
@birchmd 
Not only the standalone module was moved, but actually all execution functions, making the call occur at a single point. In turn, in the `engine-sdk`, these implementations were placed under the “contract” flag, and an additional “bls” flag was added to restrict access to these functions by default, making the `engine-sdk` more lightweight.

engine-precompiles/Cargo.toml
```diff
@@ -34,8 +33,8 @@ workspace = true
 
 [features]
 default = ["std"]
-std = ["aurora-engine-types/std", "aurora-engine-sdk/std", "blst", "bn/std", "aurora-evm/std", "ripemd/std", "sha2/std", "ethabi/std"]
-contract = ["aurora-engine-sdk/contract"]
+std = ["aurora-engine-types/std", "aurora-engine-sdk/std", "aurora-engine-sdk/bls", "bn/std", "aurora-evm/std", "ripemd/std", "sha2/std", "ethabi/std"]
+contract = ["aurora-engine-sdk/contract", "aurora-engine-sdk/bls"]
```
birchmd, line 37: Why is the `bls` feature needed for `contract`? I thought that feature was only if the standalone bls implementation should be included.

aleksuss: The idea was not to include the heavy, specific `bls` logic in the `aurora-engine-sdk` by default. By providing the corresponding feature only.

mrLSD: Our main idea is granularity. If the feature `contract` in `engine-sdk` is written as follows:
```
contract = "bls"
```
this means that `bls` will be included everywhere `engine-sdk.contract` is used, for example, in `engine`. However, this is redundant since the only place where it is actually used is `engine-precompiles`.

birchmd: Having a feature flag that removes everything related to bls does make more sense for external usage of the sdk crate. I was just worried about including unused code in the wasm binary, but maybe that is prevented by restricting the dependency according to compile target. Perhaps it is fine as-is.

birchmd (approved):

PR #1006, opened 2025-04-03
feat: use omni compatible eth connector

aleksuss:
## Description

The PR applies changes regarding the new API of the eth connector contract.

Tiny changes.

## Testing

Related tests have been modified.

### DIFF
--- a/engine-tests-connector/src/connector.rs
+++ b/engine-tests-connector/src/connector.rs
@@ -1,47 +1,25 @@
 use crate::utils::*;
-use aurora_engine::contract_methods::connector::deposit_event::{
-    DepositedEvent, TokenMessageData, DEPOSITED_EVENT,
-};
-use aurora_engine_types::parameters::connector::{Proof, WithdrawResult};
-use aurora_engine_types::{
-    types::{Address, Fee, NEP141Wei},
-    H256, U256,
-};
-use byte_slice_cast::AsByteSlice;
+use aurora_engine_types::types::NEP141Wei;
 use near_sdk::serde_json::{json, Value};
 use near_sdk::{json_types::U128, serde};
 use near_workspaces::types::NearToken;
-use near_workspaces::AccountId;
+use near_workspaces::{AccountId, Contract};
 use std::str::FromStr;
 
 const ONE_YOCTO: NearToken = NearToken::from_yoctonear(1);
 
-/// Bytes for a NEAR smart contract implementing `ft_on_transfer`
-fn dummy_ft_receiver_bytes() -> Vec<u8> {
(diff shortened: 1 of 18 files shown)


engine/src/contract_methods/connector/mod.rs
```diff
@@ -199,7 +202,10 @@ pub fn finish_deposit<I: IO + Copy, E: Env, H: PromiseHandler>(
     #[cfg(not(feature = "ext-connector"))]
     let result = internal::finish_deposit(io, env, handler)?;
     #[cfg(feature = "ext-connector")]
-    let result = external::finish_deposit(io, env, handler)?;
+    let result = {
```
karim-en, line 205: I am a little bit surprised that we redirected the `finish_deposit`, it was private method

engine-tests-connector/src/connector.rs
```diff
@@ -1237,27 +719,25 @@ async fn test_ft_transfer_user() -> anyhow::Result<()> {
...
         .call(contract.engine_contract.id(), "withdraw")
-        .args_borsh((recipient_addr, withdraw_amount))
-        .max_gas()
+        .args_borsh((*RECIPIENT_ADDRESS, withdraw_amount))
         .deposit(ONE_YOCTO)
+        .max_gas()
         .transact()
         .await?;
-    assert!(res.is_success(), "{res:#?}");
+    assert!(res.is_failure()); // it's ok, the controller doesn't have a correspondent code
```
karim-en, line 738: Maybe we need to mock the controller?

engine/src/contract_methods/connector/mod.rs
```diff
@@ -199,7 +202,10 @@ pub fn finish_deposit<I: IO + Copy, E: Env, H: PromiseHandler>(
...
+        let (_, _, _) = (io, env, handler);
```
mrLSD, line 206: It doesn't look good. I saw that pattern in other places in that PR as well. I understand that you want to be friendly with Clippy.

engine-tests-connector/src/connector.rs
aleksuss: Good suggestion. Thanks.

engine/src/contract_methods/connector/mod.rs
aleksuss: Polished in https://github.com/aurora-is-near/aurora-engine/pull/1006/commits/80740e9dce52a44ce504297f631585a537c3f0ce

mrLSD (approved):

birchmd (approved):

karim-en (approved):

PR #1008, opened 2025-04-05
feat: deploy erc20 token with metadata

aleksuss:
## Description

When we deploy a new ERC-20 token contract using the `deploy_erc20_token`, we cannot provide custom metadata for the token. And we need to call another transaction, `set_erc20_metadata`, to set it. It's inconvenient because the transaction `set_erc20_metadata` could be called by the owner only. The PR **extends**(with saving backward compatibility) the functionality of the `deploy_erc20_token` transaction by adding the possibility of deploying an ERC-20 token with metadata in one transaction. 

No performance changes.

## Testing

The corresponding test was added.


aleksuss:
The silo tests will be fixed after merging #1006

karim-en:
I prefer to retrieve the metadata over cross-call to the token contract, so there is no need to pass the metadata as an argument.

birchmd (approved):
Looks good to me, nice work!

mrLSD (approved):

PR #1009, opened 2025-04-06
feat: set erc-20 fallback address

aleksuss:
## Description

The PR introduces a new transaction that will allow the setting of an ERC-20 fallback address without fixed gas per transaction.

No performance changes.

## Testing

The corresponding test was added.


birchmd (approved):
Good work!

PR #1015, opened 2025-04-23
chore: bump rust version to 1.86.0

aleksuss:
## Description

The PR bumps Rust version and other dependencies. Also, it introduces the usage of the `cargo-near-build` crate for building contracts used in the tests.

NEAR gas consumption slightly decreased.

## Testing

Used existing tests.

## Additional information

The function [`compile`](https://github.com/aurora-is-near/aurora-engine/compare/chore/rust-1-86-0?expand=1#diff-b24f3f01732019f8bafc72a2ee78615fffe26d2ccb6943a17c260a4590bc3181R3) has been reworked using the `cargo-near-build` crate inside. It's important to remember that using the `cargo-near-build` is suitable for contacts which include the dependency `near-sdk-rs`.

### DIFF
--- a/engine-tests-connector/src/rust.rs
+++ b/engine-tests-connector/src/rust.rs
@@ -1,17 +1,21 @@
-use std::path::Path;
-use std::process::Command;
+use std::path::{Path, PathBuf};
 
-pub fn compile<P: AsRef<Path>>(source_path: P) {
-    let output = Command::new("cargo")
-        .current_dir(source_path)
-        .env("RUSTFLAGS", "-C link-arg=-s")
-        .args(["build", "--target", "wasm32-unknown-unknown", "--release"])
-        .output()
-        .unwrap();
+pub fn compile<P: AsRef<Path>>(manifest_path: P) -> PathBuf {
+    let opts = cargo_near_build::BuildOpts {
+        no_locked: true,
+        no_abi: true,
+        no_embed_abi: true,
+        no_doc: true,
+        manifest_path: Some(
+            cargo_near_build::camino::Utf8PathBuf::from_path_buf(
+                manifest_path.as_ref().join("Cargo.toml"),
+            )
+            .unwrap(),
+        ),
+        ..Default::default()
+    };
 
-    assert!(
-        output.status.success(),
-        "{}",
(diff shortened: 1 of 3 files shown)


engine-standalone-tracing/src/types/call_tracer.rs
```diff
@@ -117,6 +117,7 @@ impl CallTracer {
         self.call_stack.last_mut().unwrap().calls.push(frame);
     }
 
+    #[allow(clippy::missing_const_for_fn)]
```
birchmd, line 120: Why allow instead of changing the function to be const?

birchmd (approved):

engine-standalone-tracing/src/types/call_tracer.rs
aleksuss: Yes, it's an interesting case. But I guess it's because we call implicitly the [`deref_mut`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.deref_mut) function, which is not const.

aleksuss: > Funny that clippy has a false positive for this case.

Yeah. It happens from time to time 😄

engine-precompiles/src/hash.rs
```diff
@@ -107,13 +104,10 @@ impl RIPEMD160 {
...
 
 impl Precompile for RIPEMD160 {
-    // TODO: should be enable from `RUst 1.84`
-    // #[allow(clippy::manual_div_ceil)]
     fn required_gas(input: &[u8]) -> Result<EthGas, ExitError> {
         let input_len = u64::try_from(input.len()).map_err(utils::err_usize_conv)?;
         Ok(
-            (input_len + consts::RIPEMD_WORD_LEN - 1) / consts::RIPEMD_WORD_LEN
-                * costs::RIPEMD160_PER_WORD
+            input_len.div_ceil(consts::RIPEMD_WORD_LEN) * costs::RIPEMD160_PER_WORD
```
mrLSD, line 110: Nice catch!

mrLSD (approved):
Nice work! Good catch with `div_cell`. 🚀

PR #1016, opened 2025-04-24
ci: add wasm-opt dependency in scheduled tasks

aleksuss:
## Description

The PR does follow:

- fixes the scheduled and build tasks in the CI by adding the `wasm-opt` dependency
- changes the flow of building the XCC router contract by removing the profiles `mainnet` and `testnet` since it seems redundant
- brings some changes to the `aurora-engine-precompiles` by moving some feature-specific implementations to the `aurora-engine-sdk`


mrLSD (approved):
Nice work! 🚀

birchmd (approved):

PR #1018, opened 2025-05-06
chore: remove feature ext-connector

aleksuss:
## Description

The PR removes the feature: `ext-connector` and all code behind the feature. Besides that, some refactoring was done, removing unused types and errors.

Additionally, the PR adds five attempts to compile EVM contracts in some tests. The motivation is to try to fix the issues like [this](https://github.com/aurora-is-near/aurora-engine/actions/runs/14901654032/job/41854719664#step:11:1428) or like [this](https://github.com/aurora-is-near/aurora-engine/actions/runs/14901654032/job/41854720535#step:11:1428), which happen periodically in the tests that use those solidity contracts. The issue could be related to the problem of getting the solidity contract source code and other required dependencies.


engine/src/contract_methods/connector.rs
```diff
@@ -0,0 +1,629 @@
...
+        attached_balance: ONE_YOCTO,
+        attached_gas: calculate_attached_gas(env),
+    };
+    let promise_id = unsafe { io.promise_create_call(&promise_arg) };
+    io.promise_return(promise_id);
+
+    Ok(())
+}
+
+pub fn ft_transfer_call<I: IO + Env + Copy, E: Env, H: PromiseHandler>(
```
birchmd, line 218: I wonder if there is a way to generalize the NEP-141 functions that are now simply forwarding to the external connector. They all have the same structure of reading the args and constructing a promise. This is not a blocking comment, just a suggestion that might reduce some minor repetition in the code.

birchmd (approved):
Nice job! It's always good to see net lines of code removed and gas costs decrease.

engine/src/contract_methods/connector.rs
aleksuss: Yeah. It's a good point.

aleksuss: Done in ec547c8

mrLSD (approved):
Awesome work! 🚀

PR #1019, opened 2025-05-14
chore: remove features mainnet and testnet

aleksuss:
## Description

The PR removes the features `mainnet` and `testnet`. The features were used to distinguish the custodian address used in the previous version of the Rainbowbridge. After switching to the OMNI bridge, it's not so critical anymore. However, I decided to leave the check for the custodian address from MAINNET since unfinished payments still exist.


birchmd (approved):

mrLSD (approved):
Great work on that 🎉! Overall, excellent refactoring and code cleanup in the recent PRs, which not only made the code more concise, readable, and understandable, but also delivered a noticeable improvement in NEAR gas consumption! 🚀

PR #1024, opened 2025-07-08
feat: improve exit_to_near precompile security

aleksuss:
## Description

The PR improves the security of the `exit_to_near` precompile by introducing a structure for serializing arguments for the `ft_transfer_call` transaction.

## Summary by CodeRabbit

* **New Features**
  * Improved JSON serialization for token transfer calls, enhancing safety and reliability.
  * Added unit tests to ensure correct JSON construction and prevent malformed data.

* **Refactor**
  * Simplified contract import logic by removing dependency on an external testnet client.
  * Reduced the default root balance for engine contract deployment from 10,000 to 5,000 NEAR tokens.

* **Chores**
  * Updated dependencies and features for better serialization support.

### DIFF
--- a/engine-precompiles/src/native.rs
+++ b/engine-precompiles/src/native.rs
@@ -1296,4 +1296,13 @@ mod tests {
         );
         assert_eq!(args, expected);
     }
+
+    #[test]
+    #[should_panic(expected = "ERR_INVALID_AMOUNT")]
+    fn test_ft_transfer_call_args_u256() {
+        let receiver_id = "test.near".parse().unwrap();
+        let amount = U256::from(u128::MAX) + 1;
+        let msg = "some message";
+        let _ = super::ft_transfer_call_args(&receiver_id, amount, msg).unwrap();
+    }
 }
--- a/engine-precompiles/src/native.rs
+++ b/engine-precompiles/src/native.rs
@@ -787,11 +787,11 @@ fn parse_input(input: &[u8]) -> Result<&[u8], ExitError> {
     Ok(&input[1..])
 }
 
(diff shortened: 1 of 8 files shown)


engine-precompiles/src/native.rs
```diff
@@ -791,6 +787,26 @@ fn parse_input(input: &[u8]) -> Result<&[u8], ExitError> {
...
+}
+
+fn ft_transfer_call_args(
+    receiver_id: &AccountId,
+    amount: U256,
+    msg: &str,
+) -> Result<String, ExitError> {
+    serde_json::to_string(&FtTransferCallArgs {
+        receiver_id: receiver_id.clone(),
+        amount: format!("{amount}"),
```
karim-en, line 808: The amount is supposed to be U128, not U256, so we have two options here:
- Keep it as it is; in this case, it will panic after the promise call
- Try to cast in the precompile earlier, so it can be reverted.

aleksuss: Good point 👍🏻

engine-precompiles/src/native.rs
```diff
@@ -791,6 +787,30 @@ fn parse_input(input: &[u8]) -> Result<&[u8], ExitError> {
     Ok(&input[1..])
 }
 
+#[derive(serde::Serialize, serde::Deserialize)]
+struct FtTransferCallArgs {
```
birchmd: Performance nit: we could avoid some cloning in the `ft_transfer_call_args` implementation by holding references to `receiver_id` and `msg` in this struct and not deriving `Deserialize`.

birchmd (approved):

PR #1033, opened 2025-08-12
Run aurora contract in near-vm-runner

vlad9486:
## Description

Run aurora contract as a WASM binary in near-vm-runner instead of the native code.

There is defined a trait `AbstractVMRunner`. The instance that implementing the trait must be provided to `sync::execute_transaction` function of the standalone runner.

There is also available the trait implementation for tests.

The tracing must be performed inside the WASM code. The crate `aurora-engine-traced` adds tracing contract calls that works inside the WASM execution. This crate produces the wasm binary. It can be linked with different versions of the aurora contract to bring tracing in older contracts.

2.6 - 3 times slowdown of contract execution in vm runner compared to native code execution.

## Testing

Few existing tests still fails.


engine-standalone-storage/src/engine_state.rs
```diff
@@ -49,8 +51,8 @@ impl<'db, 'input, 'output> EngineStateAccess<'db, 'input, 'output> {
         input: &'input [u8],
         bound_block_height: u64,
         bound_tx_position: u16,
-        transaction_diff: &'output RefCell<Diff>,
-        output: &'output Cell<Vec<u8>>,
+        transaction_diff: &'output Mutex<Diff>,
+        output: &'output OnceLock<Vec<u8>>,
```
birchmd: I don't know if we should use `OnceLock` here. I think it is probably true that `output` is only set once in each transaction, but I don't know if we can really guarantee it. I think we should keep the previous behaviour of only keeping the last `output`.

engine-standalone-storage/src/sync/mod.rs
```diff
@@ -291,10 +299,15 @@ pub fn parse_transaction_kind(
 /// Note: this function does not automatically commit transaction messages to the storage.
 /// If you want the transaction diff committed then you must call the `commit` method on
 /// the outcome of this function.
-pub fn consume_message<M: ModExpAlgorithm + 'static>(
+pub fn consume_message<M: ModExpAlgorithm + 'static, R>(
```
birchmd: It feels somewhat convoluted to me that we are changing the standalone running to work in this way. The original point of the standalone runner is that it provided us a pathway to executing Engine transactions without Wasm.

Now we want to execute transactions in an alternate Wasm runner rather than native. But then I think the standalone runner interface no longer makes sense. We are consuming structured data only to convert it into strings and bytes to pass into the Wasm runner. It seems like it would be more direct to have an alternate interface for Borealis Engine where it can pass Near `FunctionCall` actions as they are instead of processing them into structured data (`Message`) only to then turn it back into the unstructured form of the original `FunctionCall`.

vlad9486: @birchmd Am I understanding correctly, that we can stop parsing a near block somewhere earlier, maybe [here](https://github.com/aurora-is-near/borealis-engine-lib/blob/feat/standalone-storage-with-dynamic-contract/engine/src/sync.rs#L126) and pass the given function call into near-vm-runner omitting [this](https://github.com/aurora-is-near/aurora-engine/blob/feat/wasm-tracing/engine-standalone-storage/src/sync/mod.rs#L37)?

This would simplify our code, but I suspect it would put more work on the near-vm-runner. More computation would happen in WASM, which could hurt performance. But I must be missing something.

vlad9486: I already see that it doesn't impact performance and doesn't burden the wasm with additional job. This is pure simplification on the native code side.

birchmd: Yes, you have the right idea of what I am suggesting. I do not see why it would put more work on the Wasm side. In the flow proposed in this PR it is already true that the structured data is not being used for anything except reproducing the method name that would have already been present in the original action (there is some complication around whether tracing is enabled, but I think that could be handled in a simpler way too). So the Wasm is already pulling the input from `io` and parsing it into whatever structured arguments it needs.

engine-standalone-storage/src/runner.rs
```diff
@@ -0,0 +1,19 @@
...
+pub trait AbstractContractRunner {
+    type Error;
+
+    fn call_contract<E, I>(
+        &self,
+        method: &str,
+        promise_data: Vec<Option<Vec<u8>>>,
+        env: &E,
+        io: I,
+        more_ctx: &mut [u8; 32],
```
birchmd: This variable is for keeping track of the account balance and total storage usage; something we don't have access to from the receipts. But I wonder if we need to keep track of it. By design, Borealis Engine only executes transactions that were successful on Near. So regardless of what the real account balance and storage usage are, we know they allowed the transaction to finish successfully. Maybe placeholder values that always make the execution succeed can be passed in where needed in the Near VM runner instead of the real values.

engine-standalone-storage/src/sync/mod.rs
```diff
@@ -477,217 +290,445 @@ fn compute_random_seed(action_hash: &H256, block_random_value: &H256) -> H256 {
...
+                Some(data) => {
+                    let mut slice = data.as_slice();
+                    let address =
+                        Address::deserialize(&mut slice).map_err(ExecutionError::Deserialize)?;
+                    TransactionExecutionResult::DeployErc20(address)
+                }
+                None => {
+                    //
+                    // TransactionExecutionResult::Promise(promise_args)
+                    panic!("cannot handle case where `deploy_erc20_token` returns promise")
```
birchmd: This case probably needs to be handled somehow. The Borealis Engine should not panic for any transaction.

engine-standalone-storage/src/sync/mod.rs
```diff
@@ -477,217 +290,445 @@ fn compute_random_seed(action_hash: &H256, block_random_value: &H256) -> H256 {
...
+                    promise_results,
+                    env,
+                    io,
+                    ctx,
+                    None,
+                )
+                .map_err(ExecutionError::from_vm_err)?;
 
-            maybe_result.map(|submit_result| TransactionExecutionResult::Submit(Ok(submit_result)))
+            // maybe_result.map(|submit_result| TransactionExecutionResult::Submit(Ok(submit_result)))
```
birchmd: I think this result was returned to aid the explorer in displaying withdrawals more accurately. It would be nice to continue to support it.

engine-standalone-storage/src/sync/mod.rs
```diff
@@ -477,217 +290,445 @@ fn compute_random_seed(action_hash: &H256, block_random_value: &H256) -> H256 {
...
+            None
         }
-        TransactionKind::SetConnectorData(_) => None,
-        TransactionKind::NewConnector(_) => None,
-        TransactionKind::NewEngine(_) => {
-            contract_methods::admin::new(io, env)?;
+        TransactionKindTag::SetConnectorData => None,
```
birchmd: I know the Engine no longer contains the eth-connector, so this transaction kind is now entirely a no-op. But I thought part of the point of this work is to allow executing past versions of Aurora Engine as Wasm modules during indexing, as this ensures the off-chain results match what happened on-chain. Should the processing here somehow reflect this? For example, always run `call_contract` regardless of the transaction kind and let the Wasm module decide if the method is a no-op or not.

engine-standalone-storage/src/sync/mod.rs
```diff
@@ -477,217 +290,445 @@ fn compute_random_seed(action_hash: &H256, block_random_value: &H256) -> H256 {
...
+            runner
+                .call_contract("set_erc20_metadata", promise_results, env, io, ctx, None)
+                .map_err(ExecutionError::from_vm_err)?;
 
             None
         }
-        TransactionKind::SetFixedGas(args) => {
-            silo::set_fixed_gas(&mut io, args.fixed_gas);
+        TransactionKindTag::SetFixedGas => {
+            let args = silo_params::FixedGasArgs::try_from_slice(&transaction.args).unwrap();
```
birchmd: I'm not sure why this deserialization into `args` only to reserialize into `input` is needed. Couldn't the serialized bytes have been passed directly via `io` as in the case of the other contract methods? This comment applies to the other silo methods as well.

engine-standalone-storage/src/sync/types.rs
```diff
@@ -77,434 +75,125 @@ impl TransactionMessage {
...
-    SetWhitelistStatus(silo::WhitelistStatusArgs),
-    SetWhitelistsStatuses(Vec<silo::WhitelistStatusArgs>),
-    /// Callback which mirrors existed ERC-20 contract deployed on the main contract.
-    MirrorErc20TokenCallback(connector::MirrorErc20TokenArgs),
-    /// Sentinel kind for cases where a NEAR receipt caused a
-    /// change in Aurora state, but we failed to parse the Action.
-    Unknown,
+#[borsh(crate = "aurora_engine_types::borsh")]
+pub struct TransactionKind {
+    pub(super) method_name: TransactionKindTag,
```
birchmd: Seems like we would save a lot of boilerplate code in the execution logic above if we just left this as a string. Only `submit` can have its method name altered before the contract call (to do tracing) and this could be handled as a special case.

engine-standalone-storage/src/sync/types.rs
```diff
@@ -908,254 +482,39 @@ impl<'a> TryFrom<BorshableTransactionMessage<'a>> for TransactionMessage {
...
-                Self::SetWhitelistsStatuses(Cow::Borrowed(x))
-            }
+        BorshableTransactionKind {
+            method_name: t.method_name,
+            args: Cow::Borrowed(&t.args),
+            promise_result: match &t.promise_result {
+                Some(PromiseResult::Successful(v)) => Some(Some(Cow::Borrowed(&*v))),
+                Some(PromiseResult::Failed) => Some(None),
+                Some(PromiseResult::NotReady) => {
+                    debug_assert!(false, "should never happen"); // is that correct?
```
birchmd: Indeed `NotReady` should never come up in the current implementation of Near. This was a placeholder for an `or` promise combinator that allowed racing two promises, but that functionality was never implemented.

engine-standalone-storage/src/sync/mod.rs
vlad9486: Yes, sure. I will need to override this method in the contract, because Wasm code doesn't return anything here. It was working before, because we use Rust  library directly and got the refund SubmitResult from there.

I need to do it properly, it was a quick plug.

engine-standalone-storage/src/sync/types.rs
vlad9486: I keep it because it is just 1 byte versus string that allocated in the heap and maybe will be cloned somewhere.

engine/src/lib.rs
```diff
@@ -23,25 +23,52 @@ pub mod xcc;
...
+        }
+
+        unsafe fn dealloc(&self, _ptr: *mut u8, _layout: Layout) {
+            unreachable!("should never be called, since alloc never returns non-null");
+        }
+    }
+
+    #[global_allocator]
+    static ALLOC: Allocator = Allocator;
+}
```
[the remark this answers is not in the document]
vlad9486: cargo-udeps actually build it with the not supported combination of arch and features

engine-tests/src/tests/modexp.rs
```diff
@@ -119,12 +122,18 @@ fn bench_modexp_standalone() {
...
 
         let start = std::time::Instant::now();
+        // don't care about near gas in standalone runner
+        standalone.env.prepaid_gas = NearGas::new(u64::MAX);
         standalone
             .submit_transaction(&signer.secret_key, bench_tx)
             .unwrap();
-        let duration = start.elapsed().as_millis();
+        let duration = start.elapsed();
+        let limit = Duration::from_secs(12);
```
birchmd: Does it really take 12s to run now? We have to be a little careful with this. The reason we had this performance check was as a regression test for a previous issue where it was possible to DOS our RPC by submitting `eth_call` requests that took a long time to run. If it is possible to have calls that will take 12s to run then maybe on the Borealis Engine side of things we need a timeout to kill the task early and return an error.

engine-tests/src/utils/mod.rs
```diff
@@ -327,14 +347,36 @@ impl AuroraRunner {
...
+
+    pub fn submit_with_signer_profiled_with_call_trace<F: FnOnce(U256) -> TransactionLegacy>(
+        &mut self,
+        signer: &mut Signer,
+        make_tx: F,
+    ) -> Result<SubmitResultExt, EngineError> {
+        let nonce = signer.use_nonce();
+        let tx = make_tx(nonce.into());
+        let signed_tx = sign_transaction(tx, Some(self.chain_id), &signer.secret_key);
+        // TODO: enable call trace
```
birchmd: Looks like this `TODO` is done now.

PR #1110, opened 2026-03-27
fix(ci): try pr author instead of actor for dependabot skip

rcny:
Сheck `pull_request.user.login` for the skip condition.


rcny:
Nope, `actor` is enough.

PR #1112, opened 2026-03-29
Fix: EIP-7702 is_delegated check for submit

mrLSD:
## Summary by CodeRabbit

* **Bug Fixes**
  * Refined sender validation so delegated senders with deployed code are accepted where appropriate, reducing false rejections on eligible transactions.

* **Tests**
  * Added comprehensive delegation test coverage: installation, delegated and sponsored transactions, nonce/balance/state assertions, and revocation flows.

* **Tools**
  * Added test helpers to produce signed authorizations and runtime contract samples for exercising delegation scenarios.

### DIFF
--- a/engine-tests/src/tests/transaction.rs
+++ b/engine-tests/src/tests/transaction.rs
@@ -608,7 +608,7 @@ fn test_eip_7702_authority_self_revoke() {
     // Signer untouched
     assert_eq!(runner.get_nonce(signer_address), 3.into());
 
-    // gas_preice * gas_used
+    // gas_price * gas_used
     let gas_fee = Wei::new_u64(0x0A * result_2.gas_used);
     assert_eq!(
         runner.get_balance(authority_address),
--- a/engine-tests/src/tests/transaction.rs
+++ b/engine-tests/src/tests/transaction.rs
@@ -556,7 +556,8 @@ fn example_signer() -> utils::Signer {
 /// Build contract code: PUSH20 <authority> | EXTCODESIZE | PUSH1 0 | SSTORE | STOP
 #[must_use]
 fn sample_code_for_contract_eip7702(authority_address: Address) -> Vec<u8> {
-    let mut code = vec![0x73]; // PUSH20
+    let mut code = Vec::with_capacity(26);
+    code.push(0x73); // PUSH20
     code.extend_from_slice(authority_address.as_bytes());
     code.push(0x3B); // EXTCODESIZE
     code.push(0x60); // PUSH1
--- a/engine-tests/src/tests/transaction.rs
+++ b/engine-tests/src/tests/transaction.rs
@@ -250,6 +250,7 @@ fn test_eip_7702_success() {
 /// Step 1: EIP-7702 tx installs delegation on authority address
 /// Step 2: EIP-1559 tx sent FROM that authority address succeeds
 /// Step 3 — Sponsored tx: signer calls authority, triggering delegated code in authority's context
+#[allow(clippy::too_many_lines, clippy::cognitive_complexity)]
 #[test]
 fn test_eip_7702_delegated_sender_can_transact() {
     let mut runner = utils::deploy_runner();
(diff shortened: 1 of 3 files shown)


engine/src/engine.rs
```diff
@@ -1086,8 +1086,15 @@ pub fn submit_with_alt_modexp<
 
     let mut engine: Engine<_, _, M> =
         Engine::new_with_state(state, sender, current_account_id, io, env);
-    // EIP-3607
-    if !engine.code(sender.raw()).is_empty() {
+
+    let sender_code = engine.code(sender.raw());
+    // EIP-7702 - check if it's delegated designation. If it's a delegation designation, then,
+    // even if `caller_code` is non-empty, the transaction should be executed.
+    let is_delegaated = Authorization::is_delegated(&sender_code);
```
aleksuss:
```suggestion
    let is_delegated = Authorization::is_delegated(&sender_code);
```

aleksuss:
I guess it makes sense to add end-to-end tests to cover some commonly used use cases. Let's do it.

engine-tests/src/tests/transaction.rs
```diff
@@ -312,25 +553,54 @@ fn example_signer() -> utils::Signer {
     }
 }
 
+/// Build contract code: PUSH20 <authority> | EXTCODESIZE | PUSH1 0 | SSTORE | STOP
+#[must_use]
+fn sample_code_for_contract_eip7702(authority_address: Address) -> Vec<u8> {
+    let mut code = vec![0x73]; // PUSH20
```
birchmd: Nit: since the size is known in advance (26 bytes) the vector could be initialized using `with_capacity` to avoid extra allocations.

engine-tests/src/tests/transaction.rs
```diff
@@ -239,14 +237,257 @@ fn test_eip_7702_success() {
...
+        contract_address,
+        CONTRACT_BALANCE,
+        CONTRACT_NONCE.into(),
+        contract_code.clone(),
+    );
+
+    // ══════════════════════════════════════════════════════════════
+    // Step 1: EIP-7702 tx — install delegation on authority
+    // ══════════════════════════════════════════════════════════════
+    let mut transaction = eip7702_transaction(0);
```
birchmd, line 290: I'm confused about passing `0` here instead of `runner.chain_id`. Shouldn't the authorization chain id match the the chain id for this instance of the EVM?

mrLSD: Yes, it's valid, because according to https://eips.ethereum.org/EIPS/eip-7702:
> 1. Verify the chain ID is 0 or the ID of the current chain.

engine-tests/src/tests/transaction.rs
```diff
@@ -239,14 +237,257 @@ fn test_eip_7702_success() {
...
+        INITIAL_BALANCE - transfer_value
+    );
+    // As sponsored transaction executed - storage slot now changed
+    assert_eq!(
+        runner.get_storage(authority_address, H256::zero()),
+        H256::from(H160::from_low_u64_be(23))
+    );
+
+    // ══════════════════════════════════════════════════════════════
+    // Step 4: revoke authority delegation
```
birchmd, line 435: Nit: step 4 is missing from the summary at the top of the test

birchmd (approved):
Nice job! Thanks for adding the tests.

mrLSD:
@coderabbitai full review

engine-tests/src/tests/transaction.rs
```diff
@@ -239,14 +237,385 @@ fn test_eip_7702_success() {
...
+    // Nonce: 1 → 2 (sender increment) → 3 (auth processing increment)
+    assert_eq!(
+        runner.get_nonce(authority_address),
+        3.into(),
+        "authority nonce: 1 +1 (sender) +1 (auth) = 3"
+    );
+    // Signer untouched
+    assert_eq!(runner.get_nonce(signer_address), 3.into());
+
+    // gas_preice * gas_used
```
aleksuss:
```suggestion
    // gas_price * gas_used
```

aleksuss (approved):

PR #1137, opened 2026-04-21
feat: improve EIP-7702 transaction processing and extend test coverage

mrLSD:
## Summary by CodeRabbit

- **Improvements**
  - Improved EIP-7702 authorization signing-hash generation and validation across boundary-value chain IDs and nonces.
  - Expanded transaction handling coverage for delegation, sponsored execution, revocation, re-delegation, duplicate authorities, and invalid authorization scenarios.
  - Improved failure diagnostics by reporting more accurate caller locations in test failures.

- **Documentation**
  - Expanded transaction documentation covering chain IDs, fees, access lists, and authorization lists.


engine-transactions/src/lib.rs
```diff
@@ -93,8 +93,8 @@ pub struct NormalizedEthTransaction {
     pub value: Wei,
     pub data: Vec<u8>,
     pub access_list: Vec<AccessTuple>,
-    // Contains additional information - `chain_id` for each authorization item
-    pub authorization_list: Vec<Authorization>,
+    // Length of `authorization_list` for EIP-7702. Used for gas calculations.
+    pub authorization_list_len: usize,
```
birchmd, line 97: What is the purpose of this change? It does not appear to be related to the new logic of marking `0x0` as an invalid auth address. As far as I can tell the effect it has on the code is to increase some gas costs because now the transaction needs to be cloned to have access to both the authorization list itself and its length.

mrLSD: It solves issue related to draining relayer account NEAR balance.
Current scenario:
- we're validating EIP-7702 `authorization_list` in `NormalizedTransaction`. It's have computing. As user can send list of 1000 items. For each we calling `ecrecovery`.
Vector of attack: in submit user bypass with fail: incorrect chain_id, incorrect nonce, incorrect balance.
In one transaction it's possible to spent 300TGas; And transaction will fail. And relayer will pay for that.

mrLSD: ➡️  The main point is: we shouldn't validate `authorization_list` in `NormalizedTransaction`.

birchmd: Moving the validation out of `NormalizedTransaction` does not prevent this kind of attack; it only changes where the OOG error will happen (now later in the transaction processing flow instead of while parsing the transaction).

We can keep the validation logic in the construction of `NormalizedTransaction`, but also include an early exit if the length of the list is too long.

Relayers can also protect themselves from this attack by checking the length of authorization lists before sending them to Near.

engine-transactions/src/eip_7702.rs
```diff
@@ -207,6 +241,11 @@ impl SignedTransaction7702 {
                 Address::default()
             };
 
+            // 0x0 (Zero address) is Ethereum system address. So we mark it as `valid = false`.
+            if auth_address.raw().is_zero() {
+                is_valid = false;
```
aleksuss: I don't get this. As I remember, we should [reset](https://github.com/aurora-is-near/aurora-evm/blob/master/evm/src/executor/stack/executor.rs#L1017) the code in a case of zero auth_address, but if an entry is invalid, we [skip](https://github.com/aurora-is-near/aurora-evm/blob/master/evm/src/executor/stack/executor.rs#L987) it.

engine-transactions/src/eip_7702.rs
```diff
@@ -242,7 +179,7 @@ impl Decodable for SignedTransaction7702 {
         let value = Wei::new(rlp.val_at(6)?);
         let data = rlp.val_at(7)?;
         let access_list = rlp.list_at(8)?;
-        let authorization_list = rlp.list_at(9)?;
+        let authorization_list: Vec<AuthorizationTuple> = rlp.list_at(9)?;
```
aleksuss:
```suggestion
        let authorization_list = rlp.list_at(9)?;
```

engine-transactions/src/eip_7702.rs
```diff
@@ -12,8 +11,10 @@ use crate::eip_2930::AccessTuple;
 /// Type indicator (per EIP-7702)
 pub const TYPE_BYTE: u8 = 0x04;
 
-// EIP-7702 `MAGIC` number
+/// EIP-7702 `MAGIC` number
 pub const MAGIC: u8 = 0x5;
+/// Limitation for authorization list length. It's strictly related to NEAR gas limits.
+pub const AUTHORIZATION_LIST_LENGTH: usize = 100;
```
aleksuss: Looks like it's not needed anymore.

mrLSD: Done

aleksuss (approved):

engine-transactions/src/lib.rs
```diff
@@ -93,8 +94,8 @@ pub struct NormalizedEthTransaction {
     pub value: Wei,
     pub data: Vec<u8>,
     pub access_list: Vec<AccessTuple>,
-    // Contains additional information - `chain_id` for each authorization item
-    pub authorization_list: Vec<Authorization>,
+    /// EIP-7702 authorization list. If `None`, then the transaction is not EIP-7702.
+    pub authorization_list: Option<Vec<AuthorizationTuple>>,
```
birchmd: I do not think this change makes sense. If the `Vec` was empty then the transaction was not EIP-7702 since that type of transaction requires a non-empty authorization list. Per [eip7702.io](https://eip7702.io/)

> The transaction is considered invalid if the length of authorization_list is zero.

Therefore wrapping in `Option` is redundant.

Using an `Option` also does not save any allocation as the PR description suggests because Rust does not allocate for empty `Vec`s. Per the [Rust standard library documentation](https://doc.rust-lang.org/std/vec/struct.Vec.html)

> if you construct a Vec with capacity 0 via [Vec::new](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.new), [vec![]](https://doc.rust-lang.org/std/macro.vec.html), [Vec::with_capacity(0)](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity), or by calling [shrink_to_fit](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.shrink_to_fit) on an empty Vec, it will not allocate memory.

This change also makes the diff much messier because it causes the `authorization_list` function to be moved to the `NormalizedEthTransaction` type unnecissarily.

Please revert this change and move the function back.

engine-transactions/src/lib.rs
```diff
@@ -249,6 +251,104 @@ impl NormalizedEthTransaction {
...
+            };
+
+            // `ecrecover` returns 0x0 as the convention for failed recovery (yellow paper,
+            // precompile 0x01). Treating it as a valid authority would let degenerate
+            // signatures install delegation on address(0), breaking the EVM invariant
+            // that the zero address holds no code and corrupting state assumptions
+            // downstream. Matches OpenZeppelin ECDSA / go-ethereum / reth behavior.
+            if authority.is_zero() {
+                is_valid = false;
+            }
```
birchmd: I'm not convinced this comment is correct. It is true the the `ecrecover` precompile returns `0x0` for failed recovery, but this function is not using the precompile. It is using [our wrapper over Near's host function](https://github.com/aurora-is-near/aurora-engine/blob/62793dd6bca2d99c3a4911e7b9931d1cc169b69f/engine-sdk/src/lib.rs#L95) which is will [return an error for failed recovery](https://github.com/aurora-is-near/aurora-engine/blob/62793dd6bca2d99c3a4911e7b9931d1cc169b69f/engine-sdk/src/lib.rs#L120) and therefore this would have been caught by the `unwrap_or_else` closure above.

Therefore, I think this `if` statement is actually unnecessary. Please remove it and add a test showing that a delegation to the zero address is still marked as invalid.

engine-tests/src/tests/alt_bn256_precompiles.rs
```diff
@@ -105,7 +105,7 @@ fn test_alt_bn128_add() {
         &Bn256Add::<Istanbul>::new(),
         Bn256Add::<Istanbul>::ADDRESS,
         include_str!("res/alt_bn_128/bn256_add.json"),
-        3590, // 3.590 TGas
+        3700, // 3.700 TGas
```
birchmd: Why have the gas costs increased?

birchmd (changes requested):
This PR cannot be accepted as is because it is refactoring the code to be more complex without adding any new functionality. Refactoring and improving the EIP-7702 tests is a useful thing to do, but that is the only thing this PR should do. I do not think any production code needs to be changed.

engine-transactions/src/lib.rs
mrLSD: @aleksuss  and I decided to remove the `auth_list` length check - it should be handled by the `relayer`. That’s why the previous implementation was changed.

However, using `Option<...>` clearly signals the transaction type. Outside of `EIP-7702`, this field doesn’t exist, and keeping it as just `Vec<...>` is poor design, as it doesn’t align with other transaction fields. The same issue applies to `access_list` - it’s bad design to include an empty array in `Legacy` transactions when the field shouldn’t exist there at all.

Therefore, I believe @aleksuss ’s solution is far more thoughtful than the original messy and inconsistent transaction field design.

aleksuss:
@birchmd thank you for your review. But to be honest, I'm wondering about your concern. As for me there is nothing critical, and rather a tiny omptimisation has been added. The changes move preparing (converting and signatures verification) the `authorization_list` to be ready for EVM after main checks (basically, like the `access_list` converted), that could interrupt transaction execution earlier, if let's say a chain id was set incorrectly. As for `Option` for `authorisation_list`. The reason of adding `Option` is because the `NormalizedTransaction` doesn't have a transaction type, and if the `authorisation_list != None` it means that transaction is 7702. As for empty array for `authorization_list`, I believe that's technically possible to create 7702 transaction with empty `authorisation_list` programatically. Also I don't insist that the changes must be added. We agreed that all validations should happen on the infra side. But to summarise, I guess that the changes bring some kind of optimisation, at least in the business logic of processing transactions (but at the same time, increasing gas consuming indicates the opposite, probably we need to find out why, maybe additional checks).

engine-tests/src/tests/alt_bn256_precompiles.rs
aleksuss: Good point. It would be good to find out why. cc @mrLSD

birchmd:
> possible to create 7702 transaction with empty `authorisation_list` programatically

Yes, this is correct. However it must be rejected as invalid by the engine. I view the `NormalizedEthTransaction` as a validated data type (for example the signature has been converted into the signer's address) so if it was constructed successfully then the underlying typed transaction had the right form. This is why the code was safe before; though to your point it was less intuitive. Perhaps the most expressive way to describe the `authorization_list` field would be with a type like `Option<NonEmptyVec>`. The compiler could optimize such a type to a normal `Vec` similarly to what it does for `Option<NonZeroU64>`. Unfortunately I do not see `NonEmptyVec` in the standard library currently.

> a tiny omptimisation has been added ... move preparing (converting and signatures verification) the `authorization_list` to be ready for EVM after main checks

Ok, I understand the motivation now. It makes sense to do the cheaper checks first because in the case those checks fail then less gas will be spent overall. However, I think it is more important to optimize for the happy path than the sad paths. The relayer can do as many pre-flight checks as it wants upon receiving the transaction to filter out invalid transactions before they go on-chain, but a valid transaction must go on-chain every time. Therefore the most gas is saved when the happy path is as streamlined as possible. This is why I am concerned by the gas cost increases of this change: making the happy path worse in exchange for making a sad path better is not a good trade off.

engine-tests/src/tests/alt_bn256_precompiles.rs
mrLSD: @birchmd 
After investigation, the answer became clear: the gas cost increased due to the WASM binary size increasing by 1 KB.

This was caused by the following check:

```rust
if authority.is_zero() {
    is_valid = false;
}
```

However, I recommend keeping this check as a defensive measure. In earlier versions of `libsecp256k1`, an error was returned as 0x0. In other words, based on historical data, `ecrecover -> 0x0` is a valid result for more than one case.

The change in gas consumption is negligible, while this check provides an additional layer of protection.

birchmd: The check still looks redundant to me based on my understanding of the Near host function. Please add a test showing the case where this check is needed. If one cannot be constructed then adding a test showing it is not needed (i.e. attempted delegation to 0x0 is still marked as invalid) will still provide the layer of protection against future changes to `ecrecover` without modifying any production code.

aleksuss:
@coderabbit full review

mrLSD:
@aleksuss good refactoring, thanks! 🚀

birchmd (approved):
Looks good to me! Thanks for getting this cleaned up.

PR #1156, opened 2026-05-29
feat: bump rust to 1.93 with near-dependencies updates

mrLSD:
## Summary

Bump the toolchain and `nearcore` baseline: **Rust `1.86` → `1.93`** and  **`near-vm-runner` / `near-primitives` / `near-crypto` / `near-parameters` / `near-primitives-core` `0.34` → `0.36.0`**. The change is intentionally scoped to the **minimum** dependency churn required for compatibility and unblocks the CI failures related to dependency freshness and `cargo audit`.

## Motivation

The pinned Rust `1.86` / `near-core 0.34` stack had become a recurring source of CI friction: a `cargo audit` vulnerability (`time` DoS), stale `nearcore` dependencies, and divergence from the currently supported `nearcore` baseline. This PR modernizes that baseline while deliberately avoiding unrelated dependency churn and behavioral changes.

### Toolchain & dependencies
- Rust `1.86.0` → **`1.93.0`** (`rust-toolchain`).
- `near-*` runtime crates `0.34` → **`0.36.0`** (adapted the `External` trait impl in the test harness to the new `near-vm-runner` API).
- **`time` `0.3.41` → `0.3.47`**, which closes **RUSTSEC-2026-0009** (stack-exhaustion DoS) — this is the fix that unblocks the `cargo audit` CI job.
- Dependency updates kept to the minimum required set.

## Gas & Binary Size Impact

- **Net gas: ~baseline.** Small per-test codegen drift (low single-digit %, in both directions); no systemic regression.
- **Engine WASM binary ~100 KB smaller.**

## Result

- ✅ Aligns Rust / `nearcore` with the currently supported baseline (`1.93` / `0.36.0`).
- ✅ Resolves `cargo audit` (RUSTSEC-2026-0009) and dependency-freshness CI failures.
- ✅ Avoids the ~2× gas regression; keeps net gas at ~baseline.
- ✅ Reduces engine WASM size by ~100 KB.
- ✅ `cargo make test` / `udeps` / `audit` green.

A focused, controlled maintenance update that modernizes the toolchain/dependency
baseline while preserving Aurora Engine's build and gas characteristics.

----

## Summary by CodeRabbit

* **New Features**
  * Standardized build target for Rust/WASM and enabled bulk-memory for wasm-opt.

* **Dependencies**
  * Toolchain and wide dependency updates; Binaryen version bumped.

* **Bug Fixes**
  * Harmonized gas-limit checks and input-length validations across precompiles (behavior preserved).

* **Tests**
  * Updated gas/assertion bounds and clarified ignore reasons for benchmarks/integration tests.

* **Chores**
  * Added a workspace lint rule and switched some test artifacts to on-demand download.


mrLSD:
Also needed for #1137 CI pass

aleksuss:
I bumped NEAR dependencies to the latest ones, so the PR is already ready for final review. 
cc @birchmd @mrLSD

aleksuss (approved):

mrLSD:
@aleksuss why do you enabled ` --enable-bulk-memory ` for wasm-opt?

aleksuss:
@mrLSD because without this parameter, the wasm-opt fails with errrors like:
```
[wasm-validator error in function 3] unexpected false: memory.copy operations require bulk memory operations [--enable-bulk-memory-opt], on 
(memory.copy
 (local.get $4)
 (local.get $0)
 (local.get $3)
)
```

aleksuss:
https://github.com/near/cargo-near/blob/main/cargo-near-build/src/near/build/mod.rs#L448
cc @mrLSD

engine-tests/src/tests/repro.rs
```diff
@@ -89,7 +89,7 @@ fn repro_5bEgfRQ() {
         block_timestamp: 1_651_073_772_931_594_646,
         input_path: "src/tests/res/input_5bEgfRQ.hex",
         evm_gas_used: 6_414_105,
-        near_gas_used: 452,
+        near_gas_used: 432,
```
birchmd, line 92: I love seeing gas usage go down! ❤️

engine-tests/src/tests/xcc.rs
```diff
@@ -136,7 +136,7 @@ fn test_xcc_eth_gas_cost() {
 
     assert!(
         utils::within_x_percent(
-            5,
+            25,
```
birchmd, line 139: The XCC gas cost changing is interesting. It might need a full re-calibration if the engine is significantly more efficient now. That can be done as a separate issue though.

engine-precompiles/src/secp256r1.rs
```diff
@@ -67,6 +67,7 @@ impl Secp256r1 {
         // 3. Signature Component Validation
         // Spec: "Both r and s MUST satisfy 0 < r < n and 0 < s < n"
         // `Signature::from_scalars` returns an Error if scalars are zero or >= group order (n).
+        #[allow(deprecated)]
```
birchmd, line 70: I'm curious what is deprecated here. It's not a blocker to fix it right away, but we should probably fix it before the next version of this library.

birchmd (approved):
Great job! I'm happy to see all the new clippy fixes and performance improvements from being on a more recent version of rust.

engine-tests/src/tests/xcc.rs
aleksuss: Yes. Agreed. But a full re-calibration is a potential breaking change, and it's better to do it when we have a BE with fork support.

engine-precompiles/src/secp256r1.rs
aleksuss: The current `p256` version uses `GenericArray` < `1.x.x` version, which is considered deprecated. I thought about bumping `p256`, but the next version is RC-only. Will fix it when the stable version is released.

PR #1201, opened 2026-08-22
fix(tests): synchronise solidity artifact compilation

4waan:
`compile_from_source` decided whether to compile by reading the contract's `.bin` then opened the .abi unconditionally with nothing holding those steps together against other threads. also solc writes the .bin before the .abi meaning a thread reads bytecode that another threads in flight compile had just written.. skip compiling and then panic on an .abi that didnt exist yet:

    thread 'tests::erc20_connector::workspace::test_exit_to_near_refund'
    panicked at engine-tests/src/utils/solidity/mod.rs:78:45:
    called `Result::unwrap()` on an `Err` value:
    Os { code: 2, kind: NotFound, message: "No such file or directory" }

`ERC20PresetMinterPauser` is loaded from 15 call sites and CI runs 32 test threads.. any run starting without a warm target/solidity_build has several threads compiling the same contract at once..
this could surface later as an unexplained evm failure since a truncated still decodes and reading a half writen .bin is quieter form of same race

give artifacts a lock and hold across whole check compile read sequence keyed by artifact path so unrelated contracts still compile concurrently require both artifacts to be present before reuse because one's presence doesnt say much about the other.. route `force_compile` through the same lock to avoid overwriting artifacts being read by another thread Failures now name the file that could not be read.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solidity compilation reliability when multiple operations target the same contract.
  * Prevented incomplete compilation artifacts from being reused.
  * Added clearer error reporting when compiled artifacts cannot be loaded.
  * Improved recovery after interrupted or failed compilation operations.

* **Chores**
  * Improved cleanup tasks to remove generated contract files, test build outputs, Cargo outputs, and binaries more reliably.
  * Updated cleanup documentation to reflect the expanded cleanup behavior.

### DIFF
--- a/engine-tests/src/utils/solidity/mod.rs
+++ b/engine-tests/src/utils/solidity/mod.rs
@@ -1,10 +1,12 @@
 use crate::prelude::{Address, U256, transactions::legacy::TransactionLegacy};
 use aurora_engine_types::types::Wei;
 use serde::Deserialize;
+use std::collections::HashMap;
 use std::fs;
 use std::io::BufReader;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::sync::{Arc, LazyLock, Mutex, PoisonError};
 
 pub mod erc20;
 pub mod exit_precompile;
@@ -43,13 +45,15 @@ impl ContractConstructor {
         P2: AsRef<Path>,
         P3: AsRef<Path>,
     {
-        compile(&sources_root, &contract_file, &artifacts_base_path);
-        Self::compile_from_source(
-            sources_root,
-            artifacts_base_path,
-            contract_file,
-            contract_name,
-        )
+        let artifacts_base_path = artifacts_base_path.as_ref();
+        let lock = artifact_lock(artifacts_base_path, contract_name);
+        let guard = lock.lock().unwrap_or_else(PoisonError::into_inner);
+
+        compile(&sources_root, &contract_file, artifacts_base_path);
+        let constructor = Self::load_artifacts(artifacts_base_path, contract_name);
+
+        drop(guard);
+        constructor
     }
 
     // Note: `contract_file` must be relative to `sources_root`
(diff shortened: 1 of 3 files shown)


4waan:
This fell out of the CI failure on #1200 (on file changed: `engine-modexp/src/mpnat.rs`)
```
thread 'tests::erc20_connector::workspace::test_exit_to_near_refund'
panicked at engine-tests/src/utils/solidity/mod.rs:78:45:
called `Result::unwrap()` on an `Err` value:
Os { code: 2, kind: NotFound, message: "No such file or directory" }
```
line78 col 45 is the `.unwrap()` on `fs::File::open(abi_path)`. The `.bin` read immediately above it succeeded, so at that instant the bytecode existed and the ABI did not. 

also #1200 going green doesnt close this because the difference came from cache
- [attempt 2, failed](https://github.com/aurora-is-near/aurora-engine/actions/runs/32502349228/job/96865944859): `... Restoring cache ...` then `No cache found.`
- [attempt 3, passed](https://github.com/aurora-is-near/aurora-engine/actions/runs/32502349228/job/96891994878): `Cache restored successfully`

window opens only when the artifacts are cold: a new cache key, an eviction, the first run on a branch. Rare enough to read as flakiness but frequent enough to keep costing re runs. 
NOTE: the `RUST_TEST_THREADS: 32` line in `tests.yml` is already commented "prevent the occasional failure of the tests" and that suggests this has been felt before and answered by lowering concurrency rather than by syncing..

engine-tests/src/utils/solidity/mod.rs
```diff
@@ -164,6 +193,36 @@ impl DeployedContract {
...
+fn artifact_lock(artifacts_base_path: &Path, contract_name: &str) -> Arc<Mutex<()>> {
+    let key = artifacts_base_path.join(contract_name);
+    // These locks guard files rather than data, so a panicking test leaves nothing here
+    // to be corrupted. Recovering from the poison keeps that original panic visible
+    // instead of burying it under a `PoisonError` in every test that follows.
+    let mut locks = ARTIFACT_LOCKS
+        .lock()
+        .unwrap_or_else(PoisonError::into_inner);
+
+    Arc::clone(locks.entry(key).or_default())
```
4waan:
```suggestion
fn artifact_lock(artifacts_base_path: &Path, contract_name: &str) -> Arc<Mutex<()>> {
    fs::create_dir_all(artifacts_base_path).unwrap();
    let artifacts_base_path = fs::canonicalize(artifacts_base_path).unwrap();
    let key = artifacts_base_path.join(contract_name);
    // These locks guard files rather than data, so a panicking test leaves nothing here
    // to be corrupted. Recovering from the poison keeps that original panic visible
    // instead of burying it under a `PoisonError` in every test that follows.
    let mut locks = ARTIFACT_LOCKS
        .lock()
        .unwrap_or_else(PoisonError::into_inner);

    Arc::clone(locks.entry(key).or_default())
}
```

aleksuss:
@4waan , I've added some changes:

- Refactored Solidity compilation into one synchronized helper, locking the canonical output directory across cache checks, compilation, and artifact reads.
- Poisoned locks now trigger recompilation and clear only after successful artifact loading.
- Added concurrent compilation and poison-recovery regression tests in mod.rs.

 Take a look, please.

aleksuss (approved):

aleksuss:
@coderabbit full review

mrLSD (approved):

